PR #42, opened 2016-09-10
New summary

jake-westfall:
See notes on individual commits. These changes should make summarizing and working with fitted models considerably nicer.


jake-westfall:
Hmm, strange, I can't seem to recreate this error on my machine...

jake-westfall:
This _could_ be because the travis version of pymc3 doesn't like that the `df_summary` DataFrame has all NaN values for the `mc_error` column, presumably because the number of samples in the test model (10) is too small to compute this. I'll try setting the number of samples to the smallest number that is large enough to fill in the `mc_error` column (which seems to be about 100) and see if travis likes that.

jake-westfall:
Well, now it's a different error... so I guess that's progress?

bambi/models.py
```diff
@@ -174,6 +174,13 @@ def build(self):
                 if not isinstance(t.prior, Prior):
                     scaler.scale(t)
 
+        # For binomial models with n_trials = 1 (most common use case),
+        # tell user which event is being modeled
+        if self.family.name=='binomial' and np.max(self.y.data) < 1.01:
+            event = next(i for i,x in enumerate(self.y.data.flatten()) if x==1)
+            print('NOTE: Modeling the probability that {}==\'{}\''.format(
```
tyarkoni: We should probably use a `warning` for this rather than printing; see example earlier in the code.

bambi/models.py
```diff
@@ -285,9 +292,15 @@ def add_formula(self, fixed=None, random=None, priors=None,
                 data[cats] = data[cats].apply(lambda x: x.astype('category'))
 
             if '~' in fixed:
+                event = re.match(r'^((\S+)\[(\S+)\])\s*~(.*)$', fixed)
```
tyarkoni: Can you add a comment line here to explain what's being matched?

bambi/models.py
```diff
@@ -285,9 +292,15 @@ def add_formula(self, fixed=None, random=None, priors=None,
...
+                if event is not None:
+                    fixed = '{}~{}'.format(event.group(2),event.group(4))
                 y, X = dmatrices(fixed, data=data)
                 y_label = y.design_info.term_names[0]
-                self.add_y(y_label, family=family, link=link)
+                self.add_y(y_label, family=family, link=link,
+                    data=None if event is None else
```
tyarkoni: Maybe put the data assignment on its own line, for clarity. And the data dict inside `pd.DataFrame` could also be broken up to make things clearer.

bambi/results.py
```diff
@@ -30,6 +31,10 @@ def plot(self):
     def summary(self):
         pass
 
+    @abstractmethod
+    def get_trace(self):
```
tyarkoni: I don't think we want to make this an abstract method, because not every Results class will have a `trace` attribute, only those based on MCMC sampling (e.g., right now the `PyMC3ADVIResults` class doesn't). If we want to be strict about this, we should probably add another class to the hierarchy that's called `MCMCModelResults` or something. But it seems like overkill for now, so maybe just cut this out and only leave it implemented on the `PyMCResults` class.

bambi/results.py
```diff
@@ -114,19 +132,96 @@ def plot(self, burn_in=0, names=None, annotate=True, hide_transformed=True,
                     textcoords='offset points', rotation=90, va='bottom',
                     fontsize='large', color='#AA0022')
 
+        # For binomial models with n_trials = 1 (most common use case),
+        # tell user which event is being modeled
+        if self.model.family.name=='binomial' and np.max(self.model.y.data) < 1.01:
+            event = next(i for i,x in enumerate(self.model.y.data.flatten()) if x==1)
+            print('NOTE: Modeling the probability that {}==\'{}\''.format(
```
tyarkoni: Same here--replace with `warnings.warn`

bambi/results.py
```diff
@@ -114,19 +132,96 @@ def plot(self, burn_in=0, names=None, annotate=True, hide_transformed=True,
...
+            if exclude_ranefs:
+                names = [x for x in names
+                    if x[2:] not in list(self.model.random_terms.keys())]
+
+        # get the basic DataFrame
+        df = pm.df_summary(self.trace[burn_in:], varnames=names, **kwargs)
+
+        # replace the "__\d" suffixes with an informative factor level name
+        match = [re.match('^(.*)(?:__)(\d+)?$', x) for x in df.index]
+        def replace_with_name(match):
```
tyarkoni: It might be worth abstracting this into its own class-level method (could also be a helper function in `utils.py`) and adding a test for it, because we want to make absolutely sure we never return the wrong names.

bambi/results.py
```diff
@@ -114,19 +132,96 @@ def plot(self, burn_in=0, names=None, annotate=True, hide_transformed=True,
...
+                return '{}[{}]'.format(term.name, term.levels[int(match.group(2))])
+        new = [replace_with_name(x) if x is not None else df.index[i]
+            for i,x in enumerate(match)]
+        df.set_index([new], inplace=True)
+
+        # For binomial models with n_trials = 1 (most common use case),
+        # tell user which event is being modeled
+        if self.model.family.name=='binomial' and np.max(self.model.y.data) < 1.01:
+            event = next(i for i,x in enumerate(self.model.y.data.flatten()) if x==1)
+            print('NOTE: Modeling the probability that {}==\'{}\''.format(
```
tyarkoni: Again, use `warning.warn` here

bambi/results.py
```diff
@@ -114,19 +132,96 @@ def plot(self, burn_in=0, names=None, annotate=True, hide_transformed=True,
...
 
-        return pm.summary(self.trace[burn_in:], varnames=names, **kwargs)
+        # helper function to label the trace DataFrame columns appropriately
+        def get_cols(var):
+            # handle terms with a single level
+            if len(self.trace[var].shape)==1 or self.trace[var].shape[1]==1:
+                return [var]
+            else:
+                # handle fixed terms with multiple levels
+                # (slice off the 'b_' or 'u_')
```
tyarkoni: This is unrelated to the current PR, but we may want to consider a package-level config module that lets the user specify whether they want the `b_` and `u_` prefixes added in the first place (we should probably default to not mucking with the names at all).

bambi/results.py
```diff
@@ -114,19 +132,96 @@ def plot(self, burn_in=0, names=None, annotate=True, hide_transformed=True,
...
+        # For binomial models with n_trials = 1 (most common use case),
+        # tell user which event is being modeled
+        if self.model.family.name=='binomial' and np.max(self.model.y.data) < 1.01:
+            event = next(i for i,x in enumerate(self.model.y.data.flatten()) if x==1)
+            print('NOTE: Modeling the probability that {}==\'{}\''.format(
+                self.model.y.name, str(self.model.data[self.model.y.name][event])))
+
+        return df
+
+    def get_trace(self, burn_in=0, names=None, hide_transformed=True):
```
tyarkoni: This is great! But maybe we should add the same `exclude_ranefs` flag as in `summary`, since most of the columns are taken up by individual ranef parameters.

bambi/tests/test_model.py
```diff
@@ -380,6 +382,17 @@ def test_cell_means_with_random_intercepts(crossed_data):
     priors1 = {x.name:x.prior.args['sd'].args for x in model1.terms.values() if x.random}
     assert set(priors0) == set(priors1)
 
+    # test summary
+    fitted.summary()
+    fitted.summary(exclude_ranefs=False)
```
tyarkoni: We should probably create a new `test_results` module, and add a fixture with fake, deterministically generated `Model` and PyMC3 `MultiTrace` inputs, so that we can test all of these methods in a modular way. We definitely want more assertions here.

tyarkoni:
Left some line comments that should probably be addressed, but otherwise this looks great!

jake-westfall:
Addressing the line comments as we speak! Also almost done updating the ESCS_multiple_regression notebook and wrapping up the ANES_logistic_regression notebook. I agree that it would be better to have the tests of the summary functions/plots in a separate test module.

jake-westfall:
What the hell. My version of pymc3 doesn't add this trailing underscore to transformed variables, like the travis version apparently does.

tyarkoni:
Looks good!

PR #57, opened 2017-01-05
adds more conservative PyMC3 sampling initialization defaults

tyarkoni:
In recent releases, PyMC3 defaults to using ADVI to find a sensible starting point for sampling. It also defaults to a rather large number of initialization iterations (200,000). While this no doubt ensures saner starting points, for the kind of models our users are likely to be fitting, the initial ADVI run is likely to add a non-negligible amount of computation (e.g., I just started fitting a model with ~80k rows and no very many parameters, and the estimated time just for advi to finish is nearly 2 hours!). Also, at least in the one case I tested, initializing with ADVI rather bizarrely appears to break the sampler, in that it causes NUTS to return crazy values (with virtually no variances in the traces) for most parameters.

Since all extra args and kwargs passed to `Model.run()` are passed onto the sampler, the user can easily control this by setting `init=None` and/or `n_init` to a much smaller value. But since the point is to prevent our users from having to think about sampling details as much as possible, I suggest we explicitly set the default `init` to `None` and the default `n_init` to `10000`. I've made changes accordingly.

### DIFF
--- a/bambi/backends.py
+++ b/bambi/backends.py
@@ -122,7 +122,8 @@ class PyMC3BackEnd(BackEnd):
 
             self.spec = spec
 
-    def run(self, start=None, method='mcmc', find_map=False, **kwargs):
+    def run(self, start=None, method='mcmc', init=None, n_init=10000,
+            find_map=False, **kwargs):
         '''
         Run the PyMC3 MCMC sampler.
         Args:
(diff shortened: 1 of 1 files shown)


twiecki:
@tyarkoni That sounds like a pymc3 problem to me. @aloctavodia just found a problem with starting values: https://github.com/pymc-devs/pymc3/pull/1645 Can you see if that changes something?

PR #67, opened 2017-02-23
Stan

tyarkoni:
This PR adds a new `StanBackEnd` that should (in theory) do almost everything that the `PyMC3BackEnd` does. Using it is as simple as specifying `backend='stan'` when initializing a new `Model`. The `StanBackEnd` wraps PyStan, but converts the sampling output into a PyMC3 `MultiTrace` object, which allows us to use all of the existing plotting/summarizing machinery (at least in principle; in practice, I had to hack my way around the normal `MultiTrace` initialization process, so some attributes are probably not set properly).

There are some things to iron out before we can merge this. @jake-westfall, maybe you can take a pass at these? We need to do the following:

* Test everything more thoroughly to make sure it works.

* Modify at least a few of the existing tests to run using the Stan backend. It's probably also worth adding a couple of tests that run with both backends and make sure the resulting estimates are close. We also need to modify the travis config to install PyStan (and maybe add an optional_dependencies.txt file containing PyStan that people can `pip install -r` from).

* Add support for all of the PyMC3 distributions we currently support. The trick here is to map PyMC3 distribution names and arguments onto the Stan language. This is accomplished via the `dists` dictionary in `StanBackEnd`, which maps the PyMC3 dist names onto dictionaries containing the Stan names, argument order, and (optionally) and bounds to impose. The latter is required because in Stan there are no separate distributions for clipped distributions; e.g., half-cauchy is just cauchy with a lower bound of 0. In principle I think all we need to do is add new entries to the `dists` dictionary; I don't think any of the PyMC3 distributions we currently support should introduce any new complications.

* One gotcha right now is that PyStan ignores the `incl_warmup` argument in `extract()` if `permuted=True`. This is reasonable in one sense, because the permuted results concatenate all chains together, and it's arguably weird to include the different initialization points. But it's a problem for us, because to make results commensurable with the `PyMC3BackEnd`, we should really include the burn-in and let the user drop it themselves later. The solution is unfortunately probably going to require us to duplicate some of the PyStan code for reformatting the Stan output, but with the burn-in included.

* Add support for different error distributions. Right now the `StanBackEnd` only supports normally distributed errors, but we should at minimum support the options we support with PyMC3. This shouldn't take much work; I think we just need to modify the specification of the `yhat` term and the model error.


tyarkoni:
The tests are passing on Python3 but failing on Python2 on travis. The pymc3 import seems to be failing on Python2; not sure if this is a temporary glitch somewhere or something we need to address, but I guess we can wait and see.

tyarkoni:
Couldn't figure out the Python2 import issue, but it goes away if I move all the `BackEnd` classes back into a single module. So I went with that for now.

bambi/results.py
```diff
@@ -34,80 +37,77 @@ def summary(self):
         pass
 
 
-class PyMC3Results(ModelResults):
-
+class MCMCResults(ModelResults):
```
tyarkoni: Do we actually need separate `MCMCResults` and `SampleArray` classes at this point? I don't have any objection to keeping them both, but I'm wondering if we couldn't just consolidate them both into `MCMCResults`. I guess the main benefit of keeping the separation is we could potentially store multiple `SampleArray`s in a single `MCMCResults` (especially given the suggested pattern of `.fit(..., backend=['stan', 'pymc'])`). But I think returning a list of `MCMCResult` objects is probably more intuitive in that situation anyway. Thoughts?

jake-westfall:
I'm not committed to the current organization at all. I mainly just wanted to get the SampleArray code into the codebase so we could start discussing it and what the best organization might be. Rolling that functionality into MCMCResults makes sense to me.

twiecki:
What's the benefit of the Stan backend?

tyarkoni:
I don't know that there's a clear benefit to the Stan backend in particular; the idea is more to have a backend-independent way of fitting common models—and currently PyMC and Stan are the two obvious packages to support. As far as why we would want to support multiple backends, there are several benefits:

* Different backends could in principle differ substantially in their efficiency at fitting certain kinds of models, so it's nice to be able to switch between them instantly if one backend runs into problems on a given model;
* It makes comparison of multiple backends very easy, in terms of both parameter estimates and speed;
* It separates plotting/diagnostics/summarization from sampling, so that (a) we theoretically only need to write new functionality once, which instantly applies to results from all backends, and (b) we ensure that identical code is applied independently of backend (e.g., there's no possibility for differing implementations in summaries, plots, etc.).

jake-westfall:
The latest commits completely phase out pymc3 MultiTrace objects. pymc3 and stan now return identical MCMCResults objects (except for internal/transformed parameters). And now you can do nice slicing and selection directly on MCMCResults objects like:

    fit = model.fit('Y ~ continuous + dummy + threecats',
           random=['0+threecats|subj', '1|item', '0+continuous|item',
                   'dummy|item', 'threecats|site'],
           backend='pymc3', samples=1000, chains=2)
    fit[50:, ['Intercept','threecats']].get_chains(1).to_df()

which returns a 3-column DataFrame of samples for the intercept and the two `threecats` dummy predictors. The `to_df()` method replaces the older `get_trace()` method, which never actually made it into any of the documentation or examples. 

Summary and plotting methods not implemented yet, they're next. That and implementing effective_n, gelman_rubin, and HPD intervals from scratch.

bambi/backends/stan.py
```diff
@@ -69,9 +69,13 @@ def build(self, spec, reset=True):
...
-            ''' Stan only allows alphanumeric chars and underscore, so replace
-            all invalid chars with '_' and track for later re-substitution. '''
-            clean = re.sub('[^a-zA-Z0-9\_]+', '_', name)
+            ''' Stan only allows alphanumeric chars and underscore, and
+            variable names must begin with a letter. So replace all invalid
+            chars with '_', prepend with 'b_', and track for later
+            re-substitution. '''
+            if name in self._original_names:
+                return name
+            clean = 'b_' + re.sub('[^a-zA-Z0-9\_]+', '_', name)
```
tyarkoni: I thought we were going to drop the 'u_' and 'b_' prefixes? Or do you want to leave them in for now and handle that separately? I'm fine either way.

tyarkoni:
The latest changes look great! I particularly like the implementation of `__getitem__`.

One issue that came up with the introduction of the non-centered parameterization (so far just to the `PyMC3BackEnd`, though stan is next) is that we need some way to flag the offset parameters for exclusion from default summarization/plotting. Do we want to stick with the PyMC3 convention of trailing underscores (or double underscores) denoting parameters that should be suppressed by default? Then I can rename the offset parameters to end with "offset_", and we can do the rest in all backends any time there's a parameter that should be suppressed (e.g., Stan's Lp).

bambi/backends/stan.py
jake-westfall: Yeah, so the u_ and b_ prefixes are gone as far as the user can see. (Well, unless they look at the underlying Stan model code, I guess.) The issue is just that Stan doesn't like parameter names to start with numbers, but this conflicts with our random intercept terms that have names like '1|subject'. So I had a prepend some arbitrary letter prefix to avoid this, and chose 'b_' for no particular reason -- I guess my thought was just that if a user did somehow look at the underlying Stan code and saw the b_ prefixes mysteriously added to the parameter names, maybe they wouldn't think it looked too weird.

jake-westfall:
Yeah, I saw that this would be an issue, but the current changes ignore it. (Since the offsets are random terms, they will be excluded from the `.summary` and `.to_df` output by default anyway, but they will pollute the output if someone requests hide_ranefs=False.) The trailing underscore convention seems like a simply, easy solution, but of course it could potentially lead to problems if users name parameters with underscores at the end for some reason. I kind of thought the current solution implemented in `_get_transformed_vars` (which does something like checking for parameters names that are another parameter name plus some stuff, although the code I wrote back then is not too obvious at first glance...) might automatically catch and flag these, but that didn't appear to be the case. But anyway, I think a cleaner solution might involve making a little tweak to `_get_transformed_vars` so that the offset variables get included in the resulting list, even if it's just some dumb strategy like checking for parameter names ending with '_offset' :p

bambi/models.py
```diff
@@ -746,9 +746,9 @@ def __init__(self, name, data, categorical=False, random=False, prior=None,
 
         # For the sake of computational efficiency (i.e., to avoid lots of
         # large matrix multiplications in the backends), invert the dummy-
-        # coding process and represent binary categoricals as a vector of
+        # coding process and represent full-rank dummies as a vector of
         # indices into the coefficients.
-        if data.shape[1] > 1 and ((data == 0) | (data == 1)).all():
+        if constant and data.shape[1] > 1 and ((data == 0) | (data == 1)).all():
```
tyarkoni: Hmm... can we keep using the same approach for reduced-rank coding, but just index from 0 below if `not constant`, instead of from 1? Or would that still break something?

jake-westfall: I don't think that particular tweak would work (although I admittedly haven't tried it), because the main issue is that, for the reference level, there is no valid index for indicating the non-zero column because all columns are zero for the reference level. So we would probably have to do something involving a special value (like None or a negative index) for that case to differentiate it from the case of a valid index of 0. My guess is that we could work out a solution for this if we wanted to, and did I almost try. I decided not to simply because this would only affect fixed categoricals, which will rarely have more than handful of levels anyway, so it's not clear that the compression would have any noticeable impact anyway (unlike for random effects which may have many levels).

jake-westfall: Thinking about this also gave me the idea that we might be able to efficiently handle big random terms generally (i.e., not just random intercepts but also random slopes) by using some basic sparse matrix multiplication techniques, which is essentially what we are doing now. It looks like scipy has some sparse matrix classes, but we might not need all that machinery if the only two things we need to do are represent the data and multiply it by a vector. I found [this nice page](https://www.cs.cmu.edu/~scandal/cacm/node9.html) which makes me think it might actually be fairly easy to do ourselves for cheap. Could be worth exploring more.

tyarkoni: Your explanation re: the n-1 coding makes sense. Re: sparse matrices, we could certainly represent represent the data as sparse scipy matrices, but that wouldn't really save us any work in terms of the backends, because I don't want to deal with sparse matrices in the backends (and AFAIK, Stan doesn't even support them). So we would still need to collapse the matrices into the index-based representation.

The other issue is that we use patsy to generate the predictor matrix, and unless patsy can implicitly create and return sparse matrices, we're already committed to representing the dense matrix in memory at least transiently, so we wouldn't really gain anything unless you think it will help with the scaling calculations. But so far it doesn't seem like anything other than the backend computations is slowing things down.

bambi/backends/stan.py
```diff
@@ -74,11 +97,13 @@ def build(self, spec, reset=True):
...
             ''' Stan only allows alphanumeric chars and underscore, and
             variable names must begin with a letter. So replace all invalid
-            chars with '_', prepend with 'b_', and track for later
+            chars with '_', prepend with 'b_' if needed, and track for later
             re-substitution. '''
             if name in self._original_names:
                 return name
-            clean = 'b_' + re.sub('[^a-zA-Z0-9\_]+', '_', name)
+            clean = re.sub('[^a-zA-Z0-9\_]+', '_', name)
+            if re.search(r'^\d', clean) is not None:
```
tyarkoni: This makes sense, but it turns out that Stan reserves a whole bunch of names that can't be used as variables, and I think it will probably be fairly common to run into these. E.g., the ESCS example won't run because 'floor' is a function name. So I'm going to revert to always prepending variable names, even if they don't start with numbers, just to be on the safe side.

jake-westfall:
The second image in the README (of the .summary() output) is not displaying correctly in the Stan branch, but it'll display correctly after merging into master.

tyarkoni:
Not sure why tests are failing on travis; they were passing on the stan branch just before the merge. They also pass locally.

jake-westfall:
Looks like this is discussed [here](https://github.com/pymc-devs/pymc3/pull/1934), and is [fixed](https://github.com/pymc-devs/pymc3/pull/1934/commits/33406e88305a45b9920610ca8290df6a84211c3b) on the pymc3 master as of 9 days ago. But evidently not on the pip version. Any thoughts about (or is it even possible to) making travis install pymc3 from github instead of pip? I'm really not sure how many of our users would be installing the pymc3 dependency from pip vs. github, especially since we give instructions for both ways in our README.

tyarkoni:
It's certainly possible to install pymc3 from master on travis, but the problem with doing that is we then won't know if the current PyPI release fails, which is what most people will be using, and is what we should probably support. We could put both master and latest PyPI in the travis matrix, which would cover all the bases.

FWIW, it looks like the fix for that issue *is* in the latest PyPI release (from 3/30), which was before I merged the Stan branch... so I'm not sure what's going on. Also doesn't explain why the tests passed in the stan branch but not when merged. Let's see if the currently running tests pass.

twiecki:
we'll release 3.1 very soon. there's already a release candidate

tyarkoni:
@twiecki cool, thanks for the update. We'll probably just sit tight and wait for that then, and in the unlikely event that anyone has problems in the meantime, we'll tell them to update from master.

PR #68, opened 2017-02-26
Result names

jake-westfall:
This fixes a few different bugs with .summary() and .get_trace(). Some of these were bugs that were introduced in the previous commits that add diagnostic info to .summary(), but that only occurred in cases that don't appear in our tests. One appeared to be a long-standing (but not previously documented) bug with models that feature random slopes of categorical predictors with >2 levels.

The reason these cases have slipped through our tests is because testing them would require actually compiling/sampling such models, which would add a lot of testing time, and we have been trying to compile as few models as possible in the tests because they already take forever. 

Sadly I suspect most of this new code will have to be reworked again when we eventually merge in the Stan branch, but in the meantime I think these fixes are fairly high-priority.

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -508,7 +508,6 @@ class Model(object):
             data = dmatrix(f, data=data, NA_action=Ignore_NA())
             cols = data.design_info.column_names
             data = pd.DataFrame(data, columns=cols)
-            # data = pd.DataFrame(data, columns=[c.split(':')[0] for c in cols])
 
             # For categorical effects, one variance term per predictor level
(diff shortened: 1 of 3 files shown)


tyarkoni:
Looks good. In theory, I don't think the `StanBackEnd` should require extensive changes to this code; I don't see (m)any places where what you're doing here specifically touches the `PyMC3BackEnd`, but I could be missing something. The naming conventions *should* be identical for both back-ends (and if they currently aren't, we should probably fix that in the respective backends, rather than accounting for discrepancies after the fact).

PR #71, opened 2017-03-13
Fix patsy import

tyarkoni:
Tests are failing on Python 3 because `bambi.external.patsy` is trying to import from `patsy`, which has the same name. Solution is to treat unqualified import statements as absolute.


tyarkoni:
Incidentally, this is almost certainly the same issue that I was running into when trying to restructure the backends module into separate `pymc` and `stan` modules, so now we can revisit that as well.

tyarkoni:
Actually, I'm closing this PR and adding it to the stan instead, since this includes all of those commits.

PR #85, opened 2017-04-02
Added basic API documentation from docstrings.

ejolly:


tyarkoni:
Thanks! I'll leave the issue open for now, as there's still some more work to be done, but this is a great start!

PR #92, opened 2017-04-16
Preserve variable ordering when filtering names

amelio-vazquez-reina:
This change will preserve variable ordering in the `results` interface.

For example, the following will preserve the order of variables in `variables_of_interest` when plotting (top-to-bottom):
``` python
results[400:].plot(ranefs=True, varnames=variables_of_interest)
```


tyarkoni:
Thanks!

PR #93, opened 2017-04-21
Categorical na

jake-westfall:
Categorical NAs are now handled nicely, and this situation is now reflected in our tests. A code review would be nice on this one!

Basically `Model.add()` and `.set_priors()` are now internal functions `_add()` and `_set_priors()`, and they are replaced by new front-end functions `.add()` and `.set_priors()`, both of which store any arguments passed to them as lists of dictionaries in the attributes `.added_terms` and `.added_priors`. The new front-end `Model.add()` also does a bit of additional work in screening the added terms for missing values (by monitoring what patsy decides to drop, so no messy string parsing is required on our end) and keeping a running list of the complete cases in the dataset, given the added terms. Then `Model.build()`, when eventually called, first loops over all the dictionaries in `added_terms` and `_add()`s them; then loops over the dictionaries in `added_priors` and calls `_set_priors()` on them. In the latter case, we discussed possibly having a single prior dictionary that gets updated with each call to `set_priors()`, but a complication there would be that the user might want different values of `match_derived_names` for different calls to `set_priors()`, in which case something akin to this solution would be necessary anyway.

Also fixes a previously-undiscovered bug where calls like `.set_priors({'BMI', 0.3})` never actually resulted in the prior for `BMI` being set to Normal with scale=0.3, instead it was always using the "wide" scale value of .577.

And finally, modifies `Term.__init__()` so that you don't need to pass in the full `Model`, just the pieces of info that the `Term` actually uses, which turn out to just be a `PriorFactory` and an `auto_scale` flag.

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -72,9 +72,9 @@ class Model(object):
         self.noncentered = noncentered
         self._backend_name = None
 
-        # build() will loop over these lists of argument dictionaries
+        # build() will loop over these, calling _add() and _set_priors()
         self.added_terms = []
-        self.added_priors
+        self.added_priors = []
 
         # if dropna=True, completes gets updated by add() to track complete cases
         self.completes = []
@@ -366,6 +366,7 @@ class Model(object):
              link=None, categorical=None, append=True):
         '''
         Internal version of add(), with the same arguments.
+        Runs during Model.build()
         '''
 
         data = self.clean_data
@@ -583,7 +584,7 @@ class Model(object):
                 and so on. If False, an exact match is required for the
                 prior to be applied.
         '''
-        # save arguments to pass to _add()
+        # save arguments to pass to _set_priors() at build time
         args = dict(zip(
             ['priors', 'fixed', 'random', 'match_derived_names'],
             [priors, fixed, random, match_derived_names]))
@@ -594,7 +595,8 @@ class Model(object):
     def _set_priors(self, priors=None, fixed=None, random=None,
                     match_derived_names=True):
         '''
-        Internal version of set_priors(). Runs during Model.build().
+        Internal version of set_priors(), with same arguments.
+        Runs during Model.build().
         '''
 
         targets = {}
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -407,7 +407,9 @@ class Model(object):
                 cols = X.design_info.column_names[_slice]
                 term_data = pd.DataFrame(X[:, _slice], columns=cols)
                 prior = priors.pop(_name, priors.get('fixed', None))
-                self.terms[_name] = Term(self, _name, term_data, prior=prior)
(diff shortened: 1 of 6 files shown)


bambi/models.py
```diff
@@ -108,49 +120,41 @@ def build(self, backend=None):
                 and look in self._backend_name.
         '''
 
+        # retain only the complete cases
+        n_total = len(self.data.index)
+        if len(self.completes):
+            completes = [set(x) for x in sum(self.completes, [])]
+            completes = set.intersection(*completes)
+        else:
```
tyarkoni: I'm confused by the `len(self.completes)` check; suppose there's a case where NA-checking is done, but 100% of rows have at least one missing value. Then won't `self.completes` also be empty? Do we want then want to use all data anyway in that situation?

If the above is correct (and I'm probably missing something), we could probably default to `self.completes = None` and check that `is not None` rather than empty.

bambi/models.py
```diff
@@ -655,16 +725,20 @@ class Term(object):
...
+        default_priors: PriorFactory object
+        auto_scale (bool): If True (default), use default prior scaled to data.
+            Otherwise use flat prior.
     '''
     random = False
 
-    def __init__(self, model, name, data, categorical=False, prior=None,
-                 constant=None):
+    def __init__(self, name, data, categorical=False, prior=None, constant=None,
+        default_priors=PriorFactory(None), auto_scale=True):
```
tyarkoni: Values initialized in the signature will point to the same object in memory, and while this may not matter in practice here, it can lead to strange bugs in many cases, so is generally discouraged. Better to set to `None` and then initialize inside the method.

Also, do we need to pass an entire `PriorFactory` object? Can't we move the logic that checks if `auto_scale` is set (and appends `_flat` if so) to `Model._add` or `Model.add`, and then we can just use the `prior` argument instead (if it isn't explicitly set)?

jake-westfall:
Bizarrely, the test failure is because one of the default prior SDs evaluates to 4.99 in python3 and 4.41 in python2... I assume this has something to do with differing versions of `statsmodels` or something, but it's hard to see why even that would cause such a big substantive difference in the statistical answer.

bambi/models.py
jake-westfall: `len(self.completes)` should always be equal to the number of terms, because it is a list of arrays of length N, where N is the full number of rows. The if-else is just for the unlikely case of the empty model with no predictors (which is in our tests, although it's rarely of interest in the real world). If 100% of rows have a missing value, the `clean_data` DataFrame should end up empty, and the user should see a warning that N/N rows were dropped before see the eventual error.

bambi/models.py
jake-westfall: Yeah probably, that does sound nicer. I'll try that out.

PR #109, opened 2017-11-16
Improved custom priors

tyarkoni:
This PR makes minor changes related to prior handling:
* Fixes a problem where specifying a named custom prior causes model-fitting to break because some arguments are internally stored as numeric values instead of numpy arrays;
* Simplifies the processing of priors; instead of scattering `_prepare_prior` calls in several places, they're now consolidated into one call that only happens during the build;
* Fixes a minor error in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,7 +284,7 @@ my_favorite_prior = Prior('Laplace', mu=0., b=10)
 
 # Set the prior when adding a term to the model;
 # more details on this below.
-priors = {'subject': my_favorite_prior}
+priors = {'1|subject': my_favorite_prior}
 results = model.fit('y ~ condition', random='1|subject', priors=priors)
 ```
 
@@ -293,7 +293,7 @@ Priors specified using the `Prior` class can be nested to arbitrary depths--mean
 ```python
 subject_sd = Prior('HalfCauchy', beta=5)
 subject_prior = Prior('Normal', mu=0, sd=subject_sd)
-priors = {'subject': my_favorite_prior}
+priors = {'1|subject': my_favorite_prior}
 results = model.fit('y ~ condition', random='1|subject', priors=priors)
 ```
 
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -74,7 +74,7 @@ class Model(object):
 
         # build() will loop over these, calling _add() and _set_priors()
         self.added_terms = []
-        self.added_priors = []
+        self._added_priors = {}
 
         # if dropna=True, completes gets updated by add() to track complete cases
         self.completes = []
@@ -88,7 +88,7 @@ class Model(object):
         self.y = None
         self.backend = None
         self.added_terms = []
-        self.added_priors = []
+        self._added_priors = {}
         self.completes = []
         self.clean_data = None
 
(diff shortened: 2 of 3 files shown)


tyarkoni:
I'm running into a _really_ strange bug where if you build a model, fit it, and then try to build the same model from the same code a second time (i.e., reinitializing the `Model` instance), it fails. I've tracked this down to a problem with [this line](https://github.com/bambinos/bambi/blob/master/bambi/models.py#L645); the problem is that even though the passed `prior` is, to all appearances, an instance of class `Prior`, a check of `isinstance(prior, Prior)` returns `False`. I've verified that the `__class__` of `Prior` is in fact not identical to `Prior` imported directly from `bambi`.

This would seem to suggest that somewhere along the line, the class definition is being dynamically modified, and that this then causes downstream problems when initializing new `Prior` instances (because the modified definition is being invoked, but `isinstance` checks the `Prior` definition that was imported when the module was loaded). What's bizarre about this is that nothing is touching the `Prior` code as far as I can see, and this only happens if the model is actually fitted--so it's likely happening in the back-end, or even internal to PyMC3. I'll keep digging, but in the meantime, don't merge this just yet (although I actually don't think this is a new bug--I think I just only noticed it now).

bambi/priors.py
```diff
@@ -59,6 +59,11 @@ def update(self, **kwargs):
         Args:
             kwargs (dict): Optional keyword arguments to add to prior args.
         '''
+        # Backends expect numpy arrays, so make sure all numeric values are
+        # represented as such.
+        for k, v in kwargs.items():
```
jake-westfall: You don't need the outer for loop here, right? Just the dictionary comprehension?

bambi/models.py
```diff
@@ -678,22 +674,23 @@ def plot_priors(self, varnames=None):
...
             # add priors on Y params if applicable
             y_prior = [(k, v) for k, v in self.y.prior.args.items()
                        if isinstance(v, Prior)]
             if len(y_prior):
                 for p in y_prior:
-                    dists += [getattr(pm, p[1].name)('_'.join([self.y.name,
-                                                               p[0]]), **p[1].args)]
+                    pm_attr = getattr(pm, p[1].name)
+                    p = pm_attr('_'.join([self.y.name, p[0]]), **p[1].args)
+                    dists.extend([p])
```
jake-westfall: Nice readability edits. Also I finally installed a linter, so in theory I'll be avoiding some of the patterns above (like no space around operators) in the future, and will probably fix these kinds of things that tick off the linter as I come across them in the codebase :)

README.md
```diff
@@ -284,7 +284,7 @@ my_favorite_prior = Prior('Laplace', mu=0., b=10)
 
 # Set the prior when adding a term to the model;
 # more details on this below.
-priors = {'subject': my_favorite_prior}
+priors = {'1|subject': my_favorite_prior}
```
jake-westfall: I can't remember, is the previous syntax now actually erroneous, or just no longer the preferred syntax?

tyarkoni: It errors, because it can't find a term with that key. We could probably change it to do more clever matching, but I think strictness is fine here--we no longer allow passing just `'subject'` as a term to `random` (i.e., has to be `1|random`), so I think we may as well keep it like this.

I can no longer replicate the error I mentioned above. It's possible it was actually due to the `autoreload` magic doing wonky things, and not to a problem with the code. So feel free to merge once you've reviewed the PR.

jake-westfall:
It all looks good to me, except I'd remove that extra for loop I mentioned. After that, cool with me if you want to merge it in.

tyarkoni:
I can confirm that the "bug" I described above was just due to the fact that when you edit a module that's already loaded, ipython's autoreload will consider the reloaded object a different class.

PR #112, opened 2017-11-23
fixes Poisson link function; closes #111

### LINKED ISSUE #111: Bug in Poisson link function
This was [pointed out on Stack Overflow](https://stackoverflow.com/questions/45525631/incorrect-results-with-poisson-regression-using-bambi) a while ago, but I only just saw it (we need to get better about tracking SO posts--I just created an alert, and suggest you do too, @jake-westfall). Bambi is currently giving incorrect results for Poisson regression. This turns out to be due to an incorrect link function--it's currently specified in the PyMC3 back-end as log, when it should be exp. The problem actually stems from the same issue in PyMC3, as discussed [here](https://github.com/pymc-devs/pymc3/issues/1100) (though this is entirely my fault--I should have inspected the link functions more closely when porting them over).

This is a pretty big red flag that we need to be much more comprehensive about testing results for all major model classes for reasonable convergence with other packages (e.g, statsmodels). We should probably also start testing the Stan/PyMC3 backend outputs for convergence (we currently test priors, but I don't think we test parameter estimates). That would have helped here, because the Stan back-end produces the right result. I'll try to implement something more systematic when I have time (e.g., a simple step would be to write a helper function that takes two Stan and PyMC3-fitted models and tests that all parameter estimates are reasonably close).

Fix coming momentarily; I'm just detailing the bug here for posterity.

tyarkoni:
Fixes the link function for the Poisson family (now uses exp instead of log). See #111 for details.

### DIFF
--- a/bambi/backends/pymc.py
+++ b/bambi/backends/pymc.py
@@ -22,7 +22,8 @@ class PyMC3BackEnd(BackEnd):
         'identity': lambda x: x,
         'logit': theano.tensor.nnet.sigmoid,
         'inverse': theano.tensor.inv,
-        'log': theano.tensor.log
+        'log': theano.tensor.log,
+        'exp': theano.tensor.exp
     }
 
     dists = {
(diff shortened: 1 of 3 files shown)


tyarkoni:
Okay, I think this is ready. I set all link functions to use their inverses internally, and added an inverse square link to both backends. Will probably merge even if tests fail; they pass locally and appear to be timing out on travis.

jake-westfall:
Looks great to me. Thanks for taking care of this.

PR #119, opened 2018-02-13
Improve computation of effective sample size

junpenglao:
Cherrypicking the pymc3 PR: https://github.com/pymc-devs/pymc3/pull/2854. More information see https://github.com/pymc-devs/pymc3/issues/2810

### DIFF
--- a/bambi/diagnostics.py
+++ b/bambi/diagnostics.py
@@ -1,8 +1,50 @@
 import numpy as np
 import pandas as pd
+from scipy.signal import fftconvolve
 
 # code adapted from pymc3.diagnostics:
 # https://github.com/pymc-devs/pymc3/blob/master/pymc3/diagnostics.py
+
+
+def autocorr(x):
+    """
+    Compute autocorrelation using FFT for every lag for the input array
+    https://en.wikipedia.org/wiki/Autocorrelation#Efficient_computation
+
+    Parameters
+    ----------
+    x : Numpy array
+        An array containing MCMC samples
+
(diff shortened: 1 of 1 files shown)


tyarkoni:
Awesome, thanks!

This is exactly what's in PyMC3, right? @jake-westfall one thing we should consider is reverting to the former state of affairs where PyMC3 was a hard dependency. I doubt many people will have major objections to installing PyMC3 (even those who intend to use the Stan back-end). That way we could just pull this stuff out of PyMC3 rather than having to maintain the same code in two places.

Also, I notice one test seems to be failing, but can't take a careful look right now to see if it's something that was introduced by this PR, or random travis breakage. Will look at it when I can.

junpenglao:
Yep this is exactly what is in PyMC3. 

Actually you can call `pm.effective_n` as well, as it also accept numpy array with the right shape. Alternatively, we can expose the `get_neff` function so you can import it directly.

I think in the long run it might be useful to break out the diagnostic and visualization code to have a library similar to `coda` in R. 

The travis fail doesnt seem to related to this PR (actually I dont think you have a test for diagnostic), but seems it is a legit test fail.

tyarkoni:
Sorry about the absurdly long delay--forgot to follow up on the testing issue at the time. Thanks!

junpenglao:
No worries! We are developing https://github.com/arviz-devs/arviz which are cross backend (pymc3 and pystan currently), will likely implement faster version of effective sample size inside soon.

Issue #114, opened 2017-11-28
Better ordering of variables in plot()


tyarkoni:
I'm not sure how the order of variables is determined at the moment (probably just by the order of the dict keys when requested), but in practice, there isn't usually any rhyme or reason to the default order in the plot (though this can be manually controlled with `varnames`). We should probably default to at least grouping the fixed and random effects--and potentially grouping within the latter based on some simple guesses about parameter names (e.g., all the `1|*` terms and `_sd` terms get grouped separately).

aloctavodia:
Is this solved?

tyarkoni:
Looks that way; guess I forgot to close it. Thanks for the reminder (will get to your PRs ASAP, @aloctavodia—thanks!).

PR #124, opened 2018-06-05
Ordering of variables in plot()

GWeindel:
Following suggestion #114 Added a simple re-ordering code to produce an easier to read plot output. Order is defined as follow : Intercept, main effects, interactions, random effects and sd. Within a subcategory all names are ordered with length of the name.
Would be nice to apply to the summary output but I did not find a quick way to do it.


tyarkoni:
Sorry about the delay getting to this; looks good, thank you!

PR #125, opened 2018-07-23
Added example notebook comparing frequentist and bayesian analyses.

ejolly:
@tyarkoni Here's the example tutorial I mentioned I made in case it was a useful example.


tyarkoni:
This is awesome, thanks!

PR #135, opened 2019-03-21
update travis config

aloctavodia:
Attempt to fix conflict with conda and current travis configuration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ services:
   - xvfb
 
 script:
-  - travis_wait 30 pytest --pyargs bambi --cov-report term-missing --cov=bambi -vx
+  - pytest -vx
 
 after_success:
   - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+dist: xenial
+
 language: python
 
 python:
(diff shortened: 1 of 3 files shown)


aloctavodia:
OK, Travis is working again, but they seem unrelated to this PR. @tyarkoni do you have any hint?

tyarkoni:
Ugh, so many deprecation warnings... I think the whole codebase is in need of some spring cleaning. :(

Can't tell what's going wrong at a glance. The proximal problem looks to be in the renaming of the Stan  outputs, but hard to say if that's because something's breaking in Stan, PyStan, or Bambi itself (I haven't kept up with Stan developments, so if there have been changes to the language, it wouldn't shock me if the back-end no longer works properly). I'll try to find the time to look at this, but can't promise that it'll be very soon. @jake-westfall, any immediate insights?

ahartikainen:
We changed to 1 based indexing in PyStan 6 months ago

tyarkoni:
Ah, well that almost certainly explains it. :) Will try to work up a fix shortly.

aloctavodia:
I can also help with the spring cleaning. Can we accept PRs even when failing until the spring cleaning is done?

tyarkoni:
In principle, yes. But I should have a fix for this very shortly (it was indeed the 1-off error), so let's wait at least for that. If there are failures beyond that, accepting PRs anyway seems fine.

aloctavodia:
Sounds good to me!

tyarkoni:
The Python 2 tests now pass, so if you merge the changes in #136, we should be good to go. Thanks!

PR #143, opened 2019-03-28
Reorder requirements; Add matplotlib to requirements.txt

camenpihor:
Tests require matplotlib

I was not sure if python2.7 is still supported, and so I floored the requirement at `2.2.4` so that it would work with python3 as well.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,0 +1,3 @@
+pytest>=4.3.1
+matplotlib>=2.2.4
+
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-matplotlib>=2.2.4
 numpy
 pandas>=0.18.0
 patsy
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
+matplotlib>=2.2.4
 numpy
 pandas>=0.18.0
 patsy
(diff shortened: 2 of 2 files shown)


tyarkoni:
Thanks for the contribution, @camenpihor. I'm not sure we should include MPL in requirements if it's not strictly necessary for most functionality. It's a pretty heavy dependency to foist on people. The test environment can be (and currently is) set up independently. If we want to list matplotlib in `setup.py` as a test requirement, that seems fine as an alternative.

ahartikainen:
`test-requirements.txt` ?

ColCarroll:
+1 requirements-dev.txt. We also did this in pymc3 after [some discussion](https://github.com/pymc-devs/pymc3/issues/793).

camenpihor:
I added a requirements-dev.txt! 

Are there other things that you would like my help with? I took a look at the issues, and would be willing to learn to start tackling them, but would probably need some direction? Or I could start by just familiarizing myself with the library and do some tech-debt-like stuff (use a pylintrc, black, pydocstyle, take a look at speeding up tests, etc.)

Anyways, down to help in any ways that y'all would like :)

aloctavodia:
LGTM.

camenpihor:
Thanks for the comments everyone :)

PR #149, opened 2019-04-13
Update requirements

camenpihor:
* Add versions to requirements
* Add libraries necessary for testing and running notebooks
    * does not add `pymer4` since its pretty tricky to install

## Notes

* Versioning is floored at versions that support python 2


aloctavodia:
Great! `pymer4` is used only in one example, right? I agree is tricky to install it, maybe we should adapt that example to use just stats-models (assuming that's possible)

tyarkoni:
Haven't looked at the notebook in detail, but it looks like at least one of the examples includes crossed RFX, which AFAIK are still unsupported in statsmodels. @ejolly, thoughts?

I'm on board with removing `pymer4` from the list of test dependencies. I like that example notebook though, so worst case, we can just exclude it from testing.

As an aside, I notice that in the process of moving the notebooks, their history was erased. This isn't ideal. I don't suppose there's a way to recover that, @camenpihor? Failing that, we should at least add a note at the top of each one indicating who authored it...

ejolly:
Yea `pymer4` can be tricky to install and if it's only a testing dependency then maybe it can be ignored as per @tyarkoni's suggestion. Statsmodels has limited support for crossed-random effects (e.g. can't do fully crossed item _and_ subject random effects), but it can handle simple crossed random effects ok. I don't tend to use it very much because I find that that the estimates it yields are sometimes wonky compared to `lme4`. However, for the purposes of a demo I just created PR https://github.com/bambinos/bambi/pull/150 which includes a separate new notebook using statsmodels for the frequentist side of things, relying on a dataset they include with the package, and renaming the comparison notebook using pymer4. One thought is to include both notebooks, especially if users want to see how to simulate iid and random effects data or look at some "fancy" random effects plots in pymer4 alongside bambi's visuals :)

camenpihor:
Thanks for the reviews! And thanks for catching the lost history @tyarkoni ! Was there a way to avoid that? Would `git mv` have retained that? I will try now to recover, and I think regardless I should add a run date and author to the notebooks. 

As an aside, the notebooks aren't included in testing, but I generally feel like its nice to have a list of dependencies for notebooks in case someone wants to re-run or use them for learning (and I find that requirements-dev.txt is a good place to put those dependencies). All that's to say is that I think it's fine to not include pymer4 in the dependencies and still retain the notebook (I think it's a pretty cool notebook!), but maybe we should include some instructions on how to install pymer4 in the notebook or at least say that one should install it before running? -- I tried installing the python port of the require R packages using conda and had some trouble, and instead just installed R and the required packages. 

BUT I'm happy to do whatever someone else thinks is best!! :)

camenpihor:
I'll open up a new PR with the authors and run dates for the notebooks and their histories if I am able to  recover them. But let me know if there are more thoughts on this PR, thanks again everyone! :)

PR #151, opened 2019-04-13
Notebook history

camenpihor:
# This PR will include

* adding notebook history, if possible
* adding author, created at, last run date to notebooks


camenpihor:
@aloctavodia @tyarkoni this is a WIP PR for trying to get the notebook history, and I wanted a place to receive advice or thoughts on that task. 

All the notebooks except `shooter_crossed_random_ANOVA.ipynb` and `ANES_logistic_regression.ipynb` have a git blame that goes back to their creation (so they retain their history), but the two mentioned above don't have this. Also, despite having a git blame that goes back to the notebooks' creation a while ago, they all list only me as the contributor, do either of you have thoughts on this?

camenpihor:
In a separate branch I tried checking out the code before the documentation merge, and use `git mv` to move the notebooks, but it results in the same thing we are experiencing now (only a single contributor, but a git blame retains the history)

camenpihor:
What do y'all think I should try next?

If we can't think of anything, would you rather:

1. not have `docs/notebooks` which might mean no notebooks in documentation? (<- unsure about this), and re-merge the old notebooks back into `examples/` which does preserve their history, albeit we can't move them still.
2. leave the notebooks where they are and just add author?

I'm sorry that this has happened :/ I just looked at what other repositories had for documentation, and tried to follow the example of having notebooks included in the documentation and symlink to another directory in the root of the repository.

tyarkoni:
I don't think it's worth spending a lot of time trying to recover the history. I don't know why it got lost—normally git is pretty good about figuring out that files were moved or renamed. In any case, as long as there's an authorship note in each one, I think that's fine. My recollection is they're pretty much each written by only one person, minus some minor edits, so it's not a big deal.

camenpihor:
Will do! Sorry again :(

camenpihor:
This is ready for review! :)

PR #153, opened 2019-04-14
ADVI

### LINKED ISSUE #115: advi raises typeerror
calling `model.fit(backend='pymc3', run=True, method='advi')` raises `TypeError: __init__() got an unexpected keyword argument 'transformed_vars'`.

Looks like this happens in the `run` method [here](https://github.com/bambinos/bambi/blob/e1abd5dd4d2aecd2e8474d4ee6d3191a0193a53e/bambi/backends/pymc.py#L168), since the `PyMC3ADVIResults` is [only defined with the arguments](https://github.com/bambinos/bambi/blob/e1abd5dd4d2aecd2e8474d4ee6d3191a0193a53e/bambi/results.py#L402) `model, params`.

camenpihor:
Attempting to fix #115 

My read of this issue is that ADVI is not yet fully implemented since it doesn't have the `plot()` and `summary()` methods required by `ModelResults`, is that correct?

**If so, this is my in progress of implementing ADVI. I wanted to create a PR as a place to ask for advice and assistance, and so that others could tell if I am going down a path that is not desired by the maintainers.**

## Acceptance Criteria (I think?)

* Test(s) for ADVI compiling
* `.plot()` method at least implements a trace plot
* `.summary()` method shows useful summary statistics (I'm not too sure what summary statistics we want
* `to_df()` method which prints out trace nicely (following the paradigm set in `MCMCResults`
* any others?

## Notes

* The issue is labelled as a bug, so I thought that I would tackle it per the [Roadmap](https://github.com/bambinos/bambi/wiki/Roadmap), but let me know if we would rather wait to do this!
* I included a notebook at https://github.com/bambinos/bambi/blob/26c277b52bd924e4b756cf30c8d4cf79e1d92fcd/docs/notebooks/advi.ipynb so that people can see the progress of this PR (will remove once this PR is complete and able to be merged)


camenpihor:
@aloctavodia @tyarkoni  Do you think I should be working on this issue?

aloctavodia:
If we are adopting ArviZ, I think most of this work should be done on the ArviZ side arviz-devs/arviz#498

PR #158, opened 2019-05-14
update to python 3, black and fix some pylint issues

aloctavodia:
I did not touch `diagnostics.py` because we should replace that with ArviZ's diagnostics.


ColCarroll:
`coveralls` isn't working because `.travis.yml` uses 
```
script:
  - pytest -vx
```

needs a `--cov=bambi/`. I can open an issue if you don't want to fix here.

PR #159, opened 2019-07-12
use arviz

aloctavodia:
This PR introduces ArviZ as a dependency and is a first step to reduce the redundacy of base code between ArviZ and Bambi

I removed the diagnostic module and make the method `summary` call `az.summary` and the method `plot` call `az.plot_trace` (kind = trace) or `az.plot_forestplot(kind='ridgeplot')` (kind = prior), in a future PR I would like to think of a better alternative.


aloctavodia:
sorry that I send so many commits, but I have problems running the tests locally.

PR #163, opened 2019-08-29
Update contributing 

camenpihor:
according to https://github.com/markdownlint/markdownlint. 

One thing to note is that this mandates that the contributed code passed `pylint bambi/`, but this isn't something that currently passes. Should this criterion be removed? Should we keep it and try to update the codebase to pass pylint? (I'd be find doing this, to gain more familiarity with the codebase). 

I also don't know where bambi is with the main initiative of the final python2 release, and its cool if this pylint stuff wants to wait until after that happens!


aloctavodia:
Hi @camenpihor.

1) We hould we keep it and try to update the codebase to pass pylint
2) Everything on master and future release should be Python 3 only.

PR #171, opened 2019-09-05
Pylint priors

camenpihor:


bambi/priors.py
```diff
@@ -244,31 +246,33 @@ def _get_slope_stats(self, exog, predictor, sd_corr, full_mod=None, points=4):
                 for val in values[:-1]
             ]
             null = np.append(null, full_mod)
-            ll = np.array([x.llf for x in null])
+            approximation = np.array([x.llf for x in null])
```
aloctavodia: We could use log_likelihood instead of approximation

bambi/priors.py
```diff
@@ -244,31 +246,33 @@ def _get_slope_stats(self, exog, predictor, sd_corr, full_mod=None, points=4):
...
                 self.model.family.smfamily().loglike(np.squeeze(self.model.y.data), val * predictor)
                 for val in values[:-1]
             ]
-            ll = np.append(null, full_mod.llf)
+            approximation = np.append(null, full_mod.llf)
 
         # compute params of quartic approximatino to log-likelihood
         # c: intercept, d: shift parameter
-        # a: quartic coefficient, b: quadratic coefficient
+        # a: quadratic coefficient, b: quadratic coefficient
```
aloctavodia: It seems to me that quartic is correct for coefficient a, as is a reaised to the power of 4

camenpihor: b419ef5

bambi/priors.py
camenpihor: b419ef5

aloctavodia:
Thanks @camenpihor!

PR #177, opened 2019-09-13
remove from __future__ import absolute_import

camenpihor:
Since we moved to python3 we don't need this anymore


aloctavodia:
Thanks @camenpihor

PR #186, opened 2020-02-14
fix parallel sampling

aloctavodia:


bambi/models.py
```diff
@@ -498,7 +498,7 @@ def _add(
             # Loop over predictor terms
             for _name, _slice in x_matrix.design_info.term_name_slices.items():
                 cols = x_matrix.design_info.column_names[_slice]
-                term_data = pd.DataFrame(x_matrix[:, _slice], columns=cols)
+                term_data = pd.DataFrame(np.asfortranarray(x_matrix[:, _slice]), columns=cols)
```
ahartikainen: are you sure this does what you think it does?

aloctavodia: I am sure this fix the problem. I am not sure why. I found the "solution" while working with splines in PyMC3, and using patsy to generate the design matrix.

PR #189, opened 2020-02-22
change default sd prior and update docs

aloctavodia:
Changing the prior seems to reduce problems during initialization. Still there are additional problem that I will explore in a future PR (it seems `init=jitter+adapt_diag` give trouble while `init=adapt_diag` does not).

I am also updating the notebooks to reflect this and other changes from previous PRs.


twiecki:
Should also rename to `sigma`.

aloctavodia:
I agree, but it should probably be a separated PR. Should we use greek?

twiecki:
Sure that sounds good to me.

PR #202, opened 2020-03-30
Fix double uppercase letters in docs

cpfiffer:
Not sure if this is intentional or not -- feel free to close if it was intentional.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,4 +1,4 @@
-BAyesian Model-Building Interface (Bambi) in Python
+Bayesian Model-Building Interface (Bambi) in Python
 ===================================================
 |Build|
 |Coverage|


ahartikainen:
BAyesian Model-Building Interface
BA.... M.... B.... I...

ColCarroll:
Its a good point that maybe the name should have been `BAMBI`. What if we wrote **Ba**yesian **m**odel-**b**uilding **i**nterface, so the letter casing at least matches?

cpfiffer:
Sorry for the noise, didn't even see the header at the top of the README! I'll sit in my room, now.

aloctavodia:
@cpfiffer do not worry! I think @ColCarroll have a good point.

PR #220, opened 2020-07-03
Add black style check on lint

### LINKED ISSUE #221: Let Travis check for black format.

camenpihor:
fix #221 

## Changes

1. I noticed that we say the library uses the black code style in the readme, and that the code is already black-ified so I added a black check to the `lint` script
    - I excluded tests from the check, because of personal preference, but happy to run the black checker on the tests too, if you want? (although they don't pass right now)
    - I think the code was already set to line length 100, so I used that, but happy for this to be whatever (I use 90 personally)
2. Call the lint script from the test script

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
-
 set -ex # fail on first error, print commands
+
+echo "Checking code style with black...."
+python -m black bambi --line-length=100  --target-version=py37 --exclude=tests/ --check
+echo "Success!"
 
 echo "Checking code style with pylint..."
 python -m pylint bambi/
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -2,7 +2,7 @@
 set -ex # fail on first error, print commands
 
 echo "Checking code style with black...."
-python -m black bambi --line-length=100  --target-version=py37 --exclude=tests/ --check
+python -m black bambi --check
 echo "Success!"
 
 echo "Checking code style with pylint..."
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-Bambi
-------
+# Bambi
+
 BAyesian Model-Building Interface in Python
 
 [![Build Status](https://travis-ci.org/bambinos/bambi.svg?branch=master)](https://travis-ci.org/bambinos/bambi)
@@ -25,7 +25,6 @@ Alternatively, if you want the bleeding edge version of the package you can inst
 ### Dependencies
 
 Bambi requires working versions of numpy, pandas, matplotlib, patsy, pymc3, and theano. Dependencies are listed in `requirements.txt`, and should all be installed by the Bambi installer; no further action should be required.
-
 
 ## Documentation
 
--- a/bambi/backends/__init__.py
+++ b/bambi/backends/__init__.py
@@ -1,6 +1,4 @@
 from .pymc import PyMC3BackEnd
 
 
-__all__ = [
-    'PyMC3BackEnd'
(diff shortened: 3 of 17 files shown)


canyon289:
This just installs black. Is it intended that a check will be performed as well?

scripts/lint.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
-
 set -ex # fail on first error, print commands
 
+echo "Checking code style with black...."
+python -m black bambi --line-length=100  --target-version=py37 --exclude=tests/ --check
```
canyon289, line 5: I would suggest adding a pyproject.toml file. Otherwise what will happen is someone will run black locally but not know the line length and black will then fail in CI

https://github.com/arviz-devs/arviz/blob/master/pyproject.toml

camenpihor: Nice! Didn't know about that. So many different ways to configure python tooling haha. 

Will add this now, thank you :)

canyon289: Np! This is what code reviews are for.

I personally would include tests in the check because other people have to read them. Its like writing a book where half the code is formatted nicely

and 
the
   other
half is
            formatted
terribly

Why not just format everything that has to be read consistently?
Not required but that's just my opinion

camenpihor: 4b25a2a

camenpihor: Gah! I didn't see you other comment until now. Happy to include tests

camenpihor: 4cf381c

canyon289, line 5: Getting there! I would suggest this for this line, as the parameters are coming from the toml file now, and we do want checks on the tests. I would add the py37 option to the toml file as well

```suggestion
python -m black bambi --check
```

camenpihor: Yea, I was having trouble adding the target version to the toml file, so I left it out.. 

```
Error: Could not open file /Users/camen/Documents/bambi/pyproject.toml: Error reading configuration file: invalid literal for int() with base 0: 'py37' (line 3 column 1 char 31)
```

re: changing `lint.sh`. I thought about changing it, but thought that it might be good to leave the entire declaration as its acting as self-documentation (instead of relying on the user to know that they have to look at pyproject.toml for what black is actually running)

re: exclude=tests/ - yea that's my b, forgot to remove...

Do you have any thoughts on the first 2 points?

camenpihor: This is the `pyproject.toml` file that produces that error

```
[tool.black]
line-length = 100
target-version = py37
```

I tried removing the space, putting in quotes, just having 37, and reading black's documentation https://black.readthedocs.io/en/stable/pyproject_toml.html

camenpihor: hahaha seconds later I have found it.... 

https://github.com/psf/black/blob/master/pyproject.toml

Needs to be a list apparently

camenpihor: So now the only thing to resolve is: what do you think about leaving the invocation in `scripts/lint.sh` as is? I also noticed that arviz also includes the parameters defined in the pyproject.toml (but that could just be an oversight)

canyon289: ah that is an oversight. For using lint.sh or using it directly whatever you prefer. I have no opinion

camenpihor: Sounds good! Thanks for the review :)

camenpihor: 279999d

camenpihor:
Thanks for the review! I'll merge after tests pass

PR #238, opened 2020-10-05
Use pm.sample_prior_predictive function

aloctavodia:
Call `pm.sample_prior_predictive` when plotting Bambi's model prior with `model.plot()`

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -713,9 +713,10 @@ class Model:
         return prior
 
     def plot(self, samples=5000, var_names=None):
+        warnings.warn("plot will be deprecated, please use plot_priors", FutureWarning)
         return self.plot_priors(samples, var_names)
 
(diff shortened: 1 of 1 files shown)


tomicapretto:
I would suggest the following two changes:

* Add a `FutureWarning` to `Model.plot()` mentioning that it has been replaced by `Model.plot_priors()`. I agree the name you suggest is much better than the one currently used.
* Change line https://github.com/aloctavodia/bambi/blob/plot_prior/bambi/models.py#L547 

```python
pred_data = pred_data[:, None]  # Must be 2D later
```

To 

```python
pred_data = pred_data.to_numpy()
pred_data = pred_data[:, None]  # Must be 2D later
```

So we avoid the warning mentioned in #240

aloctavodia:
I agree with the first change, but the second one should be done in a separated PR as in not related to plotting priors.

Issue #240, opened 2020-10-06
Convert to numpy array before indexing


tomicapretto:
Right now, when you fit a model with Bambi you' see 

```
FutureWarning: Support for multi-dimensional indexing (e.g. `obj[:, None]`) is deprecated and will be removed in a future version.  Convert to a numpy array before indexing instead.
  pred_data = pred_data[:, None]  # Must be 2D later
```

tomicapretto:
Closing because it is fixed in #242

PR #242, opened 2020-10-06
Fix FutureWarning: Support for multi-dimensional indexing (e.g. `obj[:, None]`) is deprecated

tomicapretto:
Fix in response to #240


bambi/models.py
```diff
@@ -400,6 +401,10 @@ def add(
         NA_handler = Custom_NA(dropna=self.dropna)
 
         # screen fixed terms
+        # it deletes everythin between [] and the brackets too.
```
aloctavodia, line 404: everything

bambi/models.py
```diff
@@ -400,6 +401,10 @@ def add(
...
+
+        # Q: What does it add? None of the `dmatrices()` or `dmatrix` outputs are assigned to a name
```
aloctavodia, line 406: this looks more like a personal note... maybe better to open an issue?

bambi/models.py
```diff
@@ -686,7 +692,7 @@ def _set_priors(self, priors=None, fixed=None, random=None, match_derived_names=
                             targets[term.name] = prior
                     else:
                         targets[name] = prior
-
+        # If prior is None? I think it is handled by another method.
```
aloctavodia, line 695: also looks like a personal note

Issue #256, opened 2020-10-19
Remove Bayesian label from examples


aloctavodia:
All the [examples](https://bambinos.github.io/bambi/examples.html) names are Bayesian "something" which is redundant in a Bayesian library :-)

tomicapretto:
Closing because of #261

PR #261, opened 2020-10-30
Remove 'Bayesian' and 'example' label from examples

tomicapretto:
This PR addresses #256. I've kept the "Bayesian" label for the two workflow examples because I think it would make them easier to be found if someone googles something like "Bayesian workflow example".  

I've also deleted "Example" labels because these notebooks already live under an "Examples" folder. 

Let me know if you would introduce changes.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -4,9 +4,9 @@ Examples
 .. toctree::
   :maxdepth: 1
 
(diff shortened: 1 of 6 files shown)


aloctavodia:
Great! https://github.com/bambinos/bambi/blob/master/docs/examples.rst also needs to be updated.

Issue #230, opened 2020-07-30
Revisit Automatic Priors in Bambi and possibly improve automatic priors


canyon289:
Adding an issue for revisiting the Automatic Priors in Bambi and double checking they still make sense.

Issue ticket spurred from this comment
![image](https://user-images.githubusercontent.com/7213793/88959736-5186d500-d257-11ea-858d-0c3affd432c9.png)

jake-westfall:
Feel free to reach out if you have questions about the current implementation of default priors. I'm not too interested at this time in spending time contributing new code myself, but happy to at least try and remember what we were thinking when we did it :)

Looking back on this work on default priors 3-4 years later, here's my top of head thoughts:
- For the fixed intercept and slope parameters, the current approach still seems pretty good. The underlying theory ([from the arxiv paper](https://arxiv.org/abs/1702.01201)) still seems reasonable to me and, practically speaking, the approach seems to yield pretty sensible "weakly informative" defaults for these parameters.
- For the random effect variances, from a purely practical perspective the defaults on these seem to turn out sensible enough in all the examples I've seen, but the underlying justification is a bit lacking. I mean, there's not really any explicit theoretical argument in my paper at all, just a vague intuitive argument that it seems like a reasonable way to do it. So this can probably be improved, but I don't immediately know the best way. At this point you guys probably know more than me about what good alternative options are.
- For the residual variance in Normal models, in retrospect I'm not real satisfied with how we handled these. My main concern is that I no longer buy the argument in my paper that we can deduce the upper bound on this parameter (for a uniform prior) from the observed data. As for the flat shape itself, I think this choice was basically equivocation on my part due to my uncertainty about what an appropriate shape for this prior should be. On one hand, since it's a variance parameter, it seems natural to want a distribution with nonnegative support and with decreasing density as you move away from 0 (e.g., exponential distribution). On the other hand, smaller values of the residual variance correspond to larger values of R^2 for the regression, so a prior on the residual variance concentrated around 0 seems to imply a prior on R^2 concentrated around 1, which doesn't seem like a very sensible default for most situations -- it seems like you'd probably want a prior on R^2 to have most density closer to 0, which would imply a prior on the residual variance with most density closer to var(Y), i.e., decreasing rather than increasing density as you move down toward 0. So I'm not really sure. Anyway, I see that you already changed this prior from uniform to half-t some time ago, which I think seems fine in practice and likely less surprising for users.
[...]

Issue #164, opened 2019-08-30
boundary problem in logistic regression


cvitlanko:
bambi version 0.1.5

I am trying to use bambi to make a logistic regression. This is a part of my data if you need to reproduce it 

There is variable P, A and case-control that I try to predict. case_control ~ A works, but  case_control ~ P don't. 

```
d= {'P': {126: 2.251033653,
  127: 1.995079298,
  17: 1.22368089,
  15: 1.481352648,
  117: 3.199034487,
  69: 1.926201111,
  84: 1.881285966,
  128: 2.621376675,
  92: 1.770867795,
  64: 1.363825591,
  109: 2.258819785,
  66: 1.761265558,
  96: 2.037313202,
  88: 1.679958935,
  98: 2.217981983,
  55: 1.531942868,
  86: 2.153711787,
  42: 1.538899693,
  70: 1.94275202,
  13: 1.34242365,
  67: 1.80189275,
  125: 1.751800707,
  37: 1.213361474,
  61: 1.3643257830000002,
  113: 2.205566148,
  21: 1.6151316759999998,
  11: 2.562939623,
  121: 1.933595486,
  50: 1.5240257330000002,
  111: 1.75698794},
 'A': {126: 501.1194419,
  127: 601.3105828,
  17: 655.9846374,
  15: 585.0602702000001,
  117: 818.3772901000002,
  69: 585.5345570000002,
  84: 299.4704954,
  128: 727.6092216,
  92: 446.6743807,
  64: 390.0204034,
  109: 540.5108956,
  66: 571.1590227,
  96: 486.3210599,
  88: 381.1089036,
  98: 404.7017446,
  55: 487.0642407,
  86: 311.3404661,
  42: 433.415616,
  70: 788.4081388999998,
  13: 714.7701302999999,
  67: 790.5341929,
  125: 521.2274381,
  37: 433.3870228,
  61: 519.7966038,
  113: 563.4194942,
  21: 458.4685372,
  11: 1031.139464,
  121: 466.5758691,
  50: 586.6346817000001,
  111: 498.7962051},
 'case_control': {126: 1,
  127: 1,
  17: 0,
  15: 0,
  117: 1,
  69: 1,
  84: 1,
  128: 1,
  92: 1,
  64: 0,
  109: 1,
  66: 1,
  96: 1,
  88: 1,
  98: 1,
  55: 0,
  86: 1,
  42: 0,
  70: 1,
  13: 0,
  67: 1,
  125: 1,
[...]

PR #265, opened 2020-11-03
Change start_params to None in call to GLM.fit_constrained used to get automatic priors

tomicapretto:
As mentioned in https://github.com/bambinos/bambi/issues/230#issuecomment-720718643, using start_params equal to unconstrained MLE was causing the following error from statsmodels

`ValueError: The first guess on the deviance function returned a nan.  This could be a boundary  problem and should be reported.`

This error was also reported in #164

### DIFF
--- a/bambi/priors.py
+++ b/bambi/priors.py
@@ -248,7 +248,7 @@ class PriorScaler:
             null = [
                 GLM(
                     endog=self.model.y.data, exog=exog, family=self.model.family.smfamily()
-                ).fit_constrained(str(exog.columns[i]) + "=" + str(val), start_params=None)
+                ).fit_constrained(str(exog.columns[i]) + "=" + str(val))
                 for val in values[:-1]
             ]
             null = np.append(null, full_mod)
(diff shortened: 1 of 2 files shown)


bambi/priors.py
```diff
@@ -248,9 +248,7 @@ def _get_slope_stats(self, exog, predictor, sigma_corr, full_mod=None, points=4)
             null = [
                 GLM(
                     endog=self.model.y.data, exog=exog, family=self.model.family.smfamily()
-                ).fit_constrained(
-                    str(exog.columns[i]) + "=" + str(val), start_params=None
-                )
+                ).fit_constrained(str(exog.columns[i]) + "=" + str(val), start_params=None)
```
aloctavodia, line 251: `start_params` is None by default we can omit it.

tomicapretto: You're right! :man-facepalming:

aloctavodia:
Ready to merge after updating the changelog

PR #272, opened 2020-11-14
Update examples

tomicapretto:
This PR update example notebooks to include changes applied to Bambi lately (for example, using `Model.plot_priors()` instead of `Model.plot()` or calling the package Bambi instead of any other alternative). Also, some concerns were raised when we were analyzing plots of marginal posteriors and we thought it would be good to include comments within these example notebooks. Finally, I'm also trying to write models in their mathematical form within each notebook.

**Update status**

* [x] `ESCS_multiple_regression.ipynb`
* [x] `logistic_regression.ipynb`
* [x] `model_comparison.ipynb`
* [x] `multi-level_regression.ipynb`
* [x] `shooter_crossed_random_ANOVA.ipynb`
* [x] `Strack_RRR_re_analysis.ipynb`
* [x] `t-test.ipynb`


tomicapretto:
I've finished the review of all the example notebooks. **However**, since #275 I had to pass `var_names` argument to `Model.plot_priors()` in places it should not go (I expect Bambi automatically selects the correct parameters to plot). I think we can update that part once we fix the issue, but I leave this open to see if anyone has a better idea.

Issue #275, opened 2020-11-25
plot_priors() plot priors for group-level effects


tomicapretto:
Bambi used to omit group-level effects when plotting priors (only showing priors for the SD of the group-level intercept/slope). After last updates, Bambi plot all priors when calling `Model.plot_priors()`. I think we should modify it to work like it used to do before.

tomicapretto:
In addition, Bambi is not consistent when labeling with strings or index. Sometimes it uses index (or nothing) and sometimes strings (I think this is related to #224). For example

```
subj_model.random_terms
```
```
{'1|subject': <bambi.models.RandomTerm at 0x7fc2d8e27340>,
 'C(race, Sum)[S.black]|subject': <bambi.models.RandomTerm at 0x7fc2d8e27f10>,
 'C(object, Sum)[S.gun]|subject': <bambi.models.RandomTerm at 0x7fc2d8e54df0>,
 'C(race, Sum)[S.black]:C(object, Sum)[S.gun]|subject': <bambi.models.RandomTerm at 0x7fc2d8e54460>}
```
And 

```
subj_model.fixed_terms
```

```
{'Intercept': <bambi.models.Term at 0x7fc2db07b490>,
 'C(race, Sum)': <bambi.models.Term at 0x7fc2d8e54eb0>,
 'C(object, Sum)': <bambi.models.Term at 0x7fc2d8e27370>,
 'C(race, Sum):C(object, Sum)': <bambi.models.Term at 0x7fc2d8e271c0>}
```

These two issues may be addressed at the same time since are related.

aloctavodia:
Agreed. We should also remove the "offset variables" as these are just a reparametrization trick.

tomicapretto:
I've figured why we have numeric index for fixed effects and string index for group-level effects. 

The problem is that a categorical variable that represents a fixed effect corresponds to only one `Term` that has a `data` argument with several columns (as many columns as levels - 1) while a categorical variable that represents a varying effect is splitted in several `RandomTerm`s (as many terms as levels - 1) as if each level was a different term. I think it is possible to overcome this problem and make Bambi work with string index for all terms, but in the future I would like to have a cleaner implementation.

PR #276, opened 2020-11-25
Omits ploting group-level effects and offset variables

aloctavodia:
This partially address #275. Additionally, this also removes the offset variables from the posterior.


oguzhanogreden:
Thanks for this! 🙏

Issue #271, opened 2020-11-13
Bambi fails when you do not indicate which response level you want to model


tomicapretto:
Logistic regression (and similar models) don't work when outcome is categorical and you don't specify which level you want to model

For example

```python
import bambi as bmb
import numpy as np
import pandas as pd

data = pd.DataFrame({
    'x' : np.random.normal(size=100),
    'y' : np.random.choice(['Yes', 'No'], 100)
})

model = bmb.Model(data)
model_fitted = model.fit('y ~ x', family='bernoulli')
```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-16-bbf9470f932f> in <module>
      9 
     10 model = bmb.Model(data)
---> 11 model_fitted = model.fit('y ~ x', family='bernoulli')

~/anaconda3/envs/bmb/lib/python3.8/site-packages/bambi/models.py in fit(self, fixed, random, priors, family, link, run, categorical, backend, **kwargs)
    334         if run:
    335             if not self.built or backend != self._backend_name:
--> 336                 self.build(backend)
    337             return self.backend.run(**kwargs)
    338 

~/anaconda3/envs/bmb/lib/python3.8/site-packages/bambi/models.py in build(self, backend)
    159         # loop over the added terms and _add() them
    160         for term_args in self.added_terms:
--> 161             self._add(**term_args)
    162 
    163         # set custom priors

~/anaconda3/envs/bmb/lib/python3.8/site-packages/bambi/models.py in _add(self, fixed, random, priors, family, link, categorical)
    480                 else:
    481                     # use Y as-is
--> 482                     self._add_y(y_label, family=family, link=link)
    483             else:
    484                 x_matrix = dmatrix(fixed, data=data, NA_action="raise")

~/anaconda3/envs/bmb/lib/python3.8/site-packages/bambi/models.py in _add_y(self, variable, prior, family, link, *args, **kwargs)
    620 
    621         data = kwargs.pop("data", self.clean_data[variable])
--> 622         term = Term(variable, data, prior=prior, *args, **kwargs)
    623         self.y = term
    624         self.built = False

~/anaconda3/envs/bmb/lib/python3.8/site-packages/bambi/models.py in __init__(self, name, data, categorical, prior, constant)
   1005         # dummy codes), which receive special priors
   1006         if constant is None:
-> 1007             self.constant = np.atleast_2d(data.T).T.sum(1).var() == 0
   1008         else:
[...]

Issue #224, opened 2020-07-18
[Feature Request] Factor strings instead of index


canyon289:
It would be nice if bambi could include factor/contrast label in the names like stan_glm does

```
from bambi import Model
import pandas as pd


earnings = pd.read_csv("https://raw.githubusercontent.com/avehtari/ROS-Examples/master/Earnings/data/earnings.csv")
earnings.head()

earnings["c_height"] = earnings["height"] - 66
model = Model(earnings[na_filter])
fit_4 = model.fit('weight ~ c_height + male + C(ethnicity)', samples=1000, chains=4)
```

![image](https://user-images.githubusercontent.com/7213793/87844050-f83d9f80-c86d-11ea-8ba4-8d905550ab2f.png)
![image](https://user-images.githubusercontent.com/7213793/87844058-0ab7d900-c86e-11ea-8b6f-25716dc0bc6e.png)

tomicapretto:
This has been partially addressed in #277 and now prior/posterior plots have better labels. However, summary tables still suffer from this problem. Just leaving the note for the record and keeping the issue open.

tomicapretto:
Isn't this issue covered [here](https://github.com/arviz-devs/arviz/commit/226d1418a959529f887c627dc751f6627bdbd345)?

aloctavodia:
closing, feel free to open if not yet solved

PR #277, opened 2020-11-27
Refactor how variables are added to the model object

tomicapretto:
Currently, `Model._add()` is a long method that does a lot of stuff and I think it that a first sight it not clear what is going on. In this PR I'm adding a variety of changes to make `Model._add()` clearer as well as solving #275, #271 and (maybe) #224

A list of (planned) changes

* [x] Split `Model._add()` into `Model._add_fixed()` and `Model._add_random()`
* [x] Added `get_bernoulli_data()` to `utils.py`. This function is called to handle the response variable when `family='bernoulli'`.
* [x] Add more tests. For example, check Bambi works when no response level is explicitly passed and `family='bernoulli'`.
* [x] Added `ResponseTerm` class, see comments.
* [x] Add tests for new plot labels

I deleted this point 

* [Use `Term` and `RandomTerm` consistently (see https://github.com/bambinos/bambi/issues/275#issuecomment-734498678)

Because (as mentioned below) I think it should be part of another refactor.

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -199,6 +199,7 @@ class Model:
 
             x_matrix = [pd.DataFrame(x.data, columns=x.levels) for x in terms]
             x_matrix = pd.concat(x_matrix, axis=1)
+
             self.dm_statistics = {
                 "r2_x": pd.Series(
                     {
@@ -254,14 +255,12 @@ class Model:
             scaler = PriorScaler(self, taylor=taylor)
             scaler.scale()
 
-        # For bernoulli models with n_trials = 1 (most common use case),
-        # tell user which event is being modeled
-        if self.family.name == "bernoulli" and np.max(self.y.data) < 1.01:
-            event = next(i for i, x in enumerate(self.y.data.flatten()) if x > 0.99)
+        # Tell user which event is being modeled
+        if self.family.name == "bernoulli":
             _log.info(
                 "Modeling the probability that %s==%s",
                 self.y.name,
-                str(self.clean_data[self.y.name].iloc[event]),
+                str(self.y.success_event),
             )
 
         self._set_backend(backend)
@@ -506,16 +505,25 @@ class Model:
         if self.family.name == "gaussian":
             prior.update(sigma=Prior("HalfStudentT", nu=4, sigma=self.clean_data[variable].std()))
 
+        # Success event when family = 'bernoulli'
+        success_event = None
+        categorical = False
+
         if event is not None:
+            if self.family.name != "bernoulli":
+                raise ValueError("Index notation only available for 'bernoulli' family")
             # pass in new Y data that has 1 if y=event and 0 otherwise
-            data = vector[:, vector.design_info.column_names.index(event.group(1))]
+            success_event = event.group(1)
+            categorical = True
+            data = vector[:, vector.design_info.column_names.index(success_event)]
             data = pd.DataFrame({event.group(3): data})
         else:
             data = self.clean_data[variable]
             if self.family.name == "bernoulli":
-                data = get_bernoulli_data(data)
+                categorical = True
+                data, success_event = get_bernoulli_data(data)
 
-        self.y = Term(variable, data, prior=prior)
+        self.y = ResponseTerm(variable, data, categorical, prior, success_event=success_event)
         self.built = False
 
     def _add_fixed(self, fixed, data, family, link, priors):
@@ -1021,6 +1029,18 @@ class Term:
             self.constant = constant
 
         self.prior = prior
+
+
+class ResponseTerm(Term):
+    def __init__(self, name, data, categorical=False, prior=None, success_event=None):
+        super().__init__(name, data, categorical, prior)
+        self.success_event = str(success_event)
+        self.clean_event()
+
+    def clean_event(self):
+        event = re.search(r"\[([\S+]+)\]", self.success_event)
+        if event is not None:
+            self.success_event = event.group(1)
 
 
 class RandomTerm(Term):
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -17,7 +17,7 @@ import bambi.version as version
 from .backends import PyMC3BackEnd
 from .external.patsy import Custom_NA
 from .priors import Prior, PriorFactory, PriorScaler
-from .utils import listify
+from .utils import listify, get_bernoulli_data
 
 _log = logging.getLogger("bambi")
 
@@ -178,8 +178,7 @@ class Model:
         if backend is None:
             if self._backend_name is None:
                 raise ValueError(
-                    "Error: no backend was passed or set in the "
-                    "Model; did you forget to call fit()?"
+                    "No backend was passed or set in the Model; did you forget to call fit()?"
                 )
             backend = self._backend_name
 
@@ -200,7 +199,6 @@ class Model:
 
             x_matrix = [pd.DataFrame(x.data, columns=x.levels) for x in terms]
             x_matrix = pd.concat(x_matrix, axis=1)
-
             self.dm_statistics = {
                 "r2_x": pd.Series(
                     {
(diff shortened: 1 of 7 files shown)


twiecki (approved):

twiecki:
@tomicapretto This looks like a great contirbution and useful refactoring.

Note that this project requires black:
```
hecking code style with black....

+python -m black bambi --check

would reformat /home/travis/build/bambinos/bambi/bambi/tests/test_built_models.py

Oh no! 💥 💔 💥
```

tomicapretto:
I've just added a new class `ResponseTerm` that inherits from `Term` that is only used for the response variable. I think the information about the level that is being modelled should be saved in that object and not inferred as it was done before, i. e.

```python
        # For bernoulli models with n_trials = 1 (most common use case),
        # tell user which event is being modeled
        if self.family.name == "bernoulli" and np.max(self.y.data) < 1.01:
            event = next(i for i, x in enumerate(self.y.data.flatten()) if x > 0.99)
            _log.info(
                "Modeling the probability that %s==%s",
                self.y.name,
                str(self.clean_data[self.y.name].iloc[event]),
            )
```

Now it is 
```python
        # Tell user which event is being modeled
        if self.family.name == "bernoulli":
            _log.info(
                "Modeling the probability that %s==%s",
                self.y.name,
                str(self.y.success_event),
            )
```

Also, currently all response variables have `categorical` equal to `False`. Now, if we are in a model where `family='bernoulli'`, the response term has `True` for `categorical`. Finally, Bambi raises an early error when try to use index notation with other family than 'bernoulli'.


I haven't discussed this changes with anyone before this, so please let me know if anything in this PR does not make sense and we discuss about the approach to follow.

tomicapretto:
Some updates: 

1. I could get better labels for categorical variables that are used as fixed predictors as well as random terms.
Have a look at the following:

```python
import arviz as az
import bambi as bmb
import numpy as np
import pandas as pd

np.random.seed(1234)
az.style.use('arviz-darkgrid')

data = pd.DataFrame({
    'y' : np.random.normal(size=50),
    'x' : np.random.normal(size=50),
    'z' : ['Group 1'] * 10 + ['Group 2'] * 10 + ['Group 3'] * 10 + ['Group 1'] * 10 + ['Group 2'] * 10,
    'time': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10] * 5,
    'subject': ['aaa'] * 10 + ['bbb'] * 10 + ['ccc'] * 10 + ['ddd'] * 10 + ['eee'] * 10
})
model = bmb.Model(data)
fitted = model.fit('y ~ x + z + time', random=['time|subject'])

model.plot_priors();
```
priors

```python
az.plot_trace(fitted);
```
posteriors

2. While it helps with the plots, it does not help with the summary tables. It looks like some information is lost in the middle. I haven't figured out how to fix this yet.

```python
az.summary(fitted)
```

image

3. I still need to add tests. 

Finally, while what I've done seems to work,  there may be better ways. Please let me know if anything, either efficiency or style,  could be improved.

bambi/models.py
```diff
@@ -615,14 +500,138 @@ def _add_y(self, variable, prior=None, family="gaussian", link=None, *args, **kw
...
+        categorical = False
+
+        if event is not None:
+            if self.family.name != "bernoulli":
+                raise ValueError("Index notation only available for 'bernoulli' family")
+            # pass in new Y data that has 1 if y=event and 0 otherwise
+            success_event = event.group(1)
+            categorical = True
+            data = vector[:, vector.design_info.column_names.index(success_event)]
+            data = pd.DataFrame({event.group(3): data})
```
ravinkumar, line 519: What is group 3? Can a comment be added here for clarification?

ravinkumar (approved):

bambi/models.py
tomicapretto: It is already written in `_add_fixed()` method. 

```python
    def _add_fixed(self, fixed, data, family, link, priors):
        if "~" in fixed:
            # check to see if formula is using the 'y[event] ~ x' syntax.
            # If so, chop it into groups:
            # 1 = 'y[event]', 2 = 'y', 3 = 'event', 4 = 'x'
            # If this syntax is not being used, event = None
            event = re.match(r"^((\S+)\[(\S+)\])\s*~(.*)$", fixed)
```
But I think an reminder there will be helpful. Thanks!!

ravinkumar: Oh I see. my apologies. Treat this comment as optional then

aloctavodia:
What you are seeing with the summary function is related to arviz-devs/arviz#1201

bambi/utils.py
```diff
@@ -7,3 +13,36 @@ def listify(obj):
...
+    else:
+        raise ValueError("Response variable is of the wrong type.")
+    return data, success
+
+
+def extract_label(string, kind="fixed"):
+    """Receives a level as returned by patsy and returns a cleaned version of it.
+    Fixed and random effects are handled differently because their representations differ.
+    """
+    assert kind in ["fixed", "random"], "kind must be 'fixed' or 'random'"
```
aloctavodia, line 38: I think we can safely remove this assertion. and move "kind must be 'fixed' or 'random'" to the docstring

bambi/tests/test_built_models.py
```diff
@@ -421,6 +421,36 @@ def dicts_close(a, b):
     assert all([dicts_close(priors0[x], priors1[x]) for x in priors0.keys()])
 
 
+def test_logistic_regression_empty_index():
+    np.random.seed(303456)
```
aloctavodia, line 425: We can remove the random seeds from these tests

bambi/models.py
```diff
@@ -1020,6 +1037,25 @@ def __init__(self, name, data, categorical=False, prior=None, constant=None):
...
+
+    def clean_levels(self):
+        if self.random:
+            self.levels = [extract_label(level, "random") for level in self.levels]
+        else:
+            self.levels = [extract_label(level, "fixed") for level in self.levels]
+
+
+class ResponseTerm(Term):
+    def __init__(self, name, data, categorical=False, prior=None, success_event=None):
```
aloctavodia, line 1050: add docstring

bambi/models.py
```diff
@@ -615,14 +500,139 @@ def _add_y(self, variable, prior=None, family="gaussian", link=None, *args, **kw
...
+
+            if EvalFactor(_name) in factor_infos:
+                categorical = factor_infos[EvalFactor(_name)].type == "categorical"
+            else:
+                categorical = False
+            prior = priors.pop(_name, priors.get("fixed", None))
+            self.terms[_name] = Term(_name, term_data, categorical=categorical, prior=prior)
+
+    def _add_random(self, random, data, priors):
+        for random_effect in random:  # pylint: disable=too-many-nested-blocks
```
aloctavodia, line 559: I think we no longer need to disable this

tomicapretto:
While everything seems to work (all tests have passed) I think this issue is not completely solved. Some reasons

1. Now, Term-like objects have an attribute `cleaned_levels` and `levels`. The first is used  when creating the PyMC model and the latters is used a lot internally. That's why I had to leave `levels`, because many things depended on how it worked. I think this will be completely solved the day we make a greater refactor.
2. Interaction terms have more or less redundant labels. They appear like "var_a:var_b\n var_a[level_1]:var_b[level_2]. 
3. The tabulates produced with `az.summary()` still has numeric index. I know this is related to ArviZ, but in the end, the problem is not completely solved.

My next step is to add more tests related to all these functionalities and I think we will be able to merge. In the (hopefully, near) future I would like to refactor more things so we don't break much things when we want to introduce other changes in the future.

Any suggestions?

aloctavodia:
option 1, we merge once test pass and prepare a new release.
option 2, we put this on hold, make a release and then work on a more profound refactor.

I think I prefer option 1 as this PR seems to be an improvement, even if it only partially solve the problem.

twiecki:
@aloctavodia I agree, I think this is a solid piece of work that improves the status quo. Delaying it might just cause us to lose momentum on this.

tomicapretto:
I think this is done now. I've been trying to refactor how we add random terms a little but I realized it is more appropriate for a deeper refactor.

twiecki (approved):

twiecki:
Thanks @tomicapretto, this looks great to me.

PR #279, opened 2020-12-10
Replace the use of random and fixed effects terms

### LINKED ISSUE #236: Replace the use of “random effects” and “fixed effects” terms
We need to replace those terms and also add documentation explaining, what we exactly mean by those terms and other terms commonly used for the same. 

We could use [this](https://mc-stan.org/rstanarm/articles/glmer.html) as a template.

We agreed on using "common terms" and "group specific term"

aloctavodia:
Fix #236 (except for notebooks)


tomicapretto:
Are you planning to introduce any more changes? 
I can update the notebooks but first I need this changes merged so I have the new function arguments available.

aloctavodia:
No more changes.

Issue #268, opened 2020-11-04
Use the same color palette for all the example notebooks


tomicapretto:

aloctavodia:
Are we going to use a default style from matplotlib or arviz, something like "arviz-white"? Or some Bambi specific style?

tomicapretto:
I was thinking about arviz-darkgrid. Is there any drawback about using it as default?

aloctavodia:
no drawback

tomicapretto:
Closing since #282

PR #282, opened 2020-12-11
Update color palette and terms in examples

tomicapretto:
This PR complements #279 with an terms update in the examples and responds to #268 with a consistent color palette (we use arviz-darkgrid from ArviZ). Finally, I made sure all the notebooks in the examples have a watermark in the end. 

**Checklist**
* [x] Update terms in example notebooks
* [x] Use consistent color palette 
* [x] Update plots in getting started

In this PR I also correct a plot in multi level modelling example, I didn't use the group specific effects correctly the last time.
Osvaldo pointed out that he has already updated terms in getting started section.


aloctavodia:
The user guide and getting started section are already up to date. Unless you want to double check for mistakes.

tomicapretto:
**Update**

* Removed divergences in `shooter_crossed_random_ANOVA.ipynb` so this addresses #286 too
* Removed all but 2 divergences in RRR example, so this doesn't fully address #285

PR #283, opened 2020-12-12
add python 3.9 to tests

aloctavodia:


tomicapretto:
As far as I know, the problem with the current error in checks
```
bambi/backends/pymc.py:155:16: E1136: Value 'idata' is unsubscriptable (unsubscriptable-object)
bambi/backends/pymc.py:156:16: E1136: Value 'idata' is unsubscriptable (unsubscriptable-object)
```
Is because of an old ArviZ version (although it doesn't mean *old*). I've fixed it by updating ArviZ with github version.

PR #288, opened 2020-12-15
Add argument to optionally keep offsets in InferenceData object returned by Model.fit()

tomicapretto:
Offset variables were removed from the posterior in #276. However, it may be the case that someone may want to keep them. 

This PR adds an argument `keep_offsets` to `Model.fit()` which defaults to `False`.

**Checklist**

* [x] Add `keep_offsets` argument.
* [x] Fix  `omit_vars` documentation.
* [x] Add tests.
* [x] Discuss on existing argument names.

### DIFF
--- a/bambi/tests/test_model_construction.py
+++ b/bambi/tests/test_model_construction.py
@@ -259,7 +259,7 @@ def test_categorical_term():
         assert model.terms[term].categorical is expected
 
 
-def test_keep_offsets_true():
+def test_omit_offsets_false():
     data = pd.DataFrame(
         {
             "y": np.random.normal(size=100),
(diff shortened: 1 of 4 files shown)


tomicapretto:
Some thoughts on existing argument names. Currently we have

* `omit_vars` in `Model.plot_priors()`. This omits offsets and group specific terms.
* `omit_offsets` in `Model.prior_predictive()`. This omits offsets. 

I am happy with `omit_offsets` because it clearly indicates its purpose. But I think `omit_vars` is ambiguous. Also, I think someone would like to plot group specific terms without having to plot offsets. I propose to replace `omit_vars` in `Model.plot_priors()` with

* `omit_offsets`: omit offsets.
* `omit_group_specific`: omit group specific effects.

What are your thoughts on this?

bambi/tests/test_model_construction.py
```diff
@@ -257,3 +257,31 @@ def test_categorical_term():
...
+        {
+            "y": np.random.normal(size=100),
+            "x1": np.random.normal(size=100),
+            "g1": ["a"] * 50 + ["b"] * 50,
+        }
+    )
+    model = Model(data)
+    fitted = model.fit("y ~ x1", group_specific=["x1|g1"], keep_offsets=False)
+    offsets = [v for v in fitted.posterior.dims if "offset" in v]
+    assert offsets == []
```
aloctavodia, line 287: assert not offsets

aloctavodia:
I agree with split omit_var into two arguments. Also we should unify using "omit" or "keep". Both sound ok to me.

aloctavodia (approved):

PR #298, opened 2021-02-13
Github Actions

canyon289:
Switching bambi off of TravisCI which is going to make some change soon


canyon289:
This is almost done except the coveralls upload. I dont think I have permissions to get the upload token but if someone can DM it to me or something that's the last bit that needs to get figured out

canyon289:
ready for review

canyon289:
If this one isnt wanted we can always revert. Merging for now

PR #300, opened 2021-02-23
obtain model representation

### LINKED ISSUE #287: Add option to plot model specification
something like 

`model.graph()`

instead of `pm.model_to_graphiz(model.backend.model)`

We should add python-graphviz as a default requirement.

aloctavodia:
Fix #287


tomicapretto:
If no further changes are expected to be included, I think this can be merged.

Issue #293, opened 2021-01-25
Update getting started section and make it executable.


aloctavodia:

tomicapretto:
Done #336

PR #302, opened 2021-02-24
update getting started section

aloctavodia:
I tried to update this section to reflect the last changes and also turn it into a notebook, still the code is not executable, just text (so this is only a partial fix to #293 ). This also comments out the example notebooks/shooter_crossed_random_ANOVA


docs/notebooks/getting_started.ipynb
```diff
@@ -0,0 +1,414 @@
+{
```
tomicapretto, line 1: <p>In comments <code>data loaded as a pandas DataFrame</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/302/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

tomicapretto, line 1: <p>I'm not sure about the "start adding terms to", since we can't add terms anymore. Now you just call <code>.fit()</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/302/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

tomicapretto, line 1: <p>I see the table as two lines of text which are very messy</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/302/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

tomicapretto, line 1: <p>I think I'm being too picky here, but I think we should add more emphasis to packages like brms and rstanarm instead of nlme (I don't see it being mentioned that much in the R world). I would say "R packages like lme4 or brms" but it's completely fine if you don't want to change it.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/302/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

tomicapretto, line 1: <p>I would also mention that the reduced-rank coding is used by default because a full-rank coding implies a structural multicollinearity between the dummies and the intercept (i.e., the sum of the 4 dummies would be equal to the intercept). Dropping the intercept gives us the freedom to add the extra dummy. However, the next categorical variable we add must have reduced-rank coding. If not, you have the same problem with structural multicollinearity.</p><p> </p><p>And some notes about formulae. By default, when it detects a categorical variable, levels are not sorted according to appearance, but according to an alphabetic ordering (as in R and Patsy). But if you pass a pandas.DataFrame where the categorical variable has a specific order, it is respected.</p><p> </p><p>The paragraph describing group specific effects is perfect.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/302/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

tomicapretto, line 1: <p>Missing parenthesis in "(for full details, see the PyMC3 documentation"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/302/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

tomicapretto, line 1: <p>List being used in <code>results = model.fit("y ~ X1 + X2 + ['1|X3', '1|X4']", priors=priors)</code></p><p> </p><p>And I'm not sure this will work </p><pre class="ql-syntax" spellcheck="false">    # Specify model but don't build/sample just yet 
    model.fit('y ~ X1 + X3 + X4', group_specific='1|X2', run=False) 

    # Specify priors—produces same result as in Example 2 above 
    model.set_priors( 
        {'X1': 0.3}, 
        common=Prior('Normal', sd=100), 
        group_specific='wide' 
    ) 

    # Now sample 
    results = model.fit(draws=5000) 
</pre><p>I'll  try to check later and propose a solution</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/302/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

tomicapretto, line 1: <p>Same problem with table than above</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/302/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

tomicapretto, line 1: <pre class="ql-syntax" spellcheck="false">    # these two calls are equivalent 
    az.plot_trace(results, var_names=['Intercept', 'condition']) 
</pre><p>I think this belongs to a different problem? There's no <code>condition</code> variable  in <code>o + c + e + a + n</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/302/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

tomicapretto: <p>Now it looks well... I don't know what's going on</p>

aloctavodia: <p>I think maybe you can improve this section in a future PR. You are going to do a better job than me.</p>

aloctavodia: <p>thanks</p><p> </p>

aloctavodia: <p>It looks ok on my notebook and when I render the docs locally</p>

tomicapretto: Right, there's no problem in leaving it as it is right now. After all, nothing of what it says is wrong!

tomicapretto: <p>It does not work, but I have a fix for this. I think it's better to add it in a new PR, do you agree?</p>

PR #303, opened 2021-02-25
Specify formula and fit can be performed in two separated steps

tomicapretto:
Now you can do something like 

```python
model.fit("y ~ X1 + X3 + X4 + (1|X2)", run=False) 
results = model.fit(draws=5000) 
```
and it will work

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -81,9 +81,11 @@ class Model:
         self.noncentered = noncentered
         self._backend_name = None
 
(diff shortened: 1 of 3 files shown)


aloctavodia (approved):
With these changes we can now hide the build method from the users.

tomicapretto:
> With these changes we can now hide the build method from the users.

I'm adding that here

PR #308, opened 2021-03-03
 Some changes to API docs

tomicapretto:


aloctavodia:
Is this still WIP?

tomicapretto:
I mainly updated docstrings containing Python code that was not formatted as such. The only thing I haven't formatted as code are strings that represent possible argument values. Do you think this should be formatted as code too (I'm not sure, that's why asking)?

For example

```python
"""
    dists : dict
        Optional specification of named distributions to use as priors. Each key gives the name of
        a newly defined distribution; values are two-element lists, where the first element is the
        name of the built-in distribution to use ('Normal', 'Cauchy', etc.), and the second element
        is a dictionary of parameters on that distribution (e.g., ``{'mu': 0, 'sigma': 10}``).
        Priors can be nested to arbitrary depths by replacing any parameter with another prior
        specification.
"""
```

There '{mu': 0, 'sigma': 10}  is formatted as code while 'Normal' and 'Cauchy' aren't.

aloctavodia:
They should be formatted also as code.

PR #316, opened 2021-03-08
use xarray for postprocessing in model comparison notebook

OriolAbril:
I have modified the model comparison example notebook to take advantage
of xarray for automagic broadcasting and alignment.

The changes basically are to initialize `AGE` as a dataarray instead of doing the `:, None` trick
to then allow xarray to broadcast everything, `:, None`, `expand_dims` and transposes are
no longer needed. I have also removed the stacking which is done under the hood by `plot_hdi`
as I did not find anything that required an explicit stacking of the samples


tomicapretto (approved):

tomicapretto:
Thanks for this nice contribution! It makes the code much clearer.

PR #317, opened 2021-03-15
Easy way to save figure of graph from model

aloctavodia:
Add option to save a figure from model.graph() by passing the name of a file, it returns and unmodified grahviz object. It also allows to easily set the figure format and resolution.


tomicapretto:
Is it possible to pass width and height too? Looks like it always uses the same default values

Issue #323, opened 2021-03-31
Missing priors for `family='negativebinomial'`?


No-Stream:
Hi, I'm attempting to fit with the `negativebinomial` family and running into an error, despite it being listed as a valid family in the docs. I browsed around the codebase and looks like this config file might be lacking prior specs for negative binomial? 
https://github.com/bambinos/bambi/blob/master/bambi/config/priors.json

```/opt/anaconda3/envs/jan/lib/python3.7/site-packages/bambi/models.py in fit(self, formula, priors, family, link, run, categorical, omit_offsets, backend, **kwargs)
    283         if self._design.response is not None:
    284             _family = family.name if isinstance(family, Family) else family
--> 285             self._add_response(self._design.response, family=_family, link=link)
    286 
    287         if self._design.common:

/opt/anaconda3/envs/jan/lib/python3.7/site-packages/bambi/models.py in _add_response(self, response, prior, family, link)
    329         """
    330         if isinstance(family, str):
--> 331             family = self.default_priors.get(family=family)
    332         self.family = family
    333 

/opt/anaconda3/envs/jan/lib/python3.7/site-packages/bambi/priors.py in get(self, dist, term, family)
    189         elif family is not None:
    190             if family not in self.families:
--> 191                 raise ValueError(f"{family} is not a valid family name.")
    192             _f = self.families[family]
    193             prior = self._get_prior(_f["dist"])

ValueError: negativebinomial is not a valid family name.
```

canyon289:
Unfortunately you're right. It looks like that docstring for families is much out of date. Although not the ideal response I wont be able to add Negative Binomial anytime soon. So I have a couple proposals.

For you Jan, perhaps use PyMC3 directly. Itll be more lines of code but it'll get the job done
https://docs.pymc.io/notebooks/GLM-negative-binomial-regression.html.

For the other maintainers can we remove `negativebinomial`, `gama` (per spelling in docs) and `wald`. We also can add `t` distribution since that is specified. In addition to this also just add tests to see if the model compiles? 
https://github.com/bambinos/bambi/blob/master/bambi/priors.py#L177-L178

canyon289:
For the maintainers it seems like this discussion is related https://github.com/bambinos/bambi/issues/310#issuecomment-792245946

aloctavodia:
Thanks @No-Stream for taking the time to report this issue. @canyon289 not clear to me why you propose to remove families instead of fixing the config file so they actually work.

[...]

PR #324, opened 2021-03-31
Add negative binomial to config file

tomicapretto:
See #323 

This is my first time modifying the config.json file. I'm using [this example](https://docs.pymc.io/notebooks/GLM-negative-binomial-regression.html) from PyMC3. Looks like it works fine, but the prior for `mu` is very weird. 

Does anyone know if this is related to the fact that I'm using the following? 
```json
    "negativebinomial": [
      "NegativeBinomial", {
        "alpha": 1
      }
    ]
```

*UPDATE*

This is the example, the priors and the posterior

```python
model = bmb.Model(df_pois)
fitted = model.fit("nsneeze ~ alcohol + nomeds + alcohol:nomeds", family = "negativebinomial")
```
image

image

### DIFF
--- a/bambi/config/priors.json
+++ b/bambi/config/priors.json
@@ -132,7 +132,8 @@
     ],
     "negativebinomial": [
       "NegativeBinomial", {
-        "alpha": 1
+        "alpha": 1,
+        "mu": 1
       }
     ]
   }
--- a/bambi/config/priors.json
+++ b/bambi/config/priors.json
@@ -57,6 +57,16 @@
       ],
       "link": "identity",
       "parent": "mu"
+    },
+    "negativebinomial": {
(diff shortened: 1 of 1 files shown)


canyon289:
Should we add a test? If so I can do it on a follow up PR

tomicapretto:
@canyon289 that would be great! 

I'm not sure what is the usage of the parameters within the distributions in "dists" in `priors.json`. Does anyone know how they affect Bambi?

canyon289:
@tomicapretto I honestly have no idea either but I guess we could do the thing of deleting it and seeing what happens. Ill give it a run on my PR. Merging this for now

tomicapretto:
Sounds good!

PR #327, opened 2021-04-02
Add model compilation test for all families

canyon289:
@tomicapretto added test. Let me know what you think. Happy to make any changes


tomicapretto:
Hi @canyon289, thanks for adding the test, it's concise and clear! 

I was wondering how we could check whether the priors being set are correct (i.e. match priors in config). But it seems something like that is currently happening in `test_priors.py`, so we can maybe revisit that and see if something can be improved (but there's no need to do that in this PR, I'm merging it)

canyon289:
Great suggestions and always happy to make tests better

PR #333, opened 2021-04-12
API change

tomicapretto:
The major change is going to be how `Model` works. The idea is to go from here

```python
model = Model(data)
idata = model.fit("y ~ x", family, priors, link, ...)
```

To here

```python
model = Model("y ~ x", data, family, priors, link, ...)
idata = model.fit()
```
Where arguments passed to `.fit()` are only those related to the fitting process.

This is going to be a large PR because we have to update documentation, tests, examples, and the getting started section.

- [x] Change API
- [x] Update docs
- [x] Update tests
- [x] Update README 
- [x] Update getting started

**EDIT:** 
I'm thinking it could be nice to pass priors to the `.fit()` method so it can be easier to fit the same model with different priors, i.e.

```python
model = Model("y ~ x", data)
idata1 = model.fit(priors = {"x": Prior("Normal", sd=10)})
idata2 = model.fit(priors = {"x": Prior("Normal", sd=100)})
```

But it may have associated drawbacks that I'm not seeing now.


aloctavodia:
If it helps,  we can update the API first and the docs later in another PR.

If we want to pass different priors we should define different models.

tomicapretto:
I can update the getting started section, and an example to our README, which we don't have, and then we could update the example notebooks in another PR

bambi/models.py
```diff
@@ -27,81 +27,148 @@ class Model:
...
-        taylor=None,
+        auto_scale=True,
+        default_priors=None,
         noncentered=True,
+        taylor=None,
     ):
+        # attributes that are set later
+        self.terms = {}
+        self.dm_statistics = None  # build()
+        self._diagnostics = None  # build()
```
aloctavodia, line 99: I think we can safely remove this

bambi/models.py
```diff
@@ -572,7 +583,10 @@ def plot_priors(
         axes: matplotlib axes or bokeh figures
         """
         if not self.built:
```
aloctavodia, line 585: I think we do not need this anymore as a model object is instantiated with both data and formula and thus sampling from the priors and priors predictive distributions should be immediately available.

tomicapretto: I thought the same when applying the changes. But if you do something like 
```python
model = Model("y ~ x", data)
model.set_priors(common=Prior("Normal", sd=10))
```
You need to call `model.build()` before that prior affects the backend. 

Another option is to call `.build()` in `.set_priors()` but what if we set priors several times? That's why the message says `"Call .build() to build the model or .fit() to build and sample from the posterior."`

bambi/models.py
tomicapretto: `self.terms` must be an empty dictionary because `self._add_common` and `self._add_group_specific` assume it exists, i.e. they do `self.terms[name] = Term(name, term, data, prior)` and `self.terms[name] = GroupSpecificTerm(name, term, data, prior)`

tomicapretto: But I can delete `self._diagnostics` and `self.dm_statistics`. The former is not used, the latter is created in `build()`

aloctavodia: Sorry I was talking only about `self._diagnostics`

README.md
```diff
@@ -23,7 +23,48 @@ Alternatively, if you want the bleeding edge version of the package you can inst
...
+### Linear regression
+
+```python
+model = bmb.Model("y ~ x1 + x2 + x3", data)
+fitted = model.fit()
+```
+
+In the first line we create and build a Bambi `Model`. The second line tells the sampler to start
+running and it returns an `InferenceData` object, which can be passed to several ArviZ functions
+such as `az.summary()` to get a summary of model parameters posterior or `az.plot_traces()` to
```
aloctavodia, line 54: maybe "to get a summary of the parameters distribution and sample diagnostics"

README.md
```diff
@@ -23,7 +23,48 @@ Alternatively, if you want the bleeding edge version of the package you can inst
 
 ### Dependencies
 
-Bambi requires working versions of numpy, pandas, matplotlib, patsy, pymc3, and theano. Dependencies are listed in `requirements.txt`, and should all be installed by the Bambi installer; no further action should be required.
+Bambi requires working versions of arviz, formulae, numpy, pandas, pymc3, and statsmodels. Dependencies are listed in `requirements.txt`, and should all be installed by the Bambi installer; no further action should be required.
```
aloctavodia, line 26: "ArviZ, formulae, NumPy, pandas, PyMC3 and statsmodels"

README.md
```diff
@@ -23,7 +23,48 @@ Alternatively, if you want the bleeding edge version of the package you can inst
...
+running and it returns an `InferenceData` object, which can be passed to several ArviZ functions
+such as `az.summary()` to get a summary of model parameters posterior or `az.plot_traces()` to
+visualize them.
+
+### Logistic regression
+
+Here we just add the `family` argument set to `"bernoulli"` to tell Bambi we are modelling a binary
+response. By default, it uses a logit link. We can also use some syntax sugar to specify which event
+we want to model. We just say `g['Yes']` and Bambi will understand we want to model the probability
+of a `"Yes"` response. Bambi will pick a value and inform about it if we don't use this notation.
```
aloctavodia, line 62: "Bambi will pick a value and inform about it if we don't use this notation." not very clear

tomicapretto:
Thanks for the comments @aloctavodia!

aloctavodia (approved):

Issue #331, opened 2021-04-09
Make it easier to change default prior for response variable


tomicapretto:
`Model.fit()` has a `prior` argument but it is not passed to `Model._add_response()` and consequently the `prior` argument in the latter is unused and the default prior is always used. We should make it possible/easier to modify this prior.

Bambi has already had this problem in the past, see #16.

tomicapretto:
A related bug I've just encountered

We can do 
```python
priors = {
    'weight: bmb.Prior('Normal', mu=0, sd=20),
    'height_sigma': bmb.Prior('Uniform', lower=0, upper=50) 
}
model = Model('height ~ weight', data, priors=priors)
```

But we can't do

```python
priors = {
    'weight: bmb.Prior('Normal', mu=0, sd=20),
    'height_sigma': bmb.Prior('Uniform', lower=0, upper=50) 
}
model = Model('height ~ weight', data)
model.set_priors(priors)
# ValueError: No terms in model match height_sigma.
```
And so far it makes sense since the first argument expects priors for the predictors... But we should update it so it is also possible to update the priors related to the distribution of the response that are not part of the linear predictor

PR #335, opened 2021-04-16
Make it possible to change default priors for response variable

tomicapretto:
See #331 

Current changes work for built-in families `"gaussian"` and `"negativebinomial"`. Other built-in families don't have nuisance parameters in the response distribution. There's still work to be done, however.

**Example**
In the past you couldn't pass priors for nuisance parameters in the response distribution. Now you can.
```python
import bambi as bmb
import numpy as np
import pandas as pd

data = pd.DataFrame({
    "y": np.random.randint(3, 10, size=50),
    "x": np.random.normal(size=50)
})
priors = {"sigma": bmb.Prior('Uniform', lower=0, upper=50)}
bmb.Model("y ~ x", data, priors=priors)
```
```
Formula: y ~ x
Family name: Gaussian
Link: identity
Observations: 50
Priors:
  Intercept ~ Normal(mu: 5.96, sigma: 2.02021478)
  x ~ Normal(mu: 0, sigma: 1.72275744)
  sigma ~ Uniform(lower: 1, upper: 50)
```

```python
priors = {"mu": bmb.Prior('Uniform', lower=1, upper=20)}
bmb.Model("y ~ x", data, family="negativebinomial", priors=priors)
```

```
Formula: y ~ x
Family name: Negativebinomial
Link: log
Observations: 50
Priors:
  Intercept ~ Normal(mu: 1.78507048, sigma: 1.08072027)
  x ~ Normal(mu: 0, sigma: 0.63288403)
  mu ~ Uniform(lower: 1, upper: 20)
```

* [x] Priors for parameters in the response distribution can be updated using `Model.set_priors()`.
* [x] Update nuisance parameters from custom families.
* [x] Check name conflicts, i.e. no term has the same name as any nuisance parameter in the response distribution
* [x] Update documentation in `.add_response()`.
* [x] Tests.

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -401,12 +401,12 @@ class Model:
             theano tensor as the sole argument and returns one with the same shape.
         priors : dict
             Optional dictionary with specification of priors for the parameters in the family of
-            the response. Keys are nuisance parameters in the family (i.e. they cannot be equal to
-            family.parent) and values can be an instance of class ``Prior``, a numeric value,
-            or a string describing the width. In the numeric case, the distribution specified in
-            the defaults will be used, and the passed value will be used to scale the appropriate
-            variance parameter. For strings (e.g., ``'wide'``, ``'narrow'``, ``'medium'``, or
-            ``'superwide'``), predefined values will be used.
+            the response. Keys are names of other parameters than the mean in the family
+            (i.e. they cannot be equal to family.parent) and values can be an instance of class
+            ``Prior``, a numeric value, or a string describing the width. In the numeric case,
+            the distribution specified in the defaults will be used, and the passed value will be
+            used to scale the appropriate variance parameter. For strings (e.g., ``'wide'``,
+            ``'narrow'``, ``'medium'``, or ``'superwide'``), predefined values will be used.
         """
         if isinstance(family, str):
             family = self.default_priors.get(family=family)
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -152,11 +152,11 @@ class Model:
             if priors_ and self._design.common:
                 conflicts = [name for name in priors_ if name in self._design.common.terms_info]
                 if len(conflicts) > 0:
-                    _log.info(
+                    _log.warning(
                         "The prior for %s is ambiguous because its name conflicts "
                         "with the name of a parameter in the response distribution.\n"
                         "Consider renaming the term to prevent an unexpected behaviour.",
-                        ", ".join(conflicts)
+                        ", ".join(conflicts),
                     )
             self._add_response(self._design.response, family, link, priors_)
         else:
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -333,6 +333,15 @@ class Model:
             targets.update({name: group_specific for name in self.group_specific_terms.keys()})
 
         if priors is not None:
+            # Update priors related to nuisance parameters of response distribution
+            priors_ = extract_family_prior(self.family, priors)
+            if priors_:
+                # Remove keys passed to the response.
+                for key in priors_:
+                    priors.pop(key)
+                self.response.prior.args.update(priors_)
+
+            # Prepare priors for explanatory terms.
             for k, prior in priors.items():
                 for name in listify(k):
                     term_names = list(self.terms.keys())
@@ -346,6 +355,7 @@ class Model:
                     else:
                         targets[name] = prior
 
+        # Set priors for explanatory terms.
         for name, prior in targets.items():
             self.terms[name].prior = prior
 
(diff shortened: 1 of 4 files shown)


tomicapretto:
The dictionary of priors can also update parameters within custom families now. Does it make sense to create a custom family and then pass a new prior in the prior dict? I don't think so. But this feature prevents breaking the user's code.

Also, I think it will make sense to update priors in a custom family using `.set_priors()`.

What do you think?

Illustration of the update in this commit 

```python
priors = {"sigma": bmb.Prior('Uniform', lower=0, upper=20)}
family = bmb.Family(
    "gaussian",
    bmb.Prior("Normal", mu=0, sigma=bmb.Prior("Normal", mu=50, sigma=5)),
    "identity",
    "mu"
)
bmb.Model("y ~ x", data, family=family, priors=priors)
```
```
Formula: y ~ x
Family name: Gaussian
Link: identity
Observations: 50
Priors:
  Intercept ~ Normal(mu: 6.0, sigma: 1.86264073)
  x ~ Normal(mu: 0, sigma: 1.56236357)
  sigma ~ Uniform(lower: 0, upper: 20)
```

```python
priors = {"mu": bmb.Prior('Uniform', lower=1, upper=20)}
family = bmb.Family(
    "negativebinomial",
    bmb.Prior("NegativeBinomial", alpha=1, mu=bmb.Prior("HalfCauchy", beta = 1)),
    "identity",
    "alpha"
)
bmb.Model("y ~ x", data, family=family, priors=priors)
```

```
Formula: y ~ x
Family name: Negativebinomial
Link: identity
Observations: 50
Priors:
  Intercept ~ Normal(mu: 1.79175947, sigma: 1.08012603)
  x ~ Normal(mu: 0, sigma: 0.56546834)
  mu ~ Uniform(lower: 1, upper: 20)
```

bambi/priors.py
```diff
@@ -103,6 +103,17 @@ def update(self, **kwargs):
             kwargs_[key] = val
         self.args.update(kwargs_)
 
+    def __eq__(self, other):
+        if not isinstance(other, type(self)):
+            return False
+        else:
+            return (
+                self.name == other.name
```
aloctavodia, line 111: what about?

`return self.__dict__ == other.__dict__`

bambi/models.py
```diff
@@ -387,20 +400,38 @@ def _add_response(self, response, prior=None, family="gaussian", link=None):
             defined in the current backend; typically this will include at least ``'identity'``,
             ``'logit'``, ``'inverse'``, and ``'log'``), or a callable that takes a 1D ndarray or
             theano tensor as the sole argument and returns one with the same shape.
+        priors : dict
+            Optional dictionary with specification of priors for the parameters in the family of
+            the response. Keys are nuisance parameters in the family (i.e. they cannot be equal to
```
aloctavodia, line 405: nitpick: Not sure about the term "nuisance parameter", could we use "dispersion parameter"? or some other alternative.

bambi/models.py
```diff
@@ -148,8 +148,17 @@ def __init__(
...
 
         if self._design.response is not None:
-            _family = family.name if isinstance(family, Family) else family
-            self._add_response(self._design.response, family=_family, link=link)
+            priors_ = extract_family_prior(family, priors)
+            if priors_ and self._design.common:
+                conflicts = [name for name in priors_ if name in self._design.common.terms_info]
+                if len(conflicts) > 0:
+                    _log.warning(
+                        "The prior for %s is ambiguous because its name conflicts "
```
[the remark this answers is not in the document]
aloctavodia: and this should probably be an error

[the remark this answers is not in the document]
tomicapretto: Agreed

bambi/models.py
aloctavodia: Yeah, I get the intended use. When dispersion does not apply? maybe we can call it the "~mean" parameter :-P

tomicapretto: Maybe in the negative binomial family? It uses mu and alpha, and alpha plays both in the shape and scale parameter, see [here](https://github.com/pymc-devs/pymc3/blob/d7172c0a1a76301031d1b3b411d00643c416a0c4/pymc3/distributions/discrete.py#L863-L873). But I understand your point. Saying "dispersion" is more intuitive and familiar for most people, but it is not accurate for all GLMs we may implement

aloctavodia: Yes you are right, but in that case you can think of alpha as the overdispersion parameter, right?

aloctavodia: Maybe "other parameters than the mean", is not that bad

tomicapretto: I agree on the last suggestion

PR #337, opened 2021-04-22
Add gamma family

tomicapretto:
In our docs we say that we support the gamma family, but that is not true. When you attempt to pass `family="gamma"` Bambi fails when looking for priors and dists in the `config.json` file. 

In this PR I try to address that issue, including prior and dists information in the `config.json` file. However, there's something I'm not doing well, see the example. I'm not sure if this is a problem with the parameterizations in PyMC3/statsmodels or if it is related to another problem. 

**Example**

```python
import bambi as bmb
import statsmodels.api as sm

data = sm.datasets.scotland.load(as_pandas=True).raw_data
model = bmb.Model("YES ~ COUTAX", data, family="gamma")
```

```
/home/tomas/anaconda3/envs/bmb/lib/python3.8/site-packages/statsmodels/genmod/generalized_linear_model.py:293: DomainWarning: The inverse_power link function does not respect the domain of the Gamma family.
  warnings.warn((f"The {type(family.link).__name__} link function "
/home/tomas/anaconda3/envs/bmb/lib/python3.8/site-packages/statsmodels/genmod/generalized_linear_model.py:293: DomainWarning: The inverse_power link function does not respect the domain of the Gamma family.
  warnings.warn((f"The {type(family.link).__name__} link function "
/home/tomas/anaconda3/envs/bmb/lib/python3.8/site-packages/statsmodels/genmod/generalized_linear_model.py:293: DomainWarning: The inverse_power link function does not respect the domain of the Gamma family.
  warnings.warn((f"The {type(family.link).__name__} link function "
/home/tomas/anaconda3/envs/bmb/lib/python3.8/site-packages/statsmodels/genmod/generalized_linear_model.py:293: DomainWarning: The inverse_power link function does not respect the domain of the Gamma family.
  warnings.warn((f"The {type(family.link).__name__} link function "
/home/tomas/anaconda3/envs/bmb/lib/python3.8/site-packages/statsmodels/genmod/generalized_linear_model.py:293: DomainWarning: The inverse_power link function does not respect the domain of the Gamma family.
  warnings.warn((f"The {type(family.link).__name__} link function "
/home/tomas/anaconda3/envs/bmb/lib/python3.8/site-packages/statsmodels/genmod/generalized_linear_model.py:293: DomainWarning: The inverse_power link function does not respect the domain of the Gamma family.
  warnings.warn((f"The {type(family.link).__name__} link function "
/home/tomas/anaconda3/envs/bmb/lib/python3.8/site-packages/statsmodels/genmod/generalized_linear_model.py:293: DomainWarning: The inverse_power link function does not respect the domain of the Gamma family.
  warnings.warn((f"The {type(family.link).__name__} link function "
/home/tomas/anaconda3/envs/bmb/lib/python3.8/site-packages/statsmodels/genmod/generalized_linear_model.py:293: DomainWarning: The inverse_power link function does not respect the domain of the Gamma family.
  warnings.warn((f"The {type(family.link).__name__} link function "
```

I have also tried with other link functions, and the same situation persists. Then, when you try to run the sampler, it also fails saying `RuntimeError: Chain 0 failed.`


tomicapretto:
I've been trying to replicate [this brms example](https://rpubs.com/jwesner/gamma_glm). 

```python

import arviz as az
import bambi as bmb
import numpy as np
import pandas as pd

N = 1000 # I used more samples
x = np.random.uniform(-1, 1, N)
a = 0.5
b = 1.1
y_true = np.exp(a + b * x)
shape = 10
y = np.random.gamma(shape, y_true / shape , N)
fake_data = pd.DataFrame({"x": x, "y": y})

plt.plot(fake_data["x"], fake_data["y"], "o")

priors = {
    "Intercept": bmb.Prior("Normal", mu=0, sd=2),
    "x": bmb.Prior("Normal", mu=0, sd=2),
    "sigma": bmb.Prior("Gamma", alpha=0.01, shape=0.01)
}
model = bmb.Model("y ~ x", fake_data, family="gamma", priors=priors, link="log")

fitted = model.fit()
az.summary(fitted, kind="stats")
```
And you see something like this

image

Which differs quite a lot from the original output

image

I think we have a conflict with parameterizations, but I cannot be 100% sure.

aloctavodia:
with pymc3.

```
with pm.Model() as model:
    a = pm.Normal("a", 0, 2)
    b = pm.Normal("b", 0, 2)
    s = pm.HalfNormal("s", 1)
    obs = pm.Gamma("obs", mu=np.exp(a+b*x), sigma=s, observed=y)
    idata = pm.sample(return_inferencedata=True)
```

lala

for the "brms fit" I just used the parameters from the capture i.e. y=0.49*1.05*x

the bambi model 

```
priors = {
    "Intercept": bmb.Prior("Normal", mu=0, sd=2),
    "x": bmb.Prior("Normal", mu=0, sd=2)
}
model = bmb.Model("y ~ x", fake_data, family="gamma", priors=priors, link="log")
```
provides virtually the same fit as PyMC3.

tomicapretto:
Both PyMC3 and Bambi are indeed "equal". But what can explain the bias in the parameter estimates? What is the correspondence between the shape parameter and the sigma parameter in PyMC3? Not only asking you but asking myself too.

image

twiecki:
This does look odd. Visually the brms fit looks more correct, especially around -1. Maybe look at the prior and posterior predictive to see if the brms fit is achievable here?

tomicapretto:
So the good news is that we've realized what the problem was. The bad news is that I think models like this one (where one of the parameters of the response distribution depends on the mean being predicted) can't be implemented in Bambi very easily. However, I think it is doable and I'm working on it :)

tomicapretto:
Now, If we use family="gamma" it works. But many things are a mess. I think I'm going to fix #338 so I can see if other link functions, such as log, are properly passed to statsmodels. I also need to make sure default priors are properly scaled, which is not happening right now.

tomicapretto:
Now we are passing non-default links to statsmodels families. Still need to add some tests.

tomicapretto:
Now we can do 

```python
priors = {
    "alpha": bmb.Prior("Uniform", lower=0, upper=50)
}
model = bmb.Model("y ~ x", fake_data, family="gamma", link="log", priors=priors)
model
```
```
Formula: y ~ x
Family name: Gamma
Link: log
Observations: 100
Priors:
  Intercept ~ Normal(mu: 0.71835045, sigma: 0.65645946)
  x ~ Normal(mu: 0, sigma: 0.38011387)
  alpha ~ Uniform(lower: 0, upper: 50)
```
```python
fitted = model.fit()
az.summary(fitted)
```
image

tomicapretto:
I think this PR is done. I plan to open a new PR to work on the Wald family, and another PR to work on link functions. I would like to add probit, cloglog, and Cauchy link functions. I also plan to revisit the negative binomial family, I'm not completely sure it is correctly implemented.

PR #346, opened 2021-05-06
Negative binomial GLM

tomicapretto:
I still want to add a couple of plots showing the predictions of the model and review/modify the current text.


tomicapretto:
@aloctavodia what is your opinion about showing the original fit is not so good, and the proposal of the other two alternatives that follow (using exp and arcsinh transforms).

Btw, the code needs to be cleaned yet, there's too much repetition. And I'm still not explaining how `evaluate_new_data()` works.

tomicapretto:
> @aloctavodia what is your opinion about showing the original fit is not so good, and the proposal of the other two alternatives that follow (using exp and arcsinh transforms).
> 
> Btw, the code needs to be cleaned yet, there's too much repetition. And I'm still not explaining how `evaluate_new_data()` works.

I'm starting to think this example is a better fit for a binary response model. We actually have a fixed number of cells and we record whether each of them differentiates or not. I'm still wondering why the book shows this data in the context of GLMs for count data (we can't even use Poisson approximation to the binomial because we don't have  p -> very small when doses are high)

tomicapretto:
I'm now using a different data set now, which is more suitable for this type of model. The example is done (i think) but I'm leaving this open because I would like to have second opinions.

PR #349, opened 2021-05-11
Use sphinx-multiversion to support documentation for multiple versions

tomicapretto:
I added [sphinx-multiversion](https://holzhaus.github.io/sphinx-multiversion/master/index.html) to support the creation of docs for each release. I have also created a jinja template to put the links for the different versions and also some information. The `gh-pages-redirect.html` refers to a wrong link, I'm going to fix it later.

Illustrative gif

version

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,52 @@
+
+name: Publish documentation
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+
+    # Example of using a custom build-command.
+    - uses: ammaraskar/sphinx-action@master
+      with:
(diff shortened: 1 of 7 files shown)


OriolAbril:
Does this rebuild all the documentation for all versions every time docs are built? I am not sure I understand how it works so that old versions show the links to newer versions, `versions` var looks static once defined/computed from tags in `conf.py` right?

tomicapretto:
There are two projects that use sphinx-multiversion without re-building everything from 0, [here](https://github.com/juaml/julearn) and [here](https://github.com/PandABlocks/PandABlocks-client). But they use a version of the package that is not merged yet ([this PR](https://github.com/Holzhaus/sphinx-multiversion/pull/64)). I'm not sure if we should follow them and do the same here, having to keep an eye close to the development of sphinx-multiversion, or just keep using sphinx and add a GitHub action to automatically build the latest release documentation and upload to gh-pages.

tomicapretto:
I'm merging to test if this works well... Hope not to break anything :pray:

PR #351, opened 2021-05-14
Update t-test example

agustinaarroyuelo:
Updates to the t-test example, like using `plot_posterior` function. We can discuss the approach to rest of the example.


docs/notebooks/t-test.ipynb
```diff
@@ -62,18 +62,99 @@
     },
```
tomicapretto, line 62: <p>There are two equivalent ways of showing the t-test as a linear model. The difference is whether we include the intercept (as usually done in statistical packages, Bambi included) or not. This equivalence wasn't very clear to me when writing this notebook, that's why I used the default encoding that includes the intercept. But, actually, we can explain that the t-test is represented by any of the two following models.</p><p> </p><p>Model 1 (no intercept)</p><p> </p><p>$$</p><p>\mu_i = \beta_i + \epsilon_i </p><p>$$</p><p> </p><p>Each of the \beta_i represents the mean of population $i$. Here you have two betas, one for i = 0 and another for i = 1. </p><p> </p><pre class="ql-syntax" spellcheck="false">Model('y ~ 0 + Group', data)
</pre><p> </p><p>Model 2 (with intercept)</p><p> </p><p>$$</p><p>\mu_i = \beta_0 + \beta_1 (i) + \epsilon_i </p><p>$$</p><p> </p><p>Here you can't include two dummies (one for each group) because the sum of the dummies will be equal to the intercept in the design matrix, resulting in a design matrix with redundancies. The intercept results to be the mean of the group 0, the one taken as the baseline, and \beta_1 is the difference between the mean of the group 1 and the mean of the group 0.</p><p> </p><pre class="ql-syntax" spellcheck="false">Model('y ~ Group', data)
</pre><p> </p><p>These two models are equivalent, they just have different parametrizations. In the first, the means are equal when \beta_0 = \beta_1. In the latter, the means are equal when \beta_1 = 0.</p><p> </p><p>As a side note, the same thing happens if you have a Group variable with more than 2 levels. If you include the intercept, you will have <code>k - 1</code>  dummies, each one representing the deviation between a baseline mean and the mean of the group. If you don't include the intercept, you will have <code>k</code>  dummies, each one representing the mean of the group.</p><p> </p><p>I hope this helps to show the equivalence between the two models and the t-test. But please ask as many times as you want if you need any help.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/351/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

tomicapretto, line 62: <p>Mention the <code>0</code>  removes the intercept. Using <code>-1</code>  is also equivalent.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/351/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

docs/notebooks/t-test.ipynb
```diff
@@ -39,7 +39,7 @@
    "cell_type": "markdown",
```
tomicapretto, line 39: <p>We are using the same introductory sentence for both sections "Model 1" and "Model 2". In this first section, I would say something like "When we carry out a two sample t-test we are implicitly using a linear model that can be specified in different ways. One approach is the following"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/351/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

tomicapretto, line 39: <p>HDI for beta_1 has changed, just update the values.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/351/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

tomicapretto, line 39: <p>And here you can stary saying</p><p> </p><p>"The linear model implicit in the t-test can also be specified without an intercept term" </p><p> </p><p>The words don't need to be those, feel free to choose your own wording. But I think it is important to mark one model uses the intercept and the other doesn't.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/351/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

agustinaarroyuelo:
Thanks Tomás, I will be going through your comments

tomicapretto:
Thank you so much Agus for this PR!! :smile:

PR #353, opened 2021-05-15
Change doc workflow

tomicapretto:


tomicapretto:
Looks like it works now

PR #359, opened 2021-05-26
Fix automatic prior shapes

tomicapretto:
Currently, the code that computes the priors checks whether the sum of the design matrix for a given term is constant to decide whether it is an intercept or another common term. It flags categorical terms with full encoding as intercepts and then their prior has `shape = ()` when they actually have more than one column in the design matrix. 

So far, this PR fixes the shapes of the priors for the common terms. Categorical terms with full encoding aren't handled as intercepts anymore.

 I still need to fix the shapes for group-specific terms, but as far as I've seen, it won't be that simple because I still don't know what is broken.

**Edit**

This PR ended up refactoring much more than what it initially intended. Here's a summary of the changes.

* Removed the `constant` argument and property from `Term` and `GroupSpecificTerm`. It's never used when instantiating and is ambiguous when used as property. Intercept terms are flagged with the `.type` attribute, and cell-mean terms are flagged with `._is_cell_means`.
* `Model.get_pymc_coords()` does not add `*_dim_0` coords anymore. It returns the coordinate values that are used by each term.
* We don't use the `shape` argument anymore when adding PyMC distributions, we use `coords`. Among other things, categorical variables are correctly labeled in ArviZ when they contain only one level (actually it does not contain only one level, the other is taken as reference).
* Now, we add only one term when  `expr` in `(expr|factor)` represents a categorical variable, instead of as many terms as levels in `expr`. The PyMC distribution has two coordinates. The first for the grouping factor `factor` and the second for the levels of the categorical variable `expr`. Coordinates in the InferenceData object are sorted so outputs look like `variable|group[level, group_level]`, rather than `variable|group[group_level, level]`
* Contributions from group-specific terms to the linear predictor are handled a little differently. Now, there's a loop that goes through each of the columns of `term.predictor` when it has more than one column, i.e. when `expr` is a categorical variable with 2 or more levels. This is not the most beautiful implementation but resulted much faster (at least twice in my example) than doing the matrix multiplication between `term.data` and `coef`, where `coef` is the PyMC3 distribution.


tomicapretto:
I think the error in the tests is because one test depends on a development version of ArviZ (which is what I'm using locally)

tomicapretto:
I'm updating some notebooks to reflect changes in labels. They look better now!

**Before**
old

**After**
new

The same thing happens with plots!

zwelitunyiswa:
This looks great. Looks like the PYMC3 summary table. It makes it easier to
interpret.

Issue #320, opened 2021-03-28
Default Priors Bernoulli Family with Perfect Separation


jlevy44:
Hi there! Thanks for this awesome package! I see you're using the MLE estimate to help determine the scaling of the priors. This may fail in the case where there is perfect separation in a logistic regression model, which I had seen on an initial test. If prior is set to none, maybe is it possible to toggle some of the prior rules / default priors from https://mc-stan.org/rstanarm/ or brms, etc. when this happens? 

Thanks!

https://github.com/bambinos/bambi/blob/8f1679810f7cd132ebdd47a1b64ac82e197b63df/bambi/priors.py#L215

tomicapretto:
Hi @jlevy44, thanks for opening this issue! 

This is something we are aware of, and we are starting to do some research on default priors with the aim to improve what is currently being done in Bambi. #278 is also another consequence of our default prior system. 

We've been in touch with some brms/rstanarm devs and they also want to improve their current default priors. So, at least for now, the goal is to do more research on the subject and implement something that improves over Bambi/brms/rstanarm defaults, instead of porting brms/rstanarm priors to Bambi.

If you still want to use rstanarm default priors, [this guide](https://mc-stan.org/rstanarm/articles/priors.html) will be helpful. As far as I've seen in some empirical comparisons between our default priors and rstanarm default priors, you won't see any significant difference between the posteriors _in general_. 

I'm leaving this issue open and will post any update to our default prior system here. Again, thanks for writing!

jlevy44:
Thanks for your response. I am aware of the default priors offered by rstanarm and may try them out using this package. I agree, I think it's great that your group is researching ideal default priors. I usually set my own based on the use case, excited to see what you come up with!

tomicapretto:
Development version now uses priors inspired on rstanarm priors, which should overcome this perfect separation problem with older default priors.

PR #360, opened 2021-06-21
Add alternative default priors

tomicapretto:
This PR aims to add alternative default priors that will be the default priors for those models where we lack statsmodels support. This new prior does not replace existing defaults (until we've evidence they're equivalent). I will be updating the following list of changes as I commit to this PR.

**Changes**

* Remove `Model._match_derived_terms()`. This was unused because categorical group-specific terms are indeed one term in the model and not several terms (as many as dummies in the encoding of the categorical variable) as it used to be before.
* Constant terms (categoricals with one level and numerics with a unique value) are flagged more appropriately.
* Splitted `PriorFactory._get_prior()` into several methods with clearer names and goals. Also modified `config.json` as proposed in #361.
* `Prior._auto_scale` is now `Prior.auto_scale`. It's bothering to add pylint exceptions all the time.
* Nuisance parameters of the response distribution are scaled with the prior scaler and not when the prior term is added.
* Added alternative automatic priors (inspired on rstanarm priors). See `PriorScaler2`.
* Our tests should faster because I removed unnecessary `Model.build()` calls.
* Family names and priors are checked slightly differently, which makes the code simpler.
* Removed methods and attributes that were used when we had multiple backends. Now that we only have PyMC3, it does not make sense to keep asking which backend is being used.
* Added more tests.
*  `Model` has a new argument, `priors_cor`. It accepts dictionaries where keys are the names of the groups, and values are the eta parameter in the LKJ distribution for correlation matrices. If such a dictionary is present, priors for group-specific terms are a multivariate normal distribution with a non-zero correlation.

### DIFF
--- a/bambi/backends/pymc.py
+++ b/bambi/backends/pymc.py
@@ -79,7 +79,7 @@ class PyMC3BackEnd(BackEnd):
 
         ## Add group-specific effects
         # Group-specific effects always have pymc_coords. At least for the group.
-        # The loop through precitor columns is not the most beautiful alternative.
+        # The loop through predictor columns is not the most beautiful alternative.
         # But it's the fastest. Doing matrix multiplication, pm.math.dot(data, coef), is slower.
 
         # Add group specific terms that have prior for their correlation matrix
@@ -89,7 +89,7 @@ class PyMC3BackEnd(BackEnd):
                 terms = [spec.terms[name] for name in spec._get_group_specific_groups()[group]]
                 self.mu += add_lkj(terms, eta)
 
-        # Add group specific terms that dont have a prior for their correlation matrix
+        # Add group specific terms that don't have a prior for their correlation matrix
         terms = [
             term
             for term in spec.group_specific_terms.values()
--- a/bambi/backends/pymc.py
+++ b/bambi/backends/pymc.py
@@ -209,7 +209,7 @@ class PyMC3BackEnd(BackEnd):
         return distribution(label, **kwargs)
 
     def build_group_specific_distribution(self, dist, label, noncentered, **kwargs):
-        """Build and return a PyMC3 Distribution."""
+        """Build and return a PyMC3 Distribution for a group specific term."""
         dist = self.get_distribution(dist)
         if "dims" in kwargs:
             group_dim = [dim for dim in kwargs["dims"] if dim.endswith("_group_expr")]
@@ -221,7 +221,6 @@ class PyMC3BackEnd(BackEnd):
             kwargs = {
                 k: self.expand_prior_args(k, v, label, noncentered) for (k, v) in kwargs.items()
             }
-
         # Non-centered parameterization for hyperpriors
         if noncentered and has_hyperprior(kwargs):
             old_sigma = kwargs["sigma"]
@@ -357,21 +356,22 @@ def add_lkj(terms, eta=1):
 
     for term in terms:
         label = term.name
-        dims = term.pymc_coords
+        dims = list(term.pymc_coords.keys())
         predictor = term.predictor.squeeze()
         delta = term.predictor.shape[1]
 
         if delta == 1:
             idx = start
         else:
-            idx = slice(start, delta)
+            idx = slice(start, start + delta)
 
         # Add prior for the parameter
         coef = pm.Deterministic(label, coefs[:, idx], dims=dims)
         coef = coef[term.group_index]
 
         # Add standard deviation of the hyperprior distribution
-        pm.Deterministic(label + "_sigma", sigma[idx])
+        group_dim = [dim for dim in dims if dim.endswith("_group_expr")]
+        pm.Deterministic(label + "_sigma", sigma[idx], dims=group_dim)
 
         # Account for the contribution of the term to the linear predictor
         if predictor.ndim > 1:
@@ -382,5 +382,5 @@ def add_lkj(terms, eta=1):
 
         start += delta
 
-    # TOD: Add correlations
+    # TO DO: Add correlations
     return mu
--- a/bambi/backends/pymc.py
+++ b/bambi/backends/pymc.py
@@ -87,13 +87,13 @@ class PyMC3BackEnd(BackEnd):
                 dist = term.prior.name
                 args = term.prior.args
                 predictor = term.predictor.squeeze()
-
                 dims = list(term.pymc_coords.keys())
+
                 coef = self.build_group_specific_distribution(
                     dist, label, noncentered, dims=dims, **args
                 )
-
                 coef = coef[term.group_index]
+
                 if predictor.ndim > 1:
                     for col in range(predictor.shape[1]):
                         self.mu += coef[:, col] * predictor[:, col]
@@ -245,7 +245,7 @@ class PyMC3BackEnd(BackEnd):
         return dist
 
     def expand_prior_args(self, key, value, label, noncentered, **kwargs):
-        # Inspect all args in case we have hyperparameters
+        # Inspect all args in case we have hyperparameters.
         # kwargs are used to pass 'dims' for group specific terms.
         if isinstance(value, Prior):
             return self.build_group_specific_distribution(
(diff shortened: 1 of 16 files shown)


tomicapretto:
I think this is ready for a review @aloctavodia, @canyon289, @twiecki. I know there's a lot going on in this PR and many things may not be that clear. Please ask as many questions as you want.

The TL;DR of this PR would be

* We can now use LKJ prior for the correlation matrix of the prior of the group-specific terms.
* We have an alternative method to compute default priors inspired on rstanarm priors. These are going to be used in the coming implementations for t family, beta family, etc.

aloctavodia (approved):
A few small comments, overall seems good. Many small details that improve readability also some changes that I need to read with more care to understand them. I will try to keep reading tomorrow.

bambi/backends/pymc.py
```diff
@@ -279,9 +311,72 @@ def _laplace(model):
...
+
+    # Construct sigma
+    # Horizontally stack the sigma values for all the hyperpriors
+    sigma = np.hstack([term.prior.args["sigma"].args["sigma"] for term in terms])
+
+    # Reconstruct the hyperprior for the standard deviations, using one variable
+    sigma = pm.HalfNormal.dist(sigma=sigma, shape=rows)
+
+    # Obtain Cholesky factor for the covariance
+    lkj_decomp, corr, sigma = pm.LKJCholeskyCov(  # pylint: disable=unused-variable
```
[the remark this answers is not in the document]
tomicapretto: I would like to use corr in the nearby future. I've been thinking we should report it even when independent priors are used. That's why it's there.

But in the meantime, I have no problem if you think the underscore is more appropriate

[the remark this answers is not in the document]
tomicapretto: And sigma... Well, they represent the same random variable in the model. The problem is the first one is a .dist, so I have to recover the one returned by lkjcholeskycov and add it to the trace.

I don't know if there are plans in pymc3 to allow a random variable in lkjcholeskycov, that would be the best solution I think

bambi/models.py
```diff
@@ -67,9 +66,16 @@ class Model:
         dictionary containing named distributions, families, and terms (see the documentation in
         ``priors.PriorFactory`` for details), or the name of a JSON file containing the same
         information.
+    automatic_priors: str
+        An optional specification to compute/scale automatic priors. ``"default"`` means to use
+        Bambi's default method. ``"rstanarm"`` means to use default priors from the R rstanarm
+        library. The latter are available in more scenarios because they don't depend on MLE.
```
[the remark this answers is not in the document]
tomicapretto: I'm going to change the defaults for the rstanarm inspired priors. This will make it simpler to implement t and beta families.

bambi/backends/pymc.py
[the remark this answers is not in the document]
tomicapretto: I think this is done in [this PR](https://github.com/pymc-devs/pymc3/pull/4784)

bambi/backends/pymc.py
```diff
@@ -279,9 +311,72 @@ def _laplace(model):
...
-
     return (
         "sigma" in kwargs
         and "observed" not in kwargs
         and isinstance(kwargs["sigma"], pm.model.TransformedRV)
     )
+
+
+def add_lkj(terms, eta=1):
+    # Parameters
```
canyon289, line 322: I would make this a full comment in numpy style, that way it shows up under `add_lkj.__doc__`

bambi/priors/scaler_mle.py
```diff
@@ -109,6 +111,13 @@ def get_slope_stats(self, exog, name, values, sigma_corr, points=4, full_model=N
         ]
         return np.array(terms).sum() ** 0.5
 
+    def scale_response(self):
+        if self.model.response.prior.auto_scale:
+            if self.model.family.name == "gaussian":
+                sigma = np.std(self.model.response.data)
+                self.model.response.prior.update(sigma=Prior("HalfStudentT", nu=4, sigma=sigma))
+            # Add cases for other families
```
canyon289, line 119: Whats this comment for?

bambi/priors/scaler_mle.py
```diff
@@ -215,23 +224,26 @@ def scale(self):
...
-                self.fit_mle()
-
             # Convert scale names to floats
             if isinstance(term.prior.scale, str):
                 term.prior.scale = self.names[term.prior.scale]
 
+            if self.mle is None:
+                self.fit_mle()
+
             # Scale it
```
canyon289, line 247: expand it to full name, makes comment easier to read in isolation

bambi/utils.py
```diff
@@ -13,46 +24,46 @@ def listify(obj):
...
+        if prior:
+            return {name: prior}
     elif isinstance(family, Family):
         # Only work if there are nuisance parameters in the family, and if any of these nuisance
         # parameters is present in 'priors' dictionary.
         nuisance_params = [k for k in family.prior.args if k not in ["observed", family.parent]]
         if set(nuisance_params).intersection(set(priors)):
-            return {k: priors[k] for k in nuisance_params if k in priors}
+            return {k: priors.pop(k) for k in nuisance_params if k in priors}
     return None
```
canyon289, line 54: What happens if none is returned to downstream call?

canyon289 (commented):
Short review. Will do more indepth review in next 24 hours

canyon289:
> I think this is ready for a review @aloctavodia, @canyon289, @twiecki. I know there's a lot going on in this PR and many things may not be that clear. Please ask as many questions as you want.
> 
> The TL;DR of this PR would be
> 
>     * We can now use LKJ prior for the correlation matrix of the prior of the group-specific terms.
> 
>     * We have an alternative method to compute default priors inspired on rstanarm priors. These are going to be used in the coming implementations for t family, beta family, etc.

Im sorry for missing this @mention. I get so many github emails things get lost. If Im not responding to a PR fast enough message me on slack, or even proactively message. That reduces chances I'll miss PRs tremendously

tomicapretto:
> > I think this is ready for a review @aloctavodia, @canyon289, @twiecki. I know there's a lot going on in this PR and many things may not be that clear. Please ask as many questions as you want.
> > The TL;DR of this PR would be
> > ```
> > * We can now use LKJ prior for the correlation matrix of the prior of the group-specific terms.
> > 
> > * We have an alternative method to compute default priors inspired on rstanarm priors. These are going to be used in the coming implementations for t family, beta family, etc.
> > ```
> 
> Im sorry for missing this @mention. I get so many github emails things get lost. If Im not responding to a PR fast enough message me on slack, or even proactively message. That reduces chances I'll miss PRs tremendously

Thanks for the feedback Ravin. I'll message on Slack next time then! :D

canyon289:
I also missed more in depth review but will get it in this week :(

tomicapretto:
> I would make this a full comment in numpy style, that way it shows up under `add_lkj.__doc__` 

Makes much sense, I'm adding it.

bambi/utils.py
tomicapretto: Nothing. When None is returned it means the user didn't pass any prior for any parameter in the response distribution and there's no need to update them.

bambi/priors/scaler_mle.py
tomicapretto: What do you mean with full name? Something like `self.fit_maximum_likeliihood_estimator`? Or is it about the comment below?

bambi/priors/scaler_mle.py
tomicapretto: For example, the Gamma family has an auxiliary parameter `alpha` whose prior is `HalfCauchy(beta=1)`. I left the comment to remember that it might be good to consider updating `beta` to another value based on the data, as with the `sigma` in the HalfStudentT prior above.

bambi/backends/pymc.py
tomicapretto: Makes sense!

tomicapretto:
I've just realized this PR will close #320 since we're changing default priors

Issue #364, opened 2021-07-05
Add `Likelihood` class


tomicapretto:
From our [Getting Started](https://bambinos.github.io/bambi/master/notebooks/getting_started.html#Families) section, this is how you create and use a custom family.

```python
from bambi import Family
import theano.tensor as tt

# Specify how the Bernoulli p parameter is distributed
prior_p = Prior('Beta', alpha=2, beta=2)

# The response variable distribution
prior = Prior('Bernoulli', p=prior_p)

# Set the link function. Alternatively, we could just set
# the link to 'logit', since it's already built into Bambi.
# Note that we could pass in our own function here; the link
# function doesn't have to be predefined.
link = tt.nnet.sigmoid

# Construct the family
new_fam = Family('bernoulli', prior=prior, link=link, parent='p')

# Now it's business as usual
model = Model("admit ~ gre + gpa + rank", data, family=new_fam)
results = model.fit()
```

Wouldn't it make more sense to have a `likelihood` argument instead of the existing `prior` one? I think we could have a `Likelihood()` class for this purpose, which would be clearer than using `Prior()` to specify the likelihood.

And actually, this part

```python
# Specify how the Bernoulli p parameter is distributed
prior_p = Prior('Beta', alpha=2, beta=2)

# The response variable distribution
prior = Prior('Bernoulli', p=prior_p)
```
is misleading because `p` does not have that prior when the model is built because it is declared as the `parent`.

aloctavodia:
I agree this is weird. But the weird thing is the need to define 

prior = Prior('Bernoulli', p=prior_p)

Instead of just doing something like.

prior = Prior('Beta', alpha=2, beta=2)
new_fam = Family('bernoulli', prior=prior, link=link, parent='p')

tomicapretto:
But this "prior" `prior = Prior('Beta', alpha=2, beta=2)` is not used, that Beta distribution goes nowhere in the model. 

This is what I mean with the `Likelihood` class.

```python
class Likelihood(name, priors=None, parent=None):
    """
    'parent' must not be in 'priors'. 'parent' is inferred from the 'name' if it is a known name

    Parameters
    ----------
    name: str
        Name of the likelihood function (PyMC3 distribution name)
    priors: dict
        Optional dictionary that contains prior distributions for auxiliary parameters in the likelihood. 
    parent: str
        Optional specification of the name of the mean parameter in the likelihood.
        This is the parameter whose transformation is modeled by the linear predictor.
    """
```

[...]

PR #366, opened 2021-07-06
Likelihood

tomicapretto:
This PR contains a lot of internal changes, but not so many changes visible to the end-user. The motivation for this PR was the inconsistency when defining custom Families as shown in #364.

### Add `Likelihood` class

There is a new `Likelihood` class that is used to represent likelihood functions within a `Family`. This `Likelihood` stores the name of the PyMC3 distribution, the priors for the auxiliary parameters of that distribution (i.e. the prior for sigma in a normal linear regression), and the name of the mean parameter as its `parent`. Custom families can now be defined as 

```python
likelihood = Likelihood("Normal", priors={"sigma": Prior("HalfNormal", sigma=10)}, parent="mu")
family = Family("gaussian", likelihood=likelihood, link="identity")
```

and 

```python
likelihood = Likelihood("Bernoulli", parent="p")
family = Family("bernoulli", likelihood=likelihood, link="probit")
```

**Question:** What about having **kwargs instead of a dictionary for priors in the `Likelihood`? For example

```python
likelihood = Likelihood("Normal",  parent="mu", sigma = Prior("HalfNormal", sigma=10))
likelihood = Likelihood("Gamma",  parent="mu", alpha = Prior("HalfCauchy", beta=1))
```

### Split classes in their own files

The file `priors.py` under the `priors` submodule contained much more things than just the `Prior` class. Now we have 

* `prior.py` with the `Prior` class.
* `family.py`, with the `Family` class.
* `likelihood.py` with the `Likelihood` class.

This does not have any direct impact on the user experience, but I think it is better to have a more modular codebase.

### Removed `PriorFactory` class

The `PriorFactory` class was located in `priors/priors.py`. This class wouldn't generate only priors. It would generate priors and families. It would also behave differently depending on whether you passed a term type, a family name, or something that in the config.json was known as a "dist".  I think its logic was quite confusing. See [here](https://github.com/bambinos/bambi/blob/c0ca1072cc2809ba917ff0b30ce0b57bd645b98f/bambi/priors/priors.py#L206) and [here](https://github.com/bambinos/bambi/blob/c0ca1072cc2809ba917ff0b30ce0b57bd645b98f/bambi/priors/priors.py#L224)

All of these defaults/builtins functionality was replaced with the `defaults` submodule. It exports two functions, `get_default_prior()` and `get_builtin_family()`. The first takes a term type and returns the default prior for that type, and the second takes a family name and returns a `Family` instance. 

### Removed `config.json` for priors

This JSON file was replaced with Python dictionaries within `defaults/defaults.py`. All the logic for the default priors and the built-in families is self-contained in there. On the downside, users can't pass their default prior configuration as a JSON file... but honestly, I think nobody was using this.


canyon289:
`Question:` What about having **kwargs instead of a dictionary for priors in the Likelihood? 

I think kwargs might be nicer so a dictionary doesnt need to be constructed but I havent thought deeply about the implications.

I'm glad you removed PriorFactory as well, it was confusing

PR #367, opened 2021-07-12
Add t family

tomicapretto:


tomicapretto:
Based on Robust regression PyMC3 example

```python
import numpy as np
import pandas as pd

from bambi import Model

size = 100
true_intercept = 1
true_slope = 2

x = np.linspace(0, 1, size)
true_regression_line = true_intercept + true_slope * x
y = true_regression_line + np.random.normal(scale=0.5, size=size)

x_out = np.append(x, [0.1, 0.15, 0.2])
y_out = np.append(y, [8, 6, 9])

data = pd.DataFrame(dict(x = x_out, y = y_out))


# Fit model, now with t family :D
model = Model("y ~ x", data, family="t")
idata = model.fit()

# See results
a = idata.posterior["Intercept"].values.flatten()
b = idata.posterior["x"].values.flatten()

x_seq = np.linspace(0, 1, num=100)
mu_pred = np.zeros((len(x_seq), len(a)))

for i, x in enumerate(x_seq):
    mu_pred[i] = a + b * x

fig, ax = plt.subplots(figsize=(14, 10))
ax.plot(x_seq, mu_pred, color="#4d4d4d", alpha=0.1)
ax.plot(x_seq, 1 + 2 * x_seq, color="black")
ax.scatter(x_out, y_out, s=80, color="#7a5195", edgecolors="black", zorder=10)
plt.savefig("t.png", facecolor="white")
```

t

where black line is the true regression line

PR #372, opened 2021-07-13
Implement predictions

tomicapretto:
Add a `.predict()` method to the `Model` class. This enables both in-sample and out-sample prediction. The `.predict()` method has the following arguments

* `idata`: `InfereceData` with samples from the posterior distribution. This is mandatory.
* `kind`: `str` either `"mean"` or ~`"distribution"`~ `"pps"`. Defaults to `"mean"`.
* ~`scale`: `str` either `"response"` or `"link"`. Defaults to `"response"`.~
* `data`: `pd.DataFrame` or `None`. If `pd.DataFrame`, it is where we look for variables with which to predict (out of sample predictions). If `None`, it uses the original data and thus obtains in-sample predictions.

## How it works

Let's say we have a logistic regression with mixed effects

```python
import bambi as bmb
import numpy as np
import pandas as pd

f_delta = pd.DataFrame({
    "A": np.random.choice([0, 1], size=100),
    "B": np.random.normal(size=100),
    "C": np.random.normal(size=100),
    "D": np.random.normal(size=100),
    "E": ["Group 1"] * 25 + ["Group 2"] * 25 + ["Group 3"] * 25 + ["Group 4"] * 25
})

# Fit the model
model = bmb.Model('A ~ 0 + B + C + D + (1|E)', f_delta, family = 'bernoulli')
idata = model.fit()

# Select some random rows from the original data
# In practice this would be another data set.
rows = np.random.choice(np.arange(f_delta.shape[0]), size = 20)
f_delta_2 = f_delta.iloc[rows, :]
```
### Mean prediction in the response scale

This is the default behavior

```python
model.predict(idata, data=f_delta_2)
array([0.63472875, 0.57823079, 0.45717866, 0.494302  , 0.61275881,
       0.52552663, 0.51325799, 0.35138795, 0.53079801, 0.51123194,
       0.62658852, 0.51519193, 0.51138368, 0.56288993, 0.64367165,
       0.53358324, 0.42119659, 0.60092995, 0.44419184, 0.49119846])
```

### Predictive distribution in the response scale

```python
predictions = model.predict(idata, kind="distribution", data=f_delta_2)
predictions.shape # (20, 2000)
```

```python
predictions
array([[0.57790498, 0.63729905, 0.57557616, ..., 0.5984925 , 0.77373198,
        0.55506065],
       [0.5476052 , 0.49407264, 0.44980499, ..., 0.54965804, 0.76856762,
        0.6191289 ],
       [0.43681045, 0.49063546, 0.60930524, ..., 0.43931939, 0.38990382,
        0.46463923],
       ...,
       [0.49536406, 0.60846616, 0.61157835, ..., 0.53008153, 0.69289172,
        0.52488174],
       [0.4375965 , 0.38921687, 0.39291667, ..., 0.41222194, 0.33595178,
        0.52009173],
       [0.52432786, 0.48379762, 0.46095579, ..., 0.50706812, 0.53381684,
        0.40977213]])
```

### Other usages

* With `scale="link"` the predictions are obtained in the scale of the linear predictor (e.g. log-odds for the model above)
* If `data=None`, the predictions are obtained for the sample used to fit the model.

### Notes

* What about having an option that returns both the mean and quantiles and/or the whole distribution? This could be useful when you want to plot credibility bands for the mean as well as the mean value.
* This implementation uses the trace obtained when sampling from the posterior. As far as I understand, `pm.sample_posterior_predictive()` draws samples of parameters from that trace and, for each sample, it draws N random values from the response distribution with its parameters based on the given draw from the trace.

### DIFF
--- a/bambi/defaults/pps.py
+++ b/bambi/defaults/pps.py
@@ -1,0 +1,68 @@
+"""Utility functions to sample from the posterior predictive distributions"""
+
+import numpy as np
+
+from scipy import stats
+
+
+def _get_mu_and_idxs(mu, draws): #pylint: disable=unused-argument
+    # mu has shape (chain, draw, obs)
+    idxs = np.random.randint(low=0, high=draws, size=draws)
+    mu = mu[:, idxs, :]
+    return mu, idxs
+
+
+def pps_bernoulli(model, posterior, mu, draws): #pylint: disable=unused-argument
+    mu, _ = _get_mu_and_idxs(mu, draws)
+    return np.random.binomial(1, mu)
+
+
+def pps_beta(model, posterior, mu, draws):
+    mu, idxs = _get_mu_and_idxs(mu, draws)
+    kappa = posterior[model.response.name + "_kappa"].values[:, idxs, np.newaxis]
+    alpha = mu * kappa
+    beta = (1 - mu) * kappa
+    return np.random.beta(alpha, beta)
+
+
+def pps_gamma(model, posterior, mu, draws):
+    mu, idxs = _get_mu_and_idxs(mu, draws)
+    alpha = posterior[model.response.name + "_alpha"].values[:, idxs, np.newaxis]
+    beta = alpha / mu
+    return np.random.gamma(alpha, 1 / beta)
+
+
+def pps_gaussian(model, posterior, mu, draws):
+    mu, idxs = _get_mu_and_idxs(mu, draws)
+    sigma = posterior[model.response.name + "_sigma"].values[:, idxs, np.newaxis]
+    return np.random.normal(mu, sigma)
+
+
+def pps_negativebinomial(model, posterior, mu, draws):
+    mu, idxs = _get_mu_and_idxs(mu, draws)
+    n = posterior[model.response.name + "_alpha"].values[:, idxs, np.newaxis]
+    p = n / (mu + n)
+    return np.random.negative_binomial(n, p)
+
+
+def pps_poisson(model, posterior, mu, draws): #pylint: disable=unused-argument
+    mu, _ = _get_mu_and_idxs(mu, draws)
+    return np.random.poisson(mu)
+
+
+def pps_t(model, posterior, mu, draws):
+    mu, idxs = _get_mu_and_idxs(mu, draws)
+    if isinstance(model.family.likelihood.priors["nu"], (int, float)):
+        nu = model.family.likelihood.priors["nu"]
+    else:
+        nu = posterior[model.response.name + "_nu"].values[:, idxs, np.newaxis]
+
+    lam = posterior[model.response.name + "_lam"].values[:, idxs, np.newaxis]
+    sigma = lam ** -0.5
+    return stats.t.rvs(nu, mu, sigma)
+
+
+def pps_wald(model, posterior, mu, draws):
+    mu, idxs = _get_mu_and_idxs(mu, draws)
+    lam = posterior[model.response.name + "_lam"].values[:, idxs, np.newaxis]
+    return np.random.wald(mu, lam)
--- a/bambi/defaults/pps.py
+++ b/bambi/defaults/pps.py
@@ -6,6 +6,28 @@ from scipy import stats
 
 
 def _get_mu_and_idxs(mu, draws):  # pylint: disable=unused-argument
+    """Sample values from the posterior of the mean, return sampled values and their indexes
+
+    This function is used within ``pps_*`` auxiliary functions. Its goal is to simplify how we
+    sample from the posterior trace and the other parameters that may define the conditional
+    distribution of the response.
+
+    Parameters
+    ----------
+    mu: np.array
+        A 3-dimensional numpy array with samples from the posterior distribution of the mean. The
+        first dimension represents the chain, the second represents the draw, and the third the
+        individual.
+    draws: int
+        Number of samples to take.
+
+    Returns
+    -------
+    mu: np.array
+        A 3-dimensional numpy array with the new samples.
+    idxs: np.array
+        The indexes of the samples in the original sample of the posterior.
+    """
     # mu has shape (chain, draw, obs)
     idxs = np.random.randint(low=0, high=draws, size=draws)
     mu = mu[:, idxs, :]
--- a/bambi/defaults/pps.py
+++ b/bambi/defaults/pps.py
@@ -5,14 +5,14 @@ import numpy as np
 from scipy import stats
 
 
-def _get_mu_and_idxs(mu, draws): #pylint: disable=unused-argument
+def _get_mu_and_idxs(mu, draws):  # pylint: disable=unused-argument
     # mu has shape (chain, draw, obs)
     idxs = np.random.randint(low=0, high=draws, size=draws)
     mu = mu[:, idxs, :]
     return mu, idxs
 
 
-def pps_bernoulli(model, posterior, mu, draws): #pylint: disable=unused-argument
+def pps_bernoulli(model, posterior, mu, draws):  # pylint: disable=unused-argument
     mu, _ = _get_mu_and_idxs(mu, draws)
     return np.random.binomial(1, mu)
 
@@ -45,7 +45,7 @@ def pps_negativebinomial(model, posterior, mu, draws):
     return np.random.negative_binomial(n, p)
 
 
-def pps_poisson(model, posterior, mu, draws): #pylint: disable=unused-argument
+def pps_poisson(model, posterior, mu, draws):  # pylint: disable=unused-argument
     mu, _ = _get_mu_and_idxs(mu, draws)
     return np.random.poisson(mu)
 
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -641,10 +641,10 @@ class Model:
         idata : InfereceData
             ``InfereceData`` with samples from the posterior distribution.
         draws : int
-            Number of draws to sample from the prior predictive distribution. Defaults to 500.
+            Number of draws to sample from the posterior predictive distribution. Defaults to 500.
         var_names : str or list
             A list of names of variables for which to compute the posterior predictive
-            distribution. Defaults to both observed and unobserved RVs.
+            distribution. Defaults to observed RVs.
         inplace : bool
             If ``True`` it will add a ``posterior_predictive`` group to idata, otherwise it will
             return a copy of idata with the added group. If ``True`` and idata already have a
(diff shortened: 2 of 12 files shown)


tomicapretto:
Some clarifications:

This does not make predictions for punctual observations **yet**. The distribution obtained when you do `kind="distribution"` is the posterior distribution for the mean. If you take those samples and compute a credibility band, it is a credibility interval for the mean and not for punctual predictions. This functionality will be added later (and we'll discuss whether we deprecate `Model.posterior_predictive()`)

gregorystrubel:
I am looking for to test it.

I wonder if for the kind="distribution" you don't want to return a dict {'y':your array} (the same format that pymc return for samples, if I am correct).

It will permit to use all the ArviZ functions.

gregorystrubel:
> Some clarifications:
> 
> This does not make predictions for punctual observations **yet**. The distribution obtained when you do `kind="distribution"` is the posterior distribution for the mean. If you take those samples and compute a credibility band, it is a credibility interval for the mean and not for punctual predictions. This functionality will be added later (and we'll discuss whether we deprecate `Model.posterior_predictive()`)

Thanks for the precision. Naming the variable is maybe more difficult than just 'y' in that case (we might expect the influence of the noise).

gregorystrubel:
Does R have two functions to discriminate the two kind of predictions ?:
 - One that give you the distribution and the mean of the linear response (excluding noise influence) (that give you results near traditional glm)
 - One give you distribution and the mean including the noise influence.
 
 (I forget to thank you for giving us this tool)

tomicapretto:
> I am looking for to test it.
> 
> I wonder if for the kind="distribution" you don't want to return a dict {'y':your array} (the same format that pymc return for samples, if I am correct).
> 
> It will permit to use all the ArviZ functions.
> 
> EDIT: naming the variable is maybe more difficult than I think (see my next comment). This comment may be useless.

I think it's good to take into account ArviZ when thinking about the return format. An InferenceData object is also a good candidate.

On the other hand, I think functions like `az.plot_hdi()` will be the most used along with this prediction method, and that one requires numpy arrays. 

It would be good to think about the pros and cons of the different approaches for the different use cases.

tomicapretto:
> Does R have two functions to discriminate the two kind of predictions ?:
> 
> * One that give you the distribution and the mean of the linear response (excluding noise influence) (that give you results near traditional glm)
> * One give you distribution and the mean including the noise influence.
> 
> (I forget to thank you for giving us this tool)


Do you refer to [this]? So yes, we should have both. The question is whether that should be the same method or not. What we currently have in this PR is the posterior distribution for the mean. We still need the posterior distribution for a new observation (posterior predictive distribution)(*)

And thank you very much for taking the time to leave your suggestions here :smile: 

(*) This is what `Model.posterior_predictive()` does, but only for in-sample data. The goal is to extend it to out-of-sample data.

bambi/models.py
```diff
@@ -688,6 +688,74 @@ def posterior_predictive(
...
+    # pylint: disable=protected-access
+    def predict(self, idata, kind="mean", scale="response", data=None):
+        """Predict method for Bambi models
+
+        Obtains in-sample and out-sample predictions from a fitted Bambi model.
+
+
+        Parameters
+        ----------
+        idata : InfereceData
```
canyon289, line 700:
```suggestion
        idata : InferenceData
```

bambi/models.py
```diff
@@ -688,6 +688,74 @@ def posterior_predictive(
...
+            ``InfereceData`` with samples from the posterior distribution.
```
canyon289, line 701:
```suggestion
            ``InferenceData`` with samples from the posterior distribution.
```

bambi/models.py
```diff
@@ -688,6 +688,74 @@ def posterior_predictive(
...
+        data: pd.DataFrame or None
+            An optional data frame in which to look for variables with which to predict.
+            If omitted, the fitted linear predictors are used.
+
+        Returns
+        -------
+        np.ndarray
+            A NumPy array with predictions. The length of the first dimension is given by the number
+            of observations in the dataset used for the prediction. A second dimension is present
+            when ``kind=="distribution"`` and its lenght is equal to the number of draws from the
```
canyon289, line 720:
```suggestion
            when ``kind=="distribution"`` and its length is equal to the number of draws from the
```

bambi/models.py
```diff
@@ -688,6 +688,74 @@ def posterior_predictive(
...
+
+        Obtains in-sample and out-sample predictions from a fitted Bambi model.
+
+
+        Parameters
+        ----------
+        idata : InfereceData
+            ``InfereceData`` with samples from the posterior distribution.
+        kind: str
+            Indicates the type of prediction required. Can be ``"mean"`` or ``"distribution"``. The
```
aloctavodia, line 703: We should return samples and let the users compute the mean/median etc if then want. 

What about reserving the name "mean" for the posterior mean function and "pps" for the posterior predictive distribution? (once we deprecate `model.posterior_predictive()`)

Also by default we could have an `inplace` argument to control if we add the results to the passed idata or return a new one. If `kind=mean`, the results will be added to a new variable in the posterior group, we can call it "y_hat" by default or something similar (and/or let the user pass a name). And if `kind=pps` then the samples will be added to the `posterior_predictive` group for in-sample predictions and the `prediction` group for out-of-sample predictions.

aloctavodia:
I think InferenceData is the way to go. We should not worry about directly working with `az.plot_hdi()` but instead with high level functions see arviz-devs/arviz#1727

bambi/models.py
tomicapretto: My comments:

* I think `"y_hat"` is more associated with the `"pps"` concept, because the response variable is what is being predicted.
* `kind="mean"` and `kind="pps"` are good choices. Let's discard `kind="distribution"`.
* Maybe `scale="linpred"` is clearer than `scale="link"`? Do you think there's any drawback with having another scale than the response? I'm thinking that `kind="pps"` only makes sense with`scale="response"`. An error could be raised when `kind="pps"` and `scale`  is not `"response"`, but I would like to know if you think this is confusing.
* If `kind="mean"` I agree we can add a new variable to the posterior.  If `scale="response"`, this can have the name of the response + "_mean", e.g. the response is `"vote"`, then the new variable is `"vote_mean"`. If `scale="linpred"` the name could be `"vote_linpred"`.

Finally, I'm good with returning an InferenceData object. But one of the things I like most about Bambi is that it is simple for newcomers. So I think we should have an example or section called **Recipes** or **How to** where we add several examples showing how to obtain predictions and create a variety of plots with them.

aloctavodia: Not strong opinion about `"y_hat"`. I agree is ambiguous I have seen used it for the mean of Y and for the pps of Y. 

I feel more inclined to remove the argument `scale` (but not opposed if you want to leave it). I guess most user will use the default and for the few user that may want  to do a transformation of the data they could do it manually, without too much trouble. We may want to expose the link functions to the user to make they live easier...

I agree with the rest, specially about more examples about how to operate with InferenceData. ArviZ has a Working with InferenceData section in the docs https://arviz-devs.github.io/arviz/getting_started/WorkingWithInferenceData.html (is still incomplete). We should both try to provide examples and report problems back to ArviZ to improve the usability of InferenceData.

tomicapretto: Good, I agree. I will do the following

* Remove `scale`.
* `Family` already has `link` (a string) and `_link` (a function) that converts the linear predictor to the mean. I'm going to modify it a little so we have `link_str` the name of the link function, `link` the link function itself, and `invlink` the inverse of the link function that goes from the mean to the linear predictor. Then we can have one or two examples where we show how to call `model.family.invlink()` to go from the mean prediction to the link scale. Sounds good?

aloctavodia: agreed

tomicapretto: I ended up creating a `Link` class. This has a `.name`, `.link`, `.linkinv` and `.linkinv_backend`. Built-in links are obtained just with a name. Also, I also clarify the distinction between `link` and `linkinv` so it is consistent with `g` and `g^-1` in GLMs

tomicapretto: This does not imply any change for the user, unless they want to create a custom ``Family`` with custom ``Link``, which would be very infrequent I think.

tomicapretto:
Now it's possible to use `Model.predict()` with `kind="mean"` and `kind="pps"`, and the results are added to an InferenceData object. 

There are things to be improved yet. 

* How the predictions enter into the InferenceData object. I'm losing the draws dimension and I'm creating a new group when `kind="mean"`, when it should be added to `posterior`. The thing is that it will add a new dimension (the number of observations).
* The posterior predictive samples are kind of hardcoded into a function. I think there must be a better approach.

tomicapretto:
Everything seems to be working now. This will be done after adding some tests.

bambi/models.py
```diff
@@ -802,3 +906,31 @@ def common_terms(self):
     def group_specific_terms(self):
         """Return dict of all and only group specific effects in model."""
         return {k: v for (k, v) in self.terms.items() if v.group_specific}
+
+
+def posterior_predictive(model, posterior, mu, draws):
```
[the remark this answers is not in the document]
tomicapretto: To make it work with out-of-sample data without having to use shared variables via `pm.Data`. Or do you think it should be better to use shared variables?

bambi/models.py
```diff
@@ -688,6 +689,109 @@ def posterior_predictive(
...
+        mu = self.family.link.linkinv(linear_predictor).T
+
+        # Predictions for the mean
+        if kind == "mean":
+            name = self.response.name + "_mean"
+            if "predictions" in idata:
+                del idata.predictions
+
+            # Array of shape (chain, draws, obs), and chain = 1.
+            idata.add_groups({"predictions": {name: mu[np.newaxis]}})
```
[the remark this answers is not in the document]
tomicapretto: It's in the posterior now. It was easier to test if it worked with this new group, but it's not there anymore.

bambi/models.py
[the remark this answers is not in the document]
aloctavodia: Ohh right! This is fine, maybe for pymc3 v4 we could have a cleaner implementation, but in the meantime this is good.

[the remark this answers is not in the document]
tomicapretto: Great! I agree. Let's leave this as it is, and then once PyMC3 v4 is out we can revisit it.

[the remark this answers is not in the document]
aloctavodia: BTW, we should add a deprecation warning to the `posterior_predictive` method in favor of `predict`

tomicapretto:
I think this is done.

aloctavodia (approved):

bambi/defaults/pps.py
```diff
@@ -0,0 +1,68 @@
+"""Utility functions to sample from the posterior predictive distributions"""
+
+import numpy as np
+
+from scipy import stats
+
+
+def _get_mu_and_idxs(mu, draws):  # pylint: disable=unused-argument
```
twiecki, line 8: no doc strings?

zhanwenchen:
Fyi, this is what I had to do to actually get `predict` to work. Apparent it does NOT return an ndarray as advertised on the docs.

```python
from patsy import build_design_matrices

(new_x,) = build_design_matrices([X.design_info], {'sex': 0, 'bmi':28, 'age': 48})
# new_ypred_gaussian = model_gaussian.predict(new_x)

new_x_df = pd.DataFrame(new_x, columns=new_x.design_info.column_names)
new_ypred_gaussian_2 = model_bayes.predict(model_fitted, data=new_x_df, kind="mean", inplace=False)

print(new_ypred_gaussian_2.posterior.to_dataframe()['scl'].mean()) # 240.0359379078641
```

tomicapretto:
> Fyi, this is what I had to do to actually get `predict` to work. Apparent it does NOT return an ndarray as advertised on the docs.
> 
> ```python
> from patsy import build_design_matrices
> 
> (new_x,) = build_design_matrices([X.design_info], {'sex': 0, 'bmi':28, 'age': 48})
> # new_ypred_gaussian = model_gaussian.predict(new_x)
> 
> new_x_df = pd.DataFrame(new_x, columns=new_x.design_info.column_names)
> new_ypred_gaussian_2 = model_bayes.predict(model_fitted, data=new_x_df, kind="mean", inplace=False)
> 
> print(new_ypred_gaussian_2.posterior.to_dataframe()['scl'].mean()) # 240.0359379078641
> ```

I'm checking our API reference and I don't see where it says it returns an ndarray https://bambinos.github.io/bambi/main/api_reference.html. Could you point me to the wrong documentation please?

image

PR #375, opened 2021-07-16
Load data

### LINKED ISSUE #373: Add function to retrieve datasets

aloctavodia:
This PR closely follows the ArviZ implementation of `load_arviz_data`. but with some differences. closes #373

It provides two new functions `load_data()` and `clear_home_cache`, the first one load a csv file from figshare or from a home folder if already downloaded and returns a DataFrame. If called without arguments it list all available datasets.  The second one deleted the home cache folder. Currently only one dataset is available, but ideally we should add all the ones used in the examples/documentation.

### DIFF
--- a/bambi/__init__.py
+++ b/bambi/__init__.py
@@ -1,5 +1,6 @@
 import logging
 
+from .data import clear_data_home, load_data
 from .models import Model
 from .priors import Prior, Family
 from .backends import PyMC3BackEnd
--- a/bambi/data/__init__.py
+++ b/bambi/data/__init__.py
@@ -1,0 +1,4 @@
+"""Code for loading datasets."""
+from .datasets import clear_data_home, load_data
+
+__all__ = ["clear_data_home", "load_data"]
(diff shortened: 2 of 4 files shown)


tomicapretto:
It looks like a robust implementation. What I had in mind is just a function that wraps some `pd.read_csv()` and reads the csv files that are already included in Bambi, which would work without internet for example.  What are the advantages of this approach over that?

aloctavodia:
With this implementation we will be able to distribute the base code separated from the datasets. This is also useful in case we want to have an example with a big dataset. And in the future it can be extended for example to retrieve fitted models.

tomicapretto:
> With this implementation we will be able to distribute the base code separated from the datasets. This is also useful in case we want to have an example with a big dataset. And in the future it can be extended for example to retrieve fitted models.

Sounds good!! What about the host? Can I also upload datasets in there? If that's possible, I could help.

aloctavodia:
Totally, you just need an account in https://figshare.com/

PR #377, opened 2021-07-17
Add some datasets and update notebooks

tomicapretto:
* Add `adults`, `ESCS`, `ANES` and `carclaims` datasets.
* Remove `.csv` from the name of the data set. I think the extension is not meaningful.


aloctavodia (approved):
I was unsure about using or not the extension of the data. The reason to keep it, was to easily identify that the data is actual data and not a model, assuming in the future we will also may want to distribute models.

tomicapretto:
> I was unsure about using or not the extension of the data. The reason to keep it, was to easily identify that the data is actual data and not a model, assuming in the future we will also may want to distribute models. 

That make sense. I think being able to load models is a nice feature to have. But maybe that can be `load_model()`? Then we would have load_data and load_model which are clear about what they do.

Issue #381, opened 2021-07-23
Steps to troubleshoot if `fit` fails


CamDavidsonPilon:
👋 I'm really enjoying the ability to perform Bayesian mixed-effect models with _bambi_, but I had a frustrating first experience. I'll detail my thought process as an end-user.

I followed the examples in the docs, but when I went to perform a mixed-effects model with my data (csv attached below), I hit the following bug: 

```python
model = Model("od ~ temp + (1|source) + 0", df)
results = model.fit(draws=2000, chains=2)
```

```
/Users/camerondavidson-pilon/venvs/data/lib/python3.9/site-packages/pymc3/step_methods/hmc/quadpotential.py:224: RuntimeWarning: divide by zero encountered in true_divide
  np.divide(1, self._stds, out=self._inv_stds)
/Users/camerondavidson-pilon/venvs/data/lib/python3.9/site-packages/pymc3/step_methods/hmc/quadpotential.py:203: RuntimeWarning: invalid value encountered in multiply
  return np.multiply(self._var, x, out=out)
---------------------------------------------------------------------------
RemoteTraceback                           Traceback (most recent call last)
RemoteTraceback:
"""
Traceback (most recent call last):
  File "/Users/camerondavidson-pilon/venvs/data/lib/python3.9/site-packages/pymc3/parallel_sampling.py", line 137, in run
    self._start_loop()
  File "/Users/camerondavidson-pilon/venvs/data/lib/python3.9/site-packages/pymc3/parallel_sampling.py", line 191, in _start_loop
    point, stats = self._compute_point()
  File "/Users/camerondavidson-pilon/venvs/data/lib/python3.9/site-packages/pymc3/parallel_sampling.py", line 216, in _compute_point
    point, stats = self._step_method.step(self._point)
  File "/Users/camerondavidson-pilon/venvs/data/lib/python3.9/site-packages/pymc3/step_methods/arraystep.py", line 276, in step
    apoint, stats = self.astep(array)
  File "/Users/camerondavidson-pilon/venvs/data/lib/python3.9/site-packages/pymc3/step_methods/hmc/base_hmc.py", line 147, in astep
    self.potential.raise_ok(self._logp_dlogp_func._ordering.vmap)
  File "/Users/camerondavidson-pilon/venvs/data/lib/python3.9/site-packages/pymc3/step_methods/hmc/quadpotential.py", line 272, in raise_ok
    raise ValueError("\n".join(errmsg))
[...]

PR #383, opened 2021-07-26
switch initialization method if sampling fails

aloctavodia:
As mentioned in #381 sampling for some models can fail with the default initialization. The best solution will be to work on a more robust initialization method upstream in the meantime this PR offers to automatically switch the initialization method from "jitter-adapt_diag" to "adapt_diag" when sampling fails.


tomicapretto:
Is it possible to test this fallback method? What if we use the data in #381 or another synthetic data?

tomicapretto (approved):

PR #385, opened 2021-07-26
Improve priors and internal model specification

tomicapretto:
### Priors

The standard deviation of the prior for the intercept is now based on the older approach (as in Jake's paper).
The new default method based on rstanarm priors uses a fixed value for the standard deviation of the prior for the intercept, which resulted in very narrow priors in several scenarios.

### Center predictors

Bambi now centers the predictors when there is an intercept in the model, as brms does. This results in i**mproved sampling efficiency**. For example

image

versus the one in the docs 

image

I'm also running the notebooks again so you can see how it affects sampling efficiency.

## Potential problems

* The PyMC3 model now has both `"Intercept" `and `"Intercept__"` variables. I couldn't figure out how to remove variables from a PyMC3 model.


aloctavodia:
Centering makes total sense! About avoiding having two variables for the Intercept, why don't we use the shifted Intercept internally, and return samples (in the InferenceData object) in the original scale? For PyMC3 V4 it should be easier to delete variables from models in a consistent way. 

I will try to do a more detailed review tomorrow.

tomicapretto:
> Centering makes total sense! About avoiding having two variables for the Intercept, why don't we use the shifted Intercept internally, and return samples (in the InferenceData object) in the original scale? For PyMC3 V4 it should be easier to delete variables from models in a consistent way.
> 
> I will try to do a more detailed review tomorrow.

It makes much sense. I actually tried, but I don't remember why I gave up on that. I'll try again!

PR #386, opened 2021-07-30
Add Binomial family

tomicapretto:
This PR adds the Binomial family. This should be used with the new `prop()` function in formulae. For example 

```python
Model("prop(y, n) ~ x + z", data, family="binomial")
```

## Extended example

This problem shows how to work with this notation, as well as a problem I'm encountering with ArviZ (cc @aloctavodia).
This example uses the same data as [this example](https://bambinos.github.io/bambi/master/notebooks/alternative_links_binary.html) in the docs.

```python
import arviz as az
import bambi as bmb
import numpy as np
import pandas as pd

data = pd.DataFrame({
    "x": np.array([1.6907, 1.7242, 1.7552, 1.7842, 1.8113, 1.8369, 1.8610, 1.8839]),
    "n": np.array([59, 60, 62, 56, 63, 59, 62, 60]),
    "y": np.array([6, 13, 18, 28, 52, 53, 61, 60])
})

# Define model using binomial family.
model = bmb.Model("prop(y, n) ~ x", data, family="binomial")

# Fit model
idata = model.fit()

# Compute in-sample predictions
model.predict(idata) # predict the mean
model.predict(idata, kind="pps") # posterior predictive 
```

Now I plot the predictions

```python
# Posterior predictive
az.plot_hdi(data["x"], idata.posterior_predictive["prop(y, n)"].values )

# Posterior for  the mean "p", multiplied by "n" to put everything in the same scale
az.plot_hdi(data["x"], idata.posterior["prop(y, n)_mean"].values * data["n"].values[np.newaxis,  np.newaxis, :])

# Posterior for the mean "p", multiplied by "n" to put everything in the same scale
plt.plot(data["x"], idata.posterior["prop(y, n)_mean"].values.reshape((2000, 8)).mean(0) * data["n"])

# Values observed
plt.plot(data["x"], data["y"], "o", color="black");
```

And this is what I obtain, showing the consistency between the data and the fit.

image

On the other hand, I try to use `az.plot_ppc()` to see the posterior predictive distribution. See the result.

```python
az.plot_ppc(idata);
```

image

The posterior predictive mean makes sense, but the distribution for the observed data does not. 
I tried using the `data_pairs` without much success. Do you know what could be the problem here? I want to make sure that Bambi does not set up things in such a way that it breaks ArviZ.

**Edit**

The bounds for the posterior predictive are weird. `n` is at most 63, but values go up to ~ 90


tomicapretto:
With the exception of the weird behavior in `az.plot_ppc()`, I think this is ready.

aloctavodia:
The weird behaviour in `az.plot_ppc()` is most likely a bug in ArviZ I will check if that's the case and if so then solve it.

bambi/priors/link.py
```diff
@@ -90,6 +94,34 @@ def logit(mu):
...
+    ----------
+    name: str
+        The name of the link function. If it is a known name, it's not necessary to pass any
+        other arguments because functions are already defined internally. If not known, all of
+        ``link``, ``linkinv`` and ``linkinv_backend`` must be specified.
+    link: function
+        A function that maps the response to the linear predictor. Known as 'g()' in GLM jargon.
+        Does not need to be specified when ``name`` is a known name.
+    linkinv: function
+        A function that maps the linear predictor to the response. Knonwn as 'g()^-1' in GLM jargon.
```
canyon289, line 117:
```suggestion
        A function that maps the linear predictor to the response. Known as 'g()^-1' in GLM jargon.
```

bambi/priors/link.py
```diff
@@ -90,6 +94,34 @@ def logit(mu):
...
+        Does not need to be specified when ``name`` is a known name.
+    linkinv_backend: function
+        Same than ``linkinv`` but must be something that works with PyMC3 backend (i.e. it must
+        work with Theano tensors). Does not need to be specified when ``name`` is a known name.
```
canyon289, line 121:
```suggestion
        work with Aesara tensors). Does not need to be specified when ``name`` is a known name.
```

Probably better to use the new library name

bambi/terms.py
```diff
@@ -17,14 +17,28 @@ def __init__(self, term, family):
         self.name = term.name
         self.data = term.design_vector
         self.categorical = term.type == "categoric"
-        self.success_event = term.refclass if term.refclass is not None else 1
+        self.success_event = term.success if term.success is not None else 1
```
canyon289, line 20: Way more readable. thanks

bambi/priors/link.py
tomicapretto: You're right!

PR #390, opened 2021-08-08
some new docs on the necessary step before calling .graph()

CamDavidsonPilon:
👋 The error message in this function wasn't very clear, so I hope to improve it along with the doc string.

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -877,7 +877,9 @@ class Model:
 
         """
         if self.backend is None:
-            raise ValueError("The model is empty. Are you forgetting to first call .build() or .fit() on the Bambi model?")
+            raise ValueError("The model is empty. "
+                "Are you forgetting to first call .build() or .fit() on the Bambi model?"
+            )
 
         graphviz = pm.model_to_graphviz(model=self.backend.model, formatting=formatting)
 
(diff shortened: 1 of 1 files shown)


bambi/models.py
```diff
@@ -864,9 +864,16 @@ def graph(self, formatting="plain", name=None, figsize=None, dpi=300, fmt="png")
...
+
+        Example
+        --------
+        > model = Model('y ~ x + (1|z)')
+        > model.build()
+        > model.graph()
+
         """
         if self.backend is None:
-            raise ValueError("The model is empty, please define a Bambi model")
```
CamDavidsonPilon, line 869: When I encountered this error, I thought it was strange because I was literally calling `model.graph()` - how could my model be not defined? I hope the new error message suggests the solution.

aloctavodia:
Hi, thanks for your contribution. Notice that you need to either build or fit the model, so 

```python
model = Model('y ~ x + (1|z)')
model.fit()
model.graph()
```

Will also work.

aloctavodia (approved):

bambi/models.py
tomicapretto: Yes, you're completely right. Thanks for providing a better message!

PR #391, opened 2021-08-09
Update documentation for new release

tomicapretto:


bambi/priors/scaler_mle.py
```diff
@@ -218,7 +218,6 @@ def scale(self):
 
         if self.has_intercept:
             intercept = [self.model.intercept_term]
```
aloctavodia, line 220: good catch :-)

PR #399, opened 2021-08-21
Move families to their own module, improve many docstrings

tomicapretto:
Fixes and improvements to our docs. I was working on the paper and figured out that some docstrings weren't right or accurate. 

Some things I would like to agree with you before merging cc @aloctavodia @canyon289 

* I'm double-quotes, like `"string"`, in all the docstrings because that is what Black, our formatter, does. I'm not opposed to using single quotes, but we should be consistent. Any strong opinion in favor or against any of these alternatives?
* I'm referencing classes as `bambi.families.Likelihood` instead of just `Likelihood`. I've seen it in another repo and I think it is a good practice because it reduces ambiguity. But I would like to know if this is too much redundancy.
* I'm completely in favor of promoting Aesara. However, we're not using Aesara/Theano names consistently. We are using Aesara in the paper and in some parts of our docs, but Bambi actually uses Theano (e.g. see the link inv functions in the PyMC3 backend). I have two ideas:
  + Use the name Theano everywhere, but be very explicit we're going to use Aesara when PyMC3 v4 is ready.
  + Try to use Aesara as it is? I'm not following its development, so I'm not sure if it supports the things we currently use from Theano. If that's the case, I would drop Theano and use Aesara from now.

Finally, I think we should launch a new patch release that includes all these last-minute changes before we finish reviewing the paper. At the moment, the great majority of the paper is consistent with Bambi 0.6.0, but docstrings are not. I would like to be 100% consistent and reproducible so we can share with reviewers and users the code and the version of Bambi to reproduce everything.

### DIFF
--- a/bambi/families/family.py
+++ b/bambi/families/family.py
@@ -1,6 +1,6 @@
 from statsmodels.genmod import families as sm_families
 
-from .link import Link
+from bambi.families.link import Link
 
 STATSMODELS_FAMILIES = {
     "bernoulli": sm_families.Binomial,
@@ -28,13 +28,31 @@ class Family:
     Parameters
     ----------
     name : str
-        Family name.
+        The name of the family. It can be any string.
     likelihood: Likelihood
-        A ``Likelihood`` instace specifying the model likelihood function.
+        A ``bambi.families.Likelihood`` instace specifying the model likelihood function.
     link : str or Link
-        The name of the link function, or the function itself, transforming the linear model
-        prediction to the mean parameter of the likelihood. If a function, it must be able to
-        operate over theano tensors rather than numpy arrays.
+        The name of the link function or a ``bambi.families.Link`` instance. The link function
+        transforms the linear model prediction to the mean parameter of the likelihood funtion.
+
+    Examples
+    --------
+    Some imports
+
+    >>> from bambi import Family, Likelihood, Prior
+
+    Replicate the Gaussian built-in family.
+
+    >>> sigma_prior = Prior("HalfNormal", sigma=1)
+    >>> likelihood = Likelihood("Gaussian", parent="mu", sigma=sigma_prior)
+    >>> family = Family("gaussian", likelihood, "identity")
+    >>> # Then you can do
+    >>> # Model("y ~ x", data, family=family)
+
+    Replicate the Bernoulli built-in family.
+
+    >>> likelihood = Likelihood("Bernoulli", parent="p")
+    >>> family = Family("bernoulli", likelihood, "logit")
     """
 
     def __init__(self, name, likelihood, link):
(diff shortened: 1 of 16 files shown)


bambi/families/family.py
```diff
@@ -28,13 +28,31 @@ class Family:
...
     link : str or Link
-        The name of the link function, or the function itself, transforming the linear model
-        prediction to the mean parameter of the likelihood. If a function, it must be able to
-        operate over theano tensors rather than numpy arrays.
+        The name of the link function or a ``bambi.families.Link`` instance. The link function
+        transforms the linear model prediction to the mean parameter of the likelihood funtion.
+
+    Examples
+    --------
+    Some imports
```
canyon289, line 40: Canonical imports?

Some imports isnt all that informative

bambi/families/family.py
```diff
@@ -92,7 +110,7 @@ def __repr__(self):
 }
 
 
-def extract_family_prior(family, priors):
+def _extract_family_prior(family, priors):
```
canyon289, line 113: Is the idea that the user wont be using this method?

canyon289:
> I'm double-quotes, like "string", in all the docstrings because that is what Black, our formatter, does. I'm not opposed to
 > using single quotes, but we should be consistent. Any strong opinion in favor or against any of these alternatives?

I dont have a strong opinion

> I'm referencing classes as bambi.families.Likelihood instead of just Likelihood. I've seen it in another repo and I think it is a > good practice because it reduces ambiguity. But I would like to know if this is too much redundancy.

Is this question for the code or the docstrings? 
In regards to docstrings I tend to like it because for users it makes it easy to see where things are. For code its fine either way, as I think IDEs make it pretty easy to find the class reference.

> I'm completely in favor of promoting Aesara. However, we're not using Aesara/Theano names consistently. We are using > Aesara in the paper and in some parts of our docs, but Bambi actually uses Theano (e.g. see the link inv functions in the PyMC3 backend). I have two ideas:
>         Use the name Theano everywhere, but be very explicit we're going to use Aesara when PyMC3 v4 is ready.
>         Try to use Aesara as it is? I'm not following its development, so I'm not sure if it supports the things we currently use from Theano. If that's the case, I would drop Theano and use Aesara from now.

I agree its an awkward spot right now with the two names, Osvaldo and.I struggled with this in our book. My advice would be to bias to using Aesara because that were things are going, even though it comes at the expense of being exactly correct today.

bambi/families/family.py
tomicapretto: Right, it is only used internally and I don't see why I a user would use it right now

bambi/families/family.py
aloctavodia: The comment can even be omitted. And I think is better to use `import bambi as bmb` in all examples (docstrings, notebooks, papers, etc).

aloctavodia:
We are not yet using Aesara and thus docstrings and examples should mention Theano and not Aesara. Otherwise it can be potentially confusing for many user that do not need or want to know what Theano or Aesara are. For the paper we may add a line saying something like "in the future we will use Aesara as PyMC3 is going to use it" but probably that is something we do not really want to discuss in the paper (we currently have a mention, but we can safely remove it)

bambi/families/family.py
tomicapretto: You're both right, I don't know what made me think that sentence was adding something haha

tomicapretto:
Updated Pylint to version 2.10.2, which is what was being installed in the GA. This PR was having trouble with the new `unspecified-encoding` check that was added recently.

aloctavodia (approved):

PR #403, opened 2021-08-31
Change default priors for the 't' family

tomicapretto:
At the moment, we're using `nu` and `lam` to parametrize the `StudentT` likelihood. I've realized we should use `sigma` instead of `lam`, and use the default prior we use for the `"gaussian"` family, i.e. sigma ~ HalfStudentT(nu=4, sigma=sd(y)).

I'm still not sure what we should use for `nu` (which is currently set to a constant value!!!). I've placed a Gamma(2, 2), but this isn't actually something I've thought deeply about.

### DIFF
--- a/bambi/defaults/defaults.py
+++ b/bambi/defaults/defaults.py
@@ -10,7 +10,7 @@ SETTINGS_DISTRIBUTIONS = {
     "Binomial": {"n": 1, "p": 0.5},
     "Cauchy": {"alpha": 0, "beta": 1},
     "Flat": {},
-    "Gamma": {"alpha": 2, "beta": 2},
+    "Gamma": {"alpha": 2, "beta": 0.1},
     "HalfCauchy": {"beta": 1},
     "HalfFlat": {},
     "HalfNormal": {"sigma": 1},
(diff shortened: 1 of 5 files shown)


aloctavodia:
Do you mean stats.gamma(2, scale=2)? 

Kruschke recommends stats.expon(scale=30) or sometimes a shifted version stats.expon(scale=29) + 1, something like stats.gamma(2, scale=20) is similar but with less mass on lower values, but also a lighter tail.

See also https://github.com/stan-dev/stan/wiki/Prior-Choice-Recommendations#prior-for-degrees-of-freedom-in-students-t-distribution

twiecki:
As that parameter is so difficult to estimate, and we are rarely interested in the exact value but care more about the robustness aspect, we usually set it to something like `7`. Food for thought.

tomicapretto:
> As that parameter is so difficult to estimate, and we are rarely interested in the exact value but care more about the robustness aspect, we usually set it to something like `7`. Food for thought.

Hi Thomas, thanks for coming with this idea. I've been chatting about this with @aloctavodia. I suggested we could have something like `family="tn"` where `n` is the fixed degrees of freedom you want for the Student's T distribution (e.g. `family="t7"`), but Osvaldo raised some concerns about the API and some edge cases such as `"tinf"` or `"t1"`. So we agreed we're going to keep the prior, but we're also going to add an example showing how to make `nu` constant (which is just passing a constant value for the prior).

PR #407, opened 2021-09-03
Attempting to add a CITATION.cff file

tomicapretto:
I've tested the following advice [here](https://github.com/citation-file-format/citation-file-format#validation-heavy_check_mark) and it works, but I'm not sure if the output is going to be good.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -14,16 +14,11 @@ authors:
   - family-names: "Osvaldo"
     given-names: "Martin"
 title: "Bambi: A simple interface for fitting Bayesian linear models in Python"
-version: 0.6.1
-date-released: 2021-08-24
+version: 0.8.0
+date-released: 2022-05-18
 url: "https://github.com/bambinos/bambi"
 preferred-citation:
   type: article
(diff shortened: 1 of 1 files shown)


tomicapretto:
Hmm, the generated bibtex looks like

```
@article{Tomas_Bambi_A_simple,
author = {Tomas, Capretto and Camen, Piho and Ravin, Kumar and Jacob, Westfall and Tal, Yarkoni and Osvaldo, Martin},
title = {{Bambi: A simple interface for fitting Bayesian linear models in Python}}
}
```

I'll keep trying

tomicapretto:
I think this is not going to work out of the box right now. The following are going to be useful

* https://github.com/citation-file-format/citation-file-format/issues/254
* https://github.com/precice/fenics-adapter/pull/137 some people with a similar use case

aloctavodia:
It can become annoying if we need to keep updating the release version and date. Maybe we can make that automatic. Or maybe we don't need to worry about this if the preferred citation omits that information.

tomicapretto:
> 

I would keep the citation file in sync with the version used to write the paper. So I don't think we need to update it and what we already have is good.

PR #416, opened 2021-10-13
B-Spline example with cherry blossoms dataset

### LINKED ISSUE #413: Add splines example.
One option could be to replicate this example pymc-devs/pymc-examples#232

tomicapretto:
Example using splines. Closes #413


tomicapretto:
@aloctavodia I would like to have your input about the following issue.

The original model in McElreath's book, and its replication in PyMC, are equivalent to the following model in Bambi

```python
bmb.Model("doy ~ bs(year, knots=iknots, intercept=True)", ...)
```

which is the model you can see in the notebook right now.

However, one problem with this model is that the intercept is spanned twice. First, with the regular intercept that we add to linear predictors, the column of ones, alpha in the original model notation. And secondly, as the sum of the columns in the basis

```python
B = model._design.common["bs(year, knots = iknots, intercept = True)"]
np.allclose(B.sum(1), 1)
# True
```

So there's a redundancy in the model specification. If you run the model several times you'll notice it impacts on the number of effective samples, the sampling time, and also in r_hat.

Two obvious alternatives come to my mind

1. Do not span intercept in `bs()`

```python
bmb.Model("doy ~ bs(year, knots=iknots)", ...)
```

2. Do not include the regular intercept

```python
bmb.Model("doy ~ 0 + bs(year, knots=iknots, intercept=True)", ...)
```

As far as I've tried, the second approach works better because it is more flexible. Try fitting the original model and the second alternative and they look the same.

What do you think? Should we stick to the original model, or use one of the alternatives?

tomicapretto:
@aloctavodia I know this example is not 100% finished but I would like to have a second pair of eyes on what's done so far. I'm especially interested in feedback about the explanations between the blocks of code. The ones in the last section aren't polished at all.

Issue #318, opened 2021-03-23
Feature request: Multinomial regression


ricardoV94:
Using the softmax link for categorical outputs.

tomicapretto:
Hi @ricardoV94, thanks for the suggestion! It makes total sense. If nobody handles this within the next week, I will work on it.

Thanks again for the idea!

jlevy44:
Thanks!

ricardoV94:
By the way, there was some discussion in a pymc3/resources issue about multinomial models, specially on the difference between category intercepts (where you should have at most k-1 coefficients), and shared and non-shared coefficients for predictors in multinomial models (where you can have one or k coefficients): https://github.com/pymc-devs/resources/issues/117#issuecomment-714343532

I found the two references there very helpful to understand these types of GLM. Most Python libraries offer only a limited subset of multinomial models (e.g., non-shared coefficients for k-1 categories).

tomicapretto:
@ricardoV94 thanks for the reference! I'm going to have a deeper look at it when working on this. Unfortunately, I had to work on other things in Bambi that were more urgent. We're planning to do a release after #349 is merged, and then multinomial models will be one of the first things (or the first thing) we are going to do.

tomicapretto:
@ricardoV94 I don't know if you know [Categorical Data Analysis by Alan Agresti](https://www.wiley.com/en-us/Categorical+Data+Analysis%2C+3rd+Edition-p-9780470463635). I used that book in college and was really helpful. I'm reading it again to refresh this multinomial regression stuff before coding.

ricardoV94:
Yeah Agresti is a classic. Haven't heard about the second one.

On a practical standpoint I think the R library (second link in the referenced comment I shared before) is a reasonable way to specify this type of models.

It something like `y ~ X1 | X2 | X3` where X1 to X3 correspond to these different types of coefficients (shared[1], distinct[k-1], distinct[k]):

* X1 are used for cases where in each observation the different categories have different values for the same variable and you want to weight them on the same scale (e.g. price in dollars)
* X3 are like X1 but you want to use different weights for the different categories (e.g. travel duration by airplane vs travel duration by bus).
*X2 (which should probably be the first) are for those variables that vary across observations but not across categories (e.g. the socioeconomic status of the person making the choice). Intercepts are part of this category.

huangziwei:
thank you very much for this new update! Will it be available on PyPi soon?

tomicapretto:
[...]

PR #426, opened 2021-11-02
Add categorical family

tomicapretto:
It's been a while since I created the `multinomial` branch. Now I'm trying to finish this implementation. I'm not opening this PR because I think this is close to be ready, but because I want to use it as a space to share ideas.


tomicapretto:
Looks like there is the first version working. But I'm still quite unsure whether this implementation is correct or not -- specially the part where I need to adjust the intercept for the centering. Let's see the following example

```python
import arviz as az
import bambi as bmb
import numpy as np
import seaborn as sns
import pymc3 as pm
import theano.tensor as tt
import formulae


iris = sns.load_dataset('iris')

priors = {"common": bmb.Prior("Normal", mu=0, sigma=2)}
model = bmb.Model(
    "species ~ sepal_length + sepal_width + petal_length + petal_width", 
    iris, 
    family="multinomial",
    priors=priors
)
idata = model.fit()

az.summary(idata)
```
image


```python
az.plot_trace(idata);
```

posterior


Then, we can fit the same model with PyMC

```python
y_s = pd.Categorical(iris['species']).codes
x_n = iris.columns[:-1]
X = iris[x_n].values
X = X - X.mean(axis=0)
X = np.hstack([np.ones((150, 1)), X])

oords = {"species": ["versicolor", "virginica"]}
with pm.Model(coords=coords) as model_sf:
    β0 = pm.Normal('β0', mu=0, sd=2, dims="species")[None, :]
    β_sl = pm.Normal('β_sl', mu=0, sd=2, dims="species")[None, :]
    β_sw = pm.Normal('β_sw', mu=0, sd=2, dims="species")[None, :]
    β_pl = pm.Normal('β_pl', mu=0, sd=2, dims="species")[None, :]
    β_pw = pm.Normal('β_pw', mu=0, sd=2, dims="species")[None, :]
    
    β = tt.concatenate([β0, β_sl, β_sw, β_pl, β_pw])
    μ = tt.concatenate([np.zeros((150, 1)), pm.math.dot(X, β)], axis=1)
    p = tt.nnet.softmax(μ)

    yl = pm.Categorical('yl', p=p, observed=y_s)

    idata_pm = pm.sample(1000, return_inferencedata=True)

az.plot_trace(idata_pm)
```
posterior_pymc


Note the separation between the adjusted intercepts (first traceplot) and the separation between the un-adjusted intercepts (the last traceplot). I'm not sure if that is expected or not. https://github.com/bambinos/bambi/blob/f15ce460cddd27f737ed47d66fa233b4746b6b10/bambi/backend/pymc.py#L284-L302 is where I adjust the intercept.

ricardoV94:
Glancing at it, it seems to be correct. I am not sure what is going on with the intercepts in bambi and brms, I guess it's just a transformation that makes things more stable / interpretable? If it is just a one-to-one transformation it should not really affect sampling.

One question, should the family reflect how the observations are encoded? Categorical for label encoding and Multinomial for one-hot encoding? I am not sure how this maps with the rest of bambi API

tomicapretto:
> Glancing at it, it seems to be correct. I am not sure what is going on with the intercepts in bambi and brms, I guess it's just a transformation that makes things more stable / interpretable? If it is just a one-to-one transformation it should not really affect sampling, and it seems it does not...
> 
> One question, should the family reflect how the observations are encoded? Categorical for label encoding and Multinomial for one-hot encoding? I am not sure how this maps with the rest of bambi API

1. You're right. The transformation is simply centering predictors to improve stability.
2. Thanks for the suggestion. Could you clarify a little on the Multinomial one? I understand the Categorical name, but I'm not that sure what you mean with Multinomial. Do you mean something like `cols(c1, c2, c3) ~ predictors` where `c1`, `c2` and `c3` are the one-hot encoded variables?

ricardoV94:
>     2. Thanks for the suggestion. Could you clarify a little on the Multinomial one? I understand the Categorical name, but I'm not that sure what you mean with Multinomial. Do you mean something like `cols(c1, c2, c3) ~ predictors` where `c1`, `c2` and `c3` are the one-hot encoded variables?

Yes. That is the type of data that the PyMC multinomial distribution expects

tomicapretto:
> > ```
> > 2. Thanks for the suggestion. Could you clarify a little on the Multinomial one? I understand the Categorical name, but I'm not that sure what you mean with Multinomial. Do you mean something like `cols(c1, c2, c3) ~ predictors` where `c1`, `c2` and `c3` are the one-hot encoded variables?
> > ```
> 
> Yes. That is the type of data that the PyMC multinomial distribution expects

That makes sense, and I think that is how brms works (i.e. two families for these two different cases). But we still don't support multiple variables in the response term such as in `c(c1, c2, c3)`. 

So far, Bambi works with pandas DataFrames. For this case, it expects a column with a categorical variable with more than two levels and then Bambi handles the rest. Do you think it is worth considering the case where people have their categorical variable split in several one-hot encoded columns? I mean, why people would have that when it is easier to pass one column as is without having to one-hot encode?

Anyway, if it is the case we find a good reason to have multiple variables in the response such as in `c(c1, c2, c3)`, it is something we could handle with low-medium effort.

ricardoV94:
> I mean, why people would have that when it is easier to pass one column as is without having to one-hot encode?

In theory (not that I have ever seen it), you could have observations in the multinomial that do not add up to 1 such as `c[0] = (5, 0, 1)`, which would require duplicated rows in a categorical implementation.

I wouldn't make this high priority. But perhaps change the family name to `Categorical` to reflect that the input should be formatted as such?

tomicapretto:
> > I mean, why people would have that when it is easier to pass one column as is without having to one-hot encode?
> 
> In theory (not that I have ever seen it), you could have observations in the multinomial that do not add up to 1 such as `c[0] = (5, 0, 1)`, which would require duplicated rows in a categorical implementation.
> 
> I wouldn't make this high priority. But perhaps change the family name to `Categorical` to reflect that the input should be formatted as such?

Yes, that makes sense. For now, let's rename it to "categorical", but keep the rest as it is. In the future, we can discuss whether we want a multinomial family or not. 

Thanks for the ideas Ricardo! :D

ricardoV94:
Any plans about allowing for the more sophisticated categorical/multinomial models described in #318, where you can have different predictors for each category?

tomicapretto:
I don't want to say no because the idea is very cool. But there are so many things to do before it, that now I think that is not a priority. For example, we don't support multi-part formulas at all. So there we have a substantial amount of work to support that both in the parser and in Bambi. Then, we would also need to think about a good way to build the underlying PyMC model, and I think that means writing/modifying a lot of code.

the short answer is that I hope we achieve that at some point in time, but it is not something we're going to do in the near future.

ricardoV94:
Sad but understandable! :D Great work 👍

tomicapretto:
I got stuck. Let me describe the situation.

We can build and fit models of the categorical/multinomial family. For this, we use one of the levels of the response as reference as in the example in one of the comments above. In that example we have `versicolor`, `virginica`, and `setosa`, where `setosa` is the level taken as reference. For this reference, the coefficients are all 0. Since they are 0, they're not included in the InferenceData object we return. So far, so good.

The problems arose when I started to work on the implementation of the `Model.predict()` method for this family. After some attempts I managed to work out the shape handling. 

But there's another problem. The `species_coord` coordinate only contains the  `versicolor` and `virginica` entries because these are the ones with non-zero coefficients. However, predictions must be made for the three species.

* Should we have one `species_coords_trimmed` and other `species_coords`, the first with only the levels with non-zero coefficients and the second with all the levels?
* Should we include the zero coefficients? This goes against what we are doing for all the other families and is also a problem when using things like `az.plot_trace()`.
* All the shape handling and checks are getting quite hacky in my opinion. I think I should take some more time to think about how to build this. Right now, this family is the only family that has a multivariate response. But in the future we may want to have other families with multivariate response, so it's better to think well about how to design it so we don't end up creating a mess of code.

tomicapretto:
Again, looks like this starts to work. See the following toy example:

```python
import arviz as az
import bambi as bmb
import numpy as np
import pandas as pd

rng = np.random.default_rng(1234)
x = np.hstack([rng.normal(m, s, size=50) for m, s in zip([-2.5, 0, 2.5], [1.2, 0.5, 1.2])])
y = np.array(["A"] * 50 + ["B"] * 50 + ["C"] * 50)

colors = ["C0"] * 50 + ["C1"] * 50 + ["C2"] * 50
plt.scatter(x, np.random.uniform(size=150), color=colors);
```
plot

```python
data = pd.DataFrame({"y": y, "x": x})
model = bmb.Model("y ~ x", data, family="categorical")
idata = model.fit()
```

```python
x_new = np.linspace(-5, 5, num=200)
model.predict(idata, data=pd.DataFrame({"x": x_new}))
p = idata.posterior["y_mean"].values.reshape((2000, 3, 200))

for j in range(3):
    plt.plot(x_new, p[::10, j, :].T, color=f"C{j}", alpha=0.2)
```
plot2

tomicapretto:
Another example 😁

The original example is this one in R

```r
library(VGAM)

Income = c(3,10,20,35)
Diss = c(2,2,0,0)
Little = c(4,6,1,3)
Mod = c(13,22,15,13)
Very = c(3,4,8,8)
jobsat.fit1 = vglm(cbind(Diss,Little,Mod,Very) ~ Income, family=multinomial)
```

and the predictions are 

image

Now, in Python and Bambi

```python
incomes = [3, 10, 20, 35]
responses = {
    "Dissatisfied": [2, 2, 0, 0], 
    "Little satisfied": [4, 6, 1, 3],
    "Moderate satisfied": [13, 22, 15, 13],
    "Very satisfied": [3, 4, 8, 8]
}

# Create data frame
income = []
response = []
for r, c in responses.items():
    for i, count in enumerate(c):
        income += [incomes[i]] * count
        response += [r] * count
        
data = pd.DataFrame({"income": income, "response": response})

# Create model
priors = {"Intercept": bmb.Prior("Normal", mu=0, sigma=1), "income": bmb.Prior("Normal", mu=0, sigma=0.1)}
model = bmb.Model("response ~ income", data, family="categorical", priors=priors)
idata = model.fit()

# Obtain predictions
new_income = np.linspace(0, 60)
new_data = pd.DataFrame({"income": new_income})
model.predict(idata, data=new_data)
p = idata.posterior["response_mean"].values.reshape((2000, 4, 50))


# Plot predictions
for j, label in enumerate(responses):
    plt.plot(new_income, p[:, j, :].T.mean(1), color=f"C{j}", lw=2, label=label)
    az.plot_hdi(new_income, hdi_data=np.quantile(p[:, j, :], [0.05, 0.94], axis=0).T, color=f"C{j}")

plt.hlines(y=0, xmin=0, xmax=60, color="black", ls="--")
plt.legend(loc=2)


```
plot3

tomicapretto:
Everything we need seems to be working now. Two more examples.

**Example 1**

Using numerical and categorical predictors. Data were taken from Categorical Data Analysis from Alan Agresti. 

```python
length = [
    2.72, 2.79, 2.84
]

choice = [
]

    ordered=True
)

model = bmb.Model("choice ~ length + sex", data, family="categorical")
idata = model.fit()


new_length = np.linspace(1, 4)
new_data = pd.DataFrame({"length": np.tile(new_length, 2), "sex": ["Male"] * 50 + ["Female"] * 50})
model.predict(idata, data=new_data)
p = idata.posterior["choice_mean"].values.reshape((2000, 3, 100))

fig, axes = plt.subplots(1, 2, figsize=(12, 5))

choices = ["Other", "Invertebrates", "Fish"]
for j, label in enumerate(choices):
    axes[0].plot(new_length, p[:, j, :50].T.mean(1), color=f"C{j}", lw=2)
    axes[1].plot(new_length, p[:, j, 50:].T.mean(1), color=f"C{j}", lw=2)
    
    az.plot_hdi(new_length, hdi_data=np.quantile(p[:, j, :50], [0.05, 0.94], axis=0).T, color=f"C{j}", ax=axes[0])
    az.plot_hdi(new_length, hdi_data=np.quantile(p[:, j, 50:], [0.05, 0.94], axis=0).T, color=f"C{j}", ax=axes[1])

axes[0].set_title("Male")
axes[1].set_title("Female")

handles = [Line2D([], [], color=f"C{j}", label=choice) for j, choice in  enumerate(choices)]

fig.subplots_adjust(left=0.05, right=0.975, bottom=0.075, top=0.85)

fig.legend(
    handles, 
    choices,
    loc="center right",
    ncol=3,
    bbox_to_anchor=(0.99, 0.95),
    bbox_transform=fig.transFigure
)
```

plot4

**Example 2**

Using the [inhaler](https://rdrr.io/cran/brms/man/inhaler.html) dataset from the brms package. A model with a group-specific intercept.

```python
data = pd.read_csv("inhaler.csv")
data["rating"] = pd.Categorical(data["rating"], categories=[1, 2, 3, 4])

model = bmb.Model("rating ~ period + carry + treat + (1|subject)", data, family="categorical")
idata = model.fit(chains=2, cores=2)

az.plot_trace(idata)
```

plot5


Let's see some predictions

```python
d = data.drop_duplicates(["carry", "treat", "period"])
model.predict(idata, data=d)
p = idata.posterior.rating_mean.values.reshape((2000, 4, 4))

fig, axes = plt.subplots(2, 2, figsize=(8, 6), sharex=True)

for subject_idx in range(4):
    ax = axes.ravel()[subject_idx]
    for rating in range(4):
        x = p[:, rating, subject_idx]
        az.plot_kde(x, plot_kwargs={"color": f"C{rating}"}, ax=ax)
        ax.set_ylim(0, 60)
```
plot6

Some of the examples in this PR could be made a proper example in our docs by the way.

The **next step** is to add tests.

PR #435, opened 2022-01-07
Implement `Model.set_alias()`

### LINKED ISSUE #432: Add `.set_alias()` method
Our `Model` class could gain a `.set_alias()` method that can be used when your variables contain long names and you want to have shorter/better versions in tables and plots. For example, we could have something such as:

```python
model  = bmb.Model("house_size + house_size:neighborhood")
```

and then

```python
model.set_alias({"Intercept":  "α", "house_size": "β", "neighborhood": "γ"})
```

Then all the plots and summaries would show `"α"`, `"β"`, and `"γ"` instead of the original names.

tomicapretto:
This PR implements the `Model.set_alias()` method.

* [x] Works for common terms
* [x] Works for group-specific terms
* [x] Works for response terms (even though it's not used in the output, but it's useful for graph representations)
* [x] Works for auxiliary parameters of the family (i.e. we can use an alias for the `sigma` parameter in linear regression)
* [x] Works for auxiliary parameters of group-specific terms.

```python
data = pd.DataFrame(
    {
        "y": np.random.normal(size=100),
        "predictor": np.random.normal(size=100),
        "group": np.random.choice(["A", "B", "C", "D"], size=100)
    }
)

model = bmb.Model("y ~ predictor + (predictor|group)", data)

aliases = {
    "Intercept": "α",
    "predictor": "β",
    "1|group": "α_group",
    "predictor|group": "β_group",
    "sigma": "σ"
}
model.set_alias(aliases)
idata = model.fit()

az.summary(idata)
```
image

As can be seen, we still have `α_group_sigma` and `β_group_sigma`. I want to see if it is possible to add aliases for these too . The only alternative that comes to my hand is to re-use the `"sigma": "σ"`

This closes #432


aloctavodia (approved):

OriolAbril:
note on [arviz.labels](https://arviz-devs.github.io/arviz/user_guide/label_guide.html) (feel free to tag me to review or update the plotting section of an example using these aliases). After using the code from the description:

```python
model = bmb.Model("y ~ predictor + (predictor|group)", data)

aliases = {
    "Intercept": "α",
    "predictor": "β",
    "1|group": "α_group",
    "predictor|group": "β_group",
    "sigma": "σ"
}
model.set_alias(aliases)
idata = model.fit()

az.summary(idata)
```

The pymc model and consequently the inferencedata variables have these aliases as names. I believe that in general people will want to use those names for coding and for some arviz functions like summary which is what happens, ArviZ defaults to using the exact names and coordinate values in the passed objects. 

However, there might be some cases, probably plots where there is some more space and might need to be shown to external people, where we want to show the "original" names.  This can be done with the MapLabeller:

```
labeller = arviz.labels.MapLabeller(var_name_map={v: k for k, v in aliases})
az.plot_posterior(idata, labeller=labeller)
```

tomicapretto:
@OriolAbril thanks for the comments. I'm not very familiar with labellers but this seems to be something we should include in one example with aliases. 

Will keep you posted if I can do that!

Issue #439, opened 2022-01-11
Port PyMC Radon example


tomicapretto:
[A Primer on Bayesian Methods for Multilevel Modeling](https://docs.pymc.io/projects/examples/en/latest/case_studies/multilevel_modeling.html) is a very popular example in PyMC. 

We could port this example to show how easy is to create those types of hierarchical models using Bambi.

juanitorduz:
Hey! I could work on this. However, the PyMC example is quite long ... shall we mainly focus in explaining how the hierarchical models can be written in `bambi` (which is a good learning for me)? We do not wanna duplicate the material right?

juanitorduz:
I started with a draft PR: https://github.com/bambinos/bambi/pull/440 Can someone take a look into the models specification please? Thanks 😄 !

juanitorduz:
@tomicapretto @aloctavodia Thank you guys for your support! Many learnings and interesting discussions!

tomicapretto:
@juanitorduz  It's been a very interesting experience for me too! Thanks a lot!

Issue #357, opened 2021-05-19
How to implement predictions


tomicapretto:
I would like to implement a `Model.predict()` method. My original idea was to replicate `predict()` in R, where we pass a data frame and obtain predictions for the mean response. However, @aloctavodia helped me to realize we should consider both predictions for the mean response and predictions for the response variable (like what is being done with `Model.posterior_predictive()` but also available for out-of-sample data).

So, what I have in mind is something as follows 

```
predictions = model.predict(idata, kind, xdata=None)
```

where `kind = "mean"` returns predictions in terms of the mean response (i.e. g^-1(x'b)) and `kind="response"` returns the posterior predictive distribution. If `xdata` is a data frame, predictions are made using the new data set (out of sample predictions). If not, predictions are made using the original sample (in sample predictions).

What I would like to discuss

* We already have a `Model.posterior_predictive()` method that samples the posterior predictive distribution, using the observed values for the predictors. There's overlap between this method and the one I want to implement. Should we keep both? Only one? I don't want to have two methods doing something similar (or the same thing), but we have to be aware dropping `Model.posterior_predictive()` is another API change and can break existing code. Maybe a deprecation warning?
* Argument names and default behavior.
* Anything you would like to add to this discussion.

Linking #356, where this feature is requested.

ioannis12:
hello, 

congrats for the work. 
I have a model,  model = Model('A ~ 0 + B + C  + D +  (1|E)', f_delta, family = 'bernoulli')
I would like to make predictions on test set, any idea how the predict function looks like?
There is a predict function in the neg binomial example but is not for mixed effects models. 

cheers

tomicapretto:
Hi @ioannis12 

This is an example of how you could do it for mixed-effects models. I used simulated data so it may differ from yours. 


```python
import bambi as bmb
import numpy as np
import pandas as pd

from scipy.special import expit
```

My simulated data

```python
f_delta = pd.DataFrame({
    "A": np.random.choice([0, 1], size=100),
    "B": np.random.normal(size=100),
    "C": np.random.normal(size=100),
    "D": np.random.normal(size=100),
    "E": ["Group 1"] * 25 + ["Group 2"] * 25 + ["Group 3"] * 25 + ["Group 4"] * 25
})
```

Same model

```python
model = bmb.Model('A ~ 0 + B + C + D + (1|E)', f_delta, family = 'bernoulli')
idata = model.fit()
```

[...]

PR #440, opened 2022-01-14
Radon Example Implementation

juanitorduz:
This PR aims to solve https://github.com/bambinos/bambi/issues/439. 

I have pushed the first version of the models. Here are the formulas I used: 

- pooled_model: `log_radon ~ floor`
- unpooled_model: `log_radon ~ 0 + county + county:floor`
- partial_pooling_model: `log_radon ~ 0 + (0 + 1|county)`
- varying_intercept_model: `log_radon ~ 0 + (1|county) + floor` # ! How to do complete one-hot-encoding for county?
- varying_intercept_slope_model: `log_radon ~ 0 + (floor|county)` # ! How to do complete one-hot-encoding for county?

I could like to have all the countries as encoded (not removing the first one) plus I am not auto-scaling the data to make the results with the original PyMC example comparable. Does this look good? I am still getting used to the formula-like notation for hierarchical models (which is why I am working on this PR 😅 ). 

After double-checking the model specification I will continue to add text and some additional plots.


tomicapretto:
@juanitorduz thanks for this massive contribution! 

I've added several comments to the notebook. I think you've made a lot of progress! 

As you may see in the comments, I've found two bugs associated with Bambi. One is related to hyperpriors and the other is associated with a varying effect when using a categorical predictor. I think I have to fix the second issue before we can move on with this example.

Again, thanks so much!

tomicapretto:
I have found the problem. It's a shape problem in the underlying PyMC model. We're basically adding things with shape `(n,)` and `(n, 1)` and PyMC does not like that.

I'll create a new PR with the fix and make a new patch release. I have introduced this bug accidentally when modifying other underlying code I guess. I'll also add a test so this does not happen again :sweat_smile:

tomicapretto:
The shape problem is solved in the new release Bambi 0.7.1. 

I could push my changes to the notebook so you don't have to write them (if you want)

juanitorduz:
> The shape problem is solved in the new release Bambi 0.7.1.
> 
> I could push my changes to the notebook so you don't have to write them (if you want)

Hey @tomicapretto thank you very much on your feedback! Here are some comments regarding your points:
1. It totally makes sense to encode `floor` as ["Basement", "Floor"] categorical variable 👍 .
2. I actually did know know how to pass the hyper-priors to Bambi 😅 ! I do not know one could use `Prior` inside the `priors` dictionary in the `Model` object! I spend a lot of time trying to figure it out how to add it somehow to the formula. Hence, in my opinion, this Radon example is going to be super useful for the documentation. Specially for the ones comping from `pymc`.

Feel free to push your changes to the notebook! Just push whatever you have ready and I will work on top of it the open comments 💪 ! 

About the bug you discovered (I was also a bit surprised I could pass the prior as I did without getting an error... sometimes things are too good to be true 😆), the fix was merged in https://github.com/bambinos/bambi/pull/441 right? Which means I would need to re-install `bambi` from the `main` branch right?

tomicapretto:
@juanitorduz 

1. Great!
2. We prevent hyperpriors for common terms (you can't have a hyperprior there) but we missed the case where you don't pass a hyperprior for a term that actually needs it.

I've quickly made a new patch release to solve this issue because it may be affecting other people too :fearful:.  I'm about to push my changes now.

juanitorduz:
<p>Makes sense! Solved in https://github.com/bambinos/bambi/pull/440/commits/1cf6df3fa259dde97921045f9504fbeba89a22d7</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>This was added in https://github.com/bambinos/bambi/pull/440/commits/1cf6df3fa259dde97921045f9504fbeba89a22d7</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>Thanks for the input! </p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>I would actually suggest to add some of this comments (and the reference) to the notebook.</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>As for future reference, the bug was solved in https://github.com/bambinos/bambi/pull/441</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>Now everything make sense!</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>Thanks for adding it to the notebook!</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
Hi! Here I want to submit the first draft of the example implementation for review. It is still not complete but I just want to get feedback on the model specifications, if the explanations are clear and if you would like me to add more diagnostics or plots to each model. 

Here is what remains to be done (feel free to add more items if I missed something):

**Must do next:**

- [ ] Write introduction.
- [ ] Explain data and data processing steps.
- [ ] Write conclusions 

**Nice to have:**
- [ ] Posterior predictive distribution. In particular I would like to do it for the hierarchical models to illustrate the *shrinkage to the mean* phenomenon. However I have not been able do do it with the simple `predict` method. For example `varying_intercept_slope_model.predict(idata=varying_intercept_slope_results)` gives me a `ValueError`....is this related to  https://github.com/bambinos/bambi/issues/357 ?
- [ ] Add a model comparison table  using `az.compare`.
- [ ] I think the covariance model (`LKJCholeskyCov` prior) would be nice to add even if its work in progress.

juanitorduz:
<p>Added!</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

tomicapretto:
Thanks for the huge progress made! I owe you a better review related to the model specification. I'll try to do it tomorrow :D

juanitorduz:
> Thanks for the huge progress made! I owe you a better review related to the model specification. I'll try to do it tomorrow :D

No rush! I will probably come back to this PR at the end of the week :)

juanitorduz:
<p>I have added it to the notebook.</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>:) </p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>Goof point! I have change the plot titles.</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

tomicapretto:
@juanitorduz I've added several inline comments to try to clarify the model and related things. I'm still not done with that, but feel free to review and/or ask questions!

juanitorduz:
Thank for your comments and great additions @tomicapretto ! In https://github.com/bambinos/bambi/pull/440/commits/e81a22758a449668a1142995f1769c196f65fb9a I addressed all your comments and included edits. In addition, I added the posterior predictive analysis on the original scale for the pooled model. I think the issue was that we need to consider the sample variance. Let me know what you think about it. 

What do do think it remains to be done? 

As a side remark, I am having problems getting predictions with the `predict` method. Could you maybe add an example of this which we can add to all the models? Thanks :)

tomicapretto:
> As a side remark, I am having problems getting predictions with the `predict` method. Could you maybe add an example of this which we can add to all the models? Thanks :)

I'm not completely sure I understand what your problem is. But for example, let's say you want to generate the posterior predictive for the first model (which is something you're doing by hand right now)

```python
# Get samples from the posterior predictive distribution
pooled_model.predict(pooled_results, kind="pps")

# Stack chains and draws. pps stands for posterior predictive samples
pps = pooled_results.posterior_predictive.stack(sample=("chain", "draw"))["log_radon"].values

print(pps.shape) # (919, 2000) which is (n_observations, n_chain * n_draw)

# Inverse transformation
pps = np.exp(pps) - 0.1

# We need to subset pps according to the Floor level.
# NOTE: I think it would be good to reset index when we manipulate 'srrs_mn' in the very start
floor = srrs_mn["floor"].reset_index(drop=True)

# Feel free to change if you know a better way of doing it
basement_idx = np.where(floor == "Basement")[0]
floor_index = np.where(floor == "Floor")[0]

# The plot, almost exactly same code than the original
fig, ax = plt.subplots(nrows=2, ncols=2, figsize=(12, 6), layout="constrained")
ax = ax.flatten()

sns.histplot(x=pps[basement_idx].flatten(), stat="density", color="C0", ax=ax[0])
ax[0].set(title="Basement (Posterior Predictive Distribution)", xlabel="radon", ylabel="Density");
sns.histplot(x="activity", data=srrs_mn.query("floor == 'Basement'"), stat="density", ax=ax[2])
ax[2].set(title="Basement (Sample Distribution)", xlim=ax[0].get_xlim(), xlabel="radon", ylabel="Density");

sns.histplot(x=pps[floor_index].flatten(), stat="density", color="C1", ax=ax[1])
ax[1].set(title="Floor (Posterior Predictive Distribution)", xlabel="radon", ylabel="Density");
sns.histplot(x="activity", data=srrs_mn.query("floor == 'Floor'"), stat="density", color="C1", ax=ax[3])
ax[3].set(title="Floor (Sample Distribution)", xlim=ax[1].get_xlim(), xlabel="radon", ylabel="Density");
```

And this is the result

image

As you can see, the posterior predictive distribution has many more values than the one you computed earlier. That's because Bambi is computing the posterior predictive distribution for all the observations in the original dataset (the one passed when we build the model). But for the model there are only two groups, then it obtains 2000 x number_of_floor_units samples for Floor and 2000 x number_of_basement_units samples for Basement. That's why the histograms above are based on so many samples.

If you just want to obtain 2000 samples for each group, you can pass a new data frame to the predict method. This new data frame has 1 observation per group.

```python
# Get samples from the posterior predictive distribution
new_data = pd.DataFrame({"floor": ["Basement", "Floor"]})
pooled_model.predict(pooled_results, kind="pps", data=new_data)

# Stack chains and draws. pps stands for posterior predictive samples
pps = pooled_results.posterior_predictive.stack(sample=("chain", "draw"))["log_radon"].values

print(pps.shape) # (2, 2000)

# Inverse transformation
pps = np.exp(pps) - 0.1

fig, ax = plt.subplots(nrows=2, ncols=2, figsize=(12, 6), layout="constrained")
ax = ax.flatten()

sns.histplot(x=pps[0].flatten(), stat="density", color="C0", ax=ax[0])
ax[0].set(title="Basement (Posterior Predictive Distribution)", xlabel="radon", ylabel="Density");
sns.histplot(x="activity", data=srrs_mn.query("floor == 'Basement'"), stat="density", ax=ax[2])
ax[2].set(title="Basement (Sample Distribution)", xlim=ax[0].get_xlim(), xlabel="radon", ylabel="Density");

sns.histplot(x=pps[1].flatten(), stat="density", color="C1", ax=ax[1])
ax[1].set(title="Floor (Posterior Predictive Distribution)", xlabel="radon", ylabel="Density");
sns.histplot(x="activity", data=srrs_mn.query("floor == 'Floor'"), stat="density", color="C1", ax=ax[3])
ax[3].set(title="Floor (Sample Distribution)", xlim=ax[1].get_xlim(), xlabel="radon", ylabel="Density");
```

And the result is

image

**PS** I say 2000 samples because I'm using 2 chains and 1000 draws per chain. That may vary on your side.

juanitorduz:
Thank you @tomicapretto ! This PR thread is becoming great documentation itself 😂 . My problem was a typo in the `predict` argument: `idata` vs `data` 🤦 ! In https://github.com/bambinos/bambi/pull/440/commits/e00fc1b0b9a49258a935d70134231c100c9b6409 I added your suggestions (to plot less samples, i.e. option 2). In addition I added a plot comparing the predicted and sample means at county level for the partial pooling model which illustrates the shrinkage phenomenon. It is nice to compare with the analogous one of the unpooled model. Let me know what you think about it 😉 !

tomicapretto:
<p>I've already changed it in my local branch which I'm about to push in  a moment</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
Hey @tomicapretto thank you for your input in the notebook 🚀 ! I my last commits I have simply re-arranged some plots (plus completed the remaining plot from the "Adding group-level predictors" section), added some comments and did some proof reading. I think it looks like a good first draft right? Feel free to edit and change whatever you feel in order to improve the current state of the notebook. What would be the next steps? What would you like me to support you with?

tomicapretto:
Hi @juanitorduz 

I've been reading throught the notebook and these are the notes I took.

**Misc notes**

I'm not sure an answer is given to the following questions in the PyMC example:

* Why we model log(radon + 0.1) and not radon?
* Why we use log(Uppm) and not Uppm?

Those are things that come from the original example by Gelman and Hill. 
I think we should figure out why they do that, and explain that in the notebook.

* There's a little inconsistency between the first plots in the EDA.
	+ First plot uses Floor -> Blue, Basement -> Orange
	+ Second plot uses Floor -> Orange, Basement -> Blue
	+ I would also like to add one sentence with an interpretation of the plot (especially the first one). Not a fancy comment, just say the distribution for basement is shifted a little to the right.

* Many notes I added were kind of personal notes (i.e. for me) or messages I wanted to pass to you. We need to check if we should remove some of them and polish the rest.

* I would like a summary section comparing the meaning of "no pooling", "complete pooling" and "partial pooling". The plot you constructed will be very benefitial here.

* We should discuss a little if it's necessary to include the `family="gaussian"` and `link="identity"` arguments. I think it's ok to include them in the first model. Below we add these are the default values and we're not including them explicitly anymore.

* I need to clarify the "Adding group-level predictors". The comment says "floor" is a group level predictor. But actually, none of them are.

**LKJ Prior**

We should write down the model (including the correlated priors). I would like to polish the underlying implementation too.


**Model comparison**

There are a couple of issues when we use loo to perform model comparison. I'm not sure if we should do model comparison here. I mean, it's great. But there we should explore what's going on with the high k-hat values. And that would make the notebook huge.


And last but not least, I would like to say I'm really thankful for your hard work here. This is a very long and complicated notebook may have not been even started if it was not for your initiative.

---

If you stil have time, I think that modifying the plots a little to have color consistency and skimming through the original examples to look for a reason for the transformations would be very helpful. 
I can start working on the rest later (I'm not with my computer until Wednesday so I don't think I'll be able to do much)

juanitorduz:
@tomicapretto thanks for your comments! In https://github.com/bambinos/bambi/pull/440/commits/06d8fa652956d398a20ce191151b42e3baddf0a8 I addressed the low-hanging-fruits of your suggestions. We can catch up when you are back! Enjoy your time  😄 !

tomicapretto:
@juanitorduz I started to review and add content to this notebook again. I'm sorry I couldn't work on this earlier but I caught COVID and I wasn't feeling very well.

Some quick notes.

* I clarified a lot the meaning of no pooling, complete pooling, and partial pooling. Let me know if you have any comments about that.
* I removed the posterior predictive plots and the khat plots. It's not because they're unnecessary, but the notebook is already too long with the content we already have. 
* I also removed the model with the LKJ prior. It is quite buggy right now and I would like to make it more robust before we use it in an example. We can add it later once the implementation is working well.
* I still need to work on the last models, where we add the uranium predictor. There are several variants there. They are all similar, but some are better than others, and some are also non-identified. 

In the coming days, I'll try to work more on this.

@aloctavodia if you want to have a read, everything before the section titled  **Varying intercept and slope** is ready.

juanitorduz:
Oh no @tomicapretto ! I hope you are recovering successfully! Take it easy (no rush wit this PR, really!). As a matter of fact, me and my family we all caught covid as well 😞 ...but we are feeling much better now.
Your changes make total sense! The notebook was indeed getting quite long 😅 .

aloctavodia:
I agree with your comments @tomicapretto We may want to have a model comparison notebook (like a continuation of this one). But we may want to wait to write that until we implement new model comparison features in ArviZ.

I left a few comments, most if not all are about the text. Tomorrow I can keep adding comments

juanitorduz:
<p>Fixed in https://github.com/bambinos/bambi/pull/440/commits/3c68d074c045ab63af7ded6bf3e73958ea0b630b</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>Fixed in https://github.com/bambinos/bambi/pull/440/commits/3c68d074c045ab63af7ded6bf3e73958ea0b630b</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>Fixed in https://github.com/bambinos/bambi/pull/440/commits/3c68d074c045ab63af7ded6bf3e73958ea0b630b</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>Fixed in https://github.com/bambinos/bambi/pull/440/commits/3c68d074c045ab63af7ded6bf3e73958ea0b630b</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>Fixed in https://github.com/bambinos/bambi/pull/440/commits/3c68d074c045ab63af7ded6bf3e73958ea0b630b</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>Ups! I think I have lost another comment about the model specification. Here is is for future reference:</p><blockquote>Here you mention that you are removing the intercept and then using one coefficient per floor level. But when you describe the mathematical model with you called the coefficient beta the "Intercept for the Floor level&nbsp;j". The pattern repeats for the other models, so it is better to clarify the nomenclature</blockquote>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

tomicapretto:
<p>I agree that saying we don't have an Intercept and then saying "intercept for the floor" may be confusing (even though we can consider those coefficients to be intercepts). Let's use "coefficient for the Floor level j". I'm fixing that now.</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

tomicapretto:
@juanitorduz it's sad you caught it too but I'm glad you're feeling better :D 

I added several changes, with the aim to simplify the models, but still be correct in what we're doing.


I compared our Bambi models, with the PyMC models and with the models in the original example in the book by Gelman and Hill. There are **so many** ways of writing the same underlying model, and **so many** terms that mean the same or something very similar, it's been very painful.

On top of that, I realized this particular example brings a lot of confusion. I'll try to explain in the following points:

**Intercepts and slopes**

Both the PyMC example and the original example in the book talk about the intercept and the floor slope. I think they do this because `floor` is a numerical 0-1 variable. 

For example, Gelman and Hill use `y ~ 1 + floor` and say that `1` is the intercept and `floor` is the slope. This forces Bambi (and what they do in R too) to drop one level in the `floor` variable to prevent linear dependencies in the design matrix and then you are left with one coefficient for `floor` (instead of 2, the number of levels).

What I find confusing is that they talk about the "intercept" and the "slope" as if they were unrelated things. Here, `1` is not like any other intercept in a regression model. In this case, it has a particular meaning. It represents the mean for the reference floor level, and `floor` represents a shift from that reference level. So if the model has `y ~ 1 + floor`, `floor`  is a difference between intercepts instead of a slope. 

Consequently, I think it is better to omit the global intercept when we use `floor` in the model so all its levels are preserved. In other words, I mean to use `y ~ 0 + floor`. In this case, since there's no intercept, Bambi (and any other tool that works with model formulas) won't drop the reference level in `floor` and it will use 2 coefficients. One coefficient represents the mean for `"basement"` and the other represents the mean for `"floor"`.

**County-level predictors**

When the uranium level is added, both examples say that a county-level predictor is added. **However** this does not mean that there's a county-specific slope for uranium (i.e. a hierarchical effect). What they mean is that there is a unique uranium measurement for each county, but the variable enters as a regular slope in the model.

**Trying to mimic PyMC parametrizations**

I think we should not try to replicate the same parameterizations that are used in the PyMC example. Two main reasons come to my mind:

* It uses the terms "intercepts" and "slopes" when I think it's not a good idea for what I said above. They are the coefficients for the two levels of the same variable (`floor`), not unrelated things.
* What's natural to write in PyMC may not be natural in Bambi and forcing a similar parametrization adds more confusion. In particular, it's natural to write hierarchical models in PyMC in a hierarchical manner (with hyperpriors). On the other hand, Bambi uses the mixed model notation where the group-specific component is a deviation from a common component. The sum of these two components is the hierarchical coefficient in the PyMC model.

---

I'll try to have a second read to this notebook soon. I think I've left many "intercepts" written there that I could express in a different way.

juanitorduz:
Hey @tomicapretto , thank you very much for your input! Here are some comments:
- First, I agree with you that these equivalent parametrisation and notations are a bit confusing for the user. In my experience, having a bit experience in pre PyMC and not much in the formula like specification, it was not straight forward. Still, I believe your comments above are very valuable and I would actually think it could be nice to add in this notebooks (maybe as a final remark?)...however I do not have a strong opinion about this.
- I agree that using `y ~ 0 + floor` to get parameters for each `floor` level is the way to go. Specially thinking about the priors: If we would drop one of the `floor` levels, then the priors for the intercept and for the difference are not that clear to interpret (this is also the recommendation from Statistical Rethinking).
 - Again, I think because all of this discussions, this example is already paying off and I am sure is going to help Bambi users. 

Is there anything I could support you with? Or maybe I wait for your second read and feedback from others?

tomicapretto:
According to what we've discussed yesterday, these are the points we need to complete to consider this finished

* [ ] Remove "i" index from models and coefficients.
* [ ] Remove redundant forest plots.
* [ ] Add a sentence explaining why we use log(radon) and not just radon.
* [ ] Check the correctness of the "intercept" word everywhere it is used.
* [ ] Have an example where we build the model, plot the graph, and then fit. This is to show that fitting is not necessary to see the model graph.

tomicapretto:
I think this is done if you want to have a last review @juanitorduz @aloctavodia

juanitorduz:
<p>fixed</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>fixed</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>fixed</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/440/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
Great work @tomicapretto ! The change in the notation to make it light-weight really improves readability IMO. Also, the explanation on how to build the model is quite handy :) 

In https://github.com/bambinos/bambi/pull/440/commits/497b0f1f99fa6d6f2edc44f5eed739ca7412578c and https://github.com/bambinos/bambi/pull/440/commits/a6b80d91aa2259bab6371456ef29fe617c4523b7 i fixed some small typos and harmonised some formulas (lower/upper case). From my side is ready to go! We can always get feedback and keep improving this valuable notebook 🚀

tomicapretto (approved):

PR #452, opened 2022-02-14
Update docs

tomicapretto:
This PR contains a couple of small updates.

* Fix GitHub handles on a couple of notebooks.
* Add beta regression notebook to the webpage.
* Fix the model in the hierarchical binomial regression model. The previous model was badly specified.


bambi/backend/terms.py
```diff
@@ -35,18 +35,25 @@ def build(self, spec):
         response_dims = []
         if spec.response.categorical and not spec.response.binary:
             response_dims = list(spec.response.pymc_coords)
+            response_dims_n = len(spec.response.pymc_coords[response_dims[0]])
+
+            # Arguments may be of shape (a,) but we need them to be of shape (a, b)
```
canyon289, line 40: Nice docstring

bambi/tests/test_built_models.py
```diff
@@ -71,6 +71,15 @@ def inhaler():
     return data
 
 
+@pytest.fixture(scope="module")
+def categorical_family_categorical_predictor():
+    from os.path import dirname, join
```
canyon289, line 76: Why import these here and not in global namespace?

canyon289 (approved):
i skimmed this but overall it looks good

bambi/tests/test_built_models.py
tomicapretto: There's no real reason. They just have been there forever. But this is a good time to move them to the global namespace.

aloctavodia (approved):

Issue #400, opened 2021-08-24
Coarse-grained multiprocessing with bambi


krassowski:
I tried to use Python's built-in multiprocessing capabilities as I have a series of models to fit (thousands of biomolecules), but while I can pass all my arguments and receive results such as `model.fit()` output or `arviz.summary()`, the model itself is not serializable if it contains a formulae. This comes down the reference to current global namespace which of course cannot be pickled (or, it would be a very bad thing if it was pickled as it would lead to OOM errors quickly). Here is a reproducer:

```python
from bambi import Model
from pandas import DataFrame
import numpy as np

x = np.linspace(0, 1, 200)

df = DataFrame(dict(
    x=x,
    y=1 + 2 * x + np.random.normal(scale=0.5, size=size)
))

m = Model(formula = 'y ~ x', data=df)

from pickle import dumps
dumps(m)
```

Resulting in:

> `TypeError: cannot pickle 'module' object`

We can narrow this down to private attribute `_design`:

```python
for k, v in vars(m).items():
    try:
        dumps(v)
    except:
        print(f'{k} is the culprit!')
        break
```

> _design is the culprit!

Where `type(m._design) is formulae.matrices.DesignMatrices`.

We can narrow it down further to `eval_enn._namespaces` (using the same method as above). As a workaround, it is sufficient to remove the namespaces reference with:

```python
m._design.eval_env._namespaces = None
```

It does not seem that this information is super important for use cases like this. Would it be a good idea to improve [formulae](https://github.com/bambinos/formulae) to be pickle-able? Could the `namespaces` attribute be silently removed on pickle time and a warning emitted after un-pickling if user tries to access it? Or maybe `bambi.Model` should implement custom pickling handler in addition to the discussion in #259?

For search engine indexing, the error that you are likely to get when running within a Jupyter environment is:

```python
PicklingError("Can't pickle <function <lambda> at 0xsomeaddress>: attribute lookup <lambda> on jupyter_client.session failed",
```

krassowski:
Or maybe the `_namespaces` should get filtered to only the relevant members - surely formula does not need to have access to full global namespace?

krassowski:
The workaround stated above is not sufficient if model contains priors, in which case there are other attributes that break pickling the `bambi.Model`:

```python
  File "python3.9/multiprocessing/connection.py", line 211, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
[...]

PR #453, opened 2022-02-17
add VonMises family

aloctavodia:
This allows to perform regression on circular response variables.

### DIFF
--- a/bambi/backend/pymc.py
+++ b/bambi/backend/pymc.py
@@ -8,7 +8,7 @@ import theano.tensor as tt
 
 from bambi import version
 
-from bambi.backend.links import cloglog, identity, inverse_squared, logit, probit
+from bambi.backend.links import cloglog, identity, inverse_squared, logit, probit, arctan_2
 from bambi.backend.terms import CommonTerm, GroupSpecificTerm, InterceptTerm, ResponseTerm
 
 _log = logging.getLogger("bambi")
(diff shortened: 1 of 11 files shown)


bambi/backend/pymc.py
```diff
@@ -25,6 +25,7 @@ class PyMC3Model:
         "log": tt.exp,
         "logit": logit,
         "probit": probit,
+        "tan_2": lambda x: 2 * tt.arctan(x),
```
tomicapretto, line 28: Can we make this a regular function? We're still not using any pickling but this can be an impediment. See #400.

bambi/families/likelihood.py
```diff
@@ -11,6 +11,7 @@
     "NegativeBinomial": {"params": ("mu", "alpha"), "parent": "mu", "args": ("alpha",)},
     "Poisson": {"params": ("mu",), "parent": "mu", "args": None},
     "StudentT": {"params": ("mu", "sigma"), "args": ("sigma", "nu")},
+    "Vonmises": {"params": ("mu", "kappa"), "parent": "mu", "args": ("kappa",)},
```
tomicapretto, line 14: This should be `"VonMises"` with capital M, as in the PyMC distribution name.

bambi/backend/terms.py
```diff
@@ -241,6 +241,7 @@ def build(self, nu, invlinks):
             nu = tt.concatenate([np.zeros((data.shape[0], 1)), nu], axis=1)
 
         # Add mean parameter and observed data
+
```
tomicapretto, line 244: was this added on purpose? Otherwise, I would remove it, so the format is the same in all the code.

tomicapretto (changes requested):
Only minor changes. 

It would be good if we add a very simple example in this PR for the record too.

tomicapretto (approved):
Looks good!! Merge if you don't plan to add anything else.

PR #457, opened 2022-02-21
Add example with categorical family

tjburch:
Added a notebook with examples using the categorical family to address #436. I noticed on the original merge (#426) that @tomicapretto made several very nice examples, so I aggregated several into one notebook and added some comments to tie it together. 

Note that I omitted two of the examples, the satisfaction survey and the inhaler example. I made that choice because I think those two fall more in the domain of ordinal regression, rather than categorical. I'm happy to add them if you'd prefer to include them though.

PR #460, opened 2022-02-23
Return integers posterior predictive

aloctavodia:
follow up of  #458

This is related to the failing of `az.plot_ppc` reported in #457

### DIFF
--- a/bambi/families/multivariate.py
+++ b/bambi/families/multivariate.py
@@ -77,7 +77,7 @@ class Categorical(MultivariateFamily):
         obs_n = mean.shape[-1]
 
         pps = np.empty((draws_n, obs_n), dtype=int)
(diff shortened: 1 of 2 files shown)


tomicapretto:
While using integers allows us to use `az.plot_ppc()` with the `InferenceData` object returned by Bambi, the resulting plot won't contain the labels of the responses. This is because ArviZ only knows about float and integer values (correct me if I'm missing some detail @aloctavodia).

In the meantime, one could create the plot with ArviZ and then add the labels manually, this would be like

```python
model.predict(idata, kind="pps")

ax = az.plot_ppc(idata)
labels = [str(level) for level in model.response.levels]
ax.set_xticks(np.arange(len(labels)))
ax.set_xticklabels(labels)
```

PR #463, opened 2022-03-07
Require Python version at least as recent as 3.7.2 to avoid ``NoReturn`` type bug

yannmclatchie:
Hey!

I recently tried to install Bambi in a virtual environment using ``python==3.7.1``, but upon importing was faced with the following error message.

```python
In [1]: import bambi as bmb
---------------------------------------------------------------------------
/anaconda3/lib/python3.7/typing.py in _type_check(arg, msg, is_argument)
    133     if (isinstance(arg, _SpecialForm) and arg is not Any or
    134             arg in (Generic, _Protocol)):
--> 135         raise TypeError(f"Plain {arg} is not valid as type argument")
    136     if isinstance(arg, (type, TypeVar, ForwardRef)):
    137         return arg

TypeError: Plain typing.NoReturn is not valid as type argument
```
I found similar bugs in [spaCy](https://github.com/explosion/spaCy/issues/9600) and [py-ipfs-http-client](https://github.com/ipfs-shipyard/py-ipfs-http-client/issues/235), and in both cases maintainers referenced [Python Issue 34921](https://bugs.python.org/issue34921) relating to a bug in some Python versions where ``NoReturn`` types were being flagged as invalid. This bug seems to have been fixed in [this commit](https://github.com/python/cpython/commit/f71a5922916abd6cc7bf7d99ed4715b6e96e5981), and does not occur in Python versions ``>=3.7.2`` as far as I can tell. Other forum exchanges seem to indicate the same conclusion.

This pull request then has the simple modification of including the requirement ``python>=3.7.2`` to the ``setup.py`` file. Please let me know if you feel this is an appropriate and sufficient modification to the code base to prevent future users from diving down the Stack Overflow rabbit hole!

Cheers


setup.py
```diff
@@ -8,6 +8,24 @@
 README_FILE = os.path.join(PROJECT_ROOT, "README.md")
 VERSION_FILE = os.path.join(PROJECT_ROOT, "bambi", "version.py")
 REQUIREMENTS_FILE = os.path.join(PROJECT_ROOT, "requirements.txt")
+REQUIRED_PYTHON_VERSION = (3, 7, 2)
+
+
+def check_installation(rv):
+    current_version = sys.version_info
+    if (
```
aloctavodia, line 16: you can evaluate `if sys.version_info < (3, 7, 2)`

setup.py
```diff
@@ -8,6 +8,24 @@
...
+        and (current_version[2] < rv[2])
+    ):
+        sys.stderr.write(
+            f"[{sys.argv[0]}] - Error: Your Python interpreter must be "
+            + f"{rv[0], rv[1], rv[0]}or greater\n"
+        )
+        sys.exit(-1)
+    else:
+        pass
+    return 0
```
aloctavodia, line 28: is this a recommended practice? Otherwise we can remove it

yannmclatchie: I don't think it's strictly necessary and can probably be removed

tomicapretto:
I confirm I can reproduce the problem in an environment with Python 3.7.1. However the problem isn't about Bambi, it is about Theano. 

I was a little skeptical about typing errors in Bambi because we use no typing at all. Now I confirmed this is a typing error in Theano.

@yannmclatchie could you include the full traceback in the description of the problem? I think that is going to be clearer about the root of the problem. 


You can use the following template so it is collapsible

  ```
  <details><summary>Complete error traceback</summary>

    ```python
    [The complete error output here]
    ```

  </details>
  ```

setup.py
```diff
@@ -8,6 +8,16 @@
 README_FILE = os.path.join(PROJECT_ROOT, "README.md")
 VERSION_FILE = os.path.join(PROJECT_ROOT, "bambi", "version.py")
 REQUIREMENTS_FILE = os.path.join(PROJECT_ROOT, "requirements.txt")
+REQUIRED_PYTHON_VERSION = (3, 7, 2)
```
tomicapretto, line 11: Being nitpick here, but I think that `MINIMUM_PYTHON_VERSION` is better than `REQUIRED_PYTHON_VERSION`. The latter can be interpreted as the only version where Bambi works. The former means what we need it to be, the minimum version of Python where Bambi works.

yannmclatchie: Sure, I see how this could be confusing

tomicapretto (approved):
LGTM! I think this can be merged @aloctavodia

PR #467, opened 2022-03-10
[WIP] Upgrade formulae to add more contrasts options

MartinGrignard:
This pull request aims at upgrading the required version of `formulae` to allow more contrasts to be used in the formula of the models.

Since the development version of `formulae` changed a lot of internals (*e.g.* replacing the `term_info` dictionary by a `Term` object), it introduces a lot of bugs all across the package. These must first be addressed before focusing on the contrasts.

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -433,7 +433,7 @@ class Model:
         # Check if any of the names of the auxiliary params match the names of terms in the model
         # If that happens, raise an error.
         if aux_params and self._design.common:
-            conflicts = [name for name in aux_params if name in self._design.common.terms_info]
+            conflicts = [name for name in aux_params if name in self._design.common.terms]
             if conflicts:
                 raise ValueError(
                     f"The prior name for {', '.join(conflicts)} conflicts with the name of a "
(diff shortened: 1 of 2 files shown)


MartinGrignard:
@tomicapretto, I'm facing a first problem. In the development version of `formulae`, there is no `encoding` attribute to the `Term` class. Still, I need a way to determine if the term encoding is `"full"` or not. I suppose that it has to do with the new `data` attribute but I can't seem to find where such information is added.

Do you have any hint on how to actually do the following test?
```python
if self.kind == "interaction":
                self.pymc_coords[name] = term.levels
            elif term.encoding == "full":  # This line cannot run since term has no encoding attribute
                self.pymc_coords[name] = term.levels
            else:
                self.pymc_coords[name] = term.levels[1:]
```

tomicapretto:
@MartinGrignard 

The terms have a `.spans_intercept` attribute now. That being `True` is equivalent to the old `"full"`.

Since you can have arbitrary encodings now, we don't use `levels[1:]` anymore to get the actual levels of the term. You can take the new values of the levels (without having to do further action) from the `.levels` attribute.

I think having a look to the tests https://github.com/bambinos/formulae/blob/master/formulae/tests/test_design_matrices.py is going to be useful.

tomicapretto:
@MartinGrignard this is done in #488. Have a look at that if you want to use different contrasts now :)

PR #468, opened 2022-03-11
Clean setup.py

tomicapretto:
* Remove unused dependencies: `nose`
* Change maintainer.
* Cleans other setup functions a little.

@aloctavodia I added myself as maintainer, but I consider "The Developers of Bambi" is another suitable option. In that case, I think I still need to keep my email because we don't have a Bambi email and I guess we don't want to.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,3 @@
-import codecs
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


aloctavodia:
One more change could be to remove `codecs` and use  `open` with the `encoding` argument.

PR #470, opened 2022-03-14
add  include_group_specific argument to predict

alexjonesphd:
Add new argument ~`with_group_specific`~  `include_group_specific` to `model.predict()` that allows prediction without group-specific effects.

Currently when calling `model.predict()` with a hierarchical model on new data, the `data` argument must contain variables with valid group-specific labels from the original data. There is occasional need to make predictions without including group-specific effects, i.e. set them to zero. The proposed extra argument allows bypassing of the group-specific predictions to do this.

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -754,7 +754,9 @@ class Model:
         return idata
 
     # pylint: disable=protected-access
-    def predict(self, idata, kind="mean", data=None, draws=None, inplace=True, with_group_specific=True):
+    def predict(
+        self, idata, kind="mean", data=None, draws=None, inplace=True, include_group_specific=True
+    ):
         """Predict method for Bambi models
 
         Obtains in-sample and out-of-sample predictions from a fitted Bambi model.
@@ -772,7 +774,7 @@ class Model:
         data: pandas.DataFrame or None
             An optional data frame with values for the predictors that are used to obtain
             out-of-sample predictions. If omitted, the original dataset is used.
-        with_group_specific: bool
+        include_group_specific: bool
             If ``True`` make predictions including the group specific effects. Otherwise,
             predictions are made with common effects only (i.e. group specific are set
             to zero).
@@ -815,7 +817,7 @@ class Model:
             else:
                 X = self._design.common._evaluate_new_data(data).design_matrix
 
-        if self._design.group and with_group_specific:
+        if self._design.group and include_group_specific:
             if in_sample:
                 Z = self._design.group.design_matrix
             else:
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -754,7 +754,7 @@ class Model:
         return idata
 
     # pylint: disable=protected-access
-    def predict(self, idata, kind="mean", data=None, with_group_specific=True, draws=None, inplace=True):
+    def predict(self, idata, kind="mean", data=None, draws=None, inplace=True, with_group_specific=True):
         """Predict method for Bambi models
 
         Obtains in-sample and out-of-sample predictions from a fitted Bambi model.
@@ -815,7 +815,7 @@ class Model:
             else:
                 X = self._design.common._evaluate_new_data(data).design_matrix
 
-        if self._design.group anf with_group_specific:
+        if self._design.group and with_group_specific:
             if in_sample:
                 Z = self._design.group.design_matrix
             else:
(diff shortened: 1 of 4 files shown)


bambi/models.py
```diff
@@ -811,7 +815,7 @@ def predict(self, idata, kind="mean", data=None, draws=None, inplace=True):
             else:
                 X = self._design.common._evaluate_new_data(data).design_matrix
 
-        if self._design.group:
+        if self._design.group anf with_group_specific:
```
tomicapretto, line 818: There's a typo here, `anf` instead of `and`

bambi/models.py
```diff
@@ -754,7 +754,7 @@ def prior_predictive(self, draws=500, var_names=None, omit_offsets=True, random_
         return idata
 
     # pylint: disable=protected-access
-    def predict(self, idata, kind="mean", data=None, draws=None, inplace=True):
+    def predict(self, idata, kind="mean", data=None, with_group_specific=True, draws=None, inplace=True):
```
tomicapretto, line 757: Why you added it between `data` and `draws`? To me the most natural is to add it as the last argument, just in case someone has code that matches arguments positionally.

Is there any other reason?

tomicapretto (commented):
A couple of minor comments:

* I'm still not sure about the name of the argument.
* Could you add a test? Are you familiar with testing? If not, let me know and I can help.

bambi/models.py
alexjonesphd: OK, shifted it to the end. It felt natural to go there after specifying `kind` and `data` but agreed this could break compatibility if using positional args.

bambi/models.py
aloctavodia: What about `include_group_specific`? I agree we need a test for this PR. 
Additionally, we should open an issue to remember adding an example of this feature.

tomicapretto: @alexjonesphd a test is one function (or maybe it can be a couple of functions) where you evaluate the functionality you're adding. In this case, it would be the prediction with `include_group_specific` set to False.

It would be a little harder to check the correctness of the result (i.e. not only the code does work, but the result does not contain any group-specific effect). In the meantime, I think making two predictions, one with `include_group_specific=True` and other with `include_group_specific=False` and comparing they are not the same is a good first step.

For examples about tests, see here: https://github.com/bambinos/bambi/blob/f47bda0c1eb3cc430ddbd95209c06c856751c482/bambi/tests/test_built_models.py#L455-L478

If you need help, or you simply can't do it, let us now and we will do it.

tomicapretto: Yes. You can push as another commit here. Once it's added we'll enable the tests to run in th PR

tomicapretto:
@alexjonesphd I tried to push changes to this PR but I couldn't. Maybe you need to allow that on your side? 

Otherwise we can create a new PR on our side and push these changes.

alexjonesphd:
@tomicapretto do you still need a test for this? I was about to start working on it at the end of a busy semester, sorry for the delay. 

I'm not sure why you can't push changes, as far as I can tell everything seems in order my side, but I'm still new! I've got 'allow edits by maintainers' checked. It may be because I'm a first time contributor? If its easier to create a new PR your side that's fine, I can close this one. I've probably done something wrong along the way.

tomicapretto:
> @tomicapretto do you still need a test for this? I was about to start working on it at the end of a busy semester, sorry for the delay.
> 
> I'm not sure why you can't push changes, as far as I can tell everything seems in order my side, but I'm still new! I've got 'allow edits by maintainers' checked. It may be because I'm a first time contributor? If its easier to create a new PR your side that's fine, I can close this one. I've probably done something wrong along the way.

Looks like I was doing something wrong on my side. I tried something different now and it worked. Let's see how this goes!

tomicapretto:
Updated to Black 22.3.0. See https://github.com/psf/black/issues/2964

tomicapretto:
@aloctavodia this is ready to be merged I think. Maybe a second pair of eyes on the test is helpful.

alexjonesphd:
Thank you so much @tomicapretto, let me know if there's anything I can do.

aloctavodia (approved):
LGTM. Thanks @alexjonesphd. One last thing, could you please update the  `Changelog.md` file?

tomicapretto:
> LGTM. Thanks @alexjonesphd. One last thing, could you please update the `Changelog.md` file?

Done! Thanks for the reminder :)

PR #472, opened 2022-03-16
Clean internals

tomicapretto:
This PR does not add any new feature, but it cleans many internals, improves documentation, and also fixes a couple of things.

**Summary of changes**

* Fixed a bug with the inverse link family. The backend side was good, but it was wrong on Bambi's side. It was using cloglog. 
* I decided to use `namedtuple`s in a couple of places. I think it makes the code more robust (it also saves some lines but that's not the most important thing in my opinion).
* Removed a helper function named `_extract_family_prior()`. It's simply not used.
* Removed `bambi/families/utils.py`. Again, not used.
* The function `check_full_rank()` in `utils.py` now checks the array is 2 dimensional. We may not be passing arrays of other dimension, but this check makes it more robust.
* Documented the module `bambi/defaults`.
* Changed smallish things to make all docstrings more consistent.


bambi/defaults/defaults.py
```diff
@@ -157,6 +156,28 @@
...
+
+    The parameter ``kwargs`` is used to pass hyperpriors that are assigned to the parameters of
+    the prior to be built.
+
+    Parameters
+    ----------
+    dist: str, int, float
+        If a string, it is the name of the prior distribution. Default values for its parameters
+        are taken from ``SETTINGS_DISTRIBUTIONS``. If a number, a factor that is used to scale
+        the standard deviation of the priors that Bambi generates automatically.
```
aloctavodia, line 169:
```suggestion
        If a string, it is the name of the prior distribution with default values taken from ``SETTINGS_DISTRIBUTIONS``. If a number, it is a factor used to scale
        the standard deviation of the priors generated automatically by Bambi.
```

bambi/models.py
```diff
@@ -294,8 +294,7 @@ def fit(
     def build(self):
         """Set up the model for sampling/fitting.
 
-        Performs any steps that require access to all model terms (e.g., scaling priors on each
-        term), then calls the backend's ``.build()`` method.
+        Creates an instance of the underlying PyM3 model and adds all the necessary terms to it.
```
aloctavodia, line 297:
```suggestion
        Creates an instance of the underlying PyMC model and adds all the necessary terms to it.
```

aloctavodia (approved):

PR #475, opened 2022-03-22
Fix bug in custom link definition

aloctavodia:
While trying to reproduce the custom link example in the getting started notebook I got a `segmentation fault` error. This fix it and add a test.


tomicapretto (approved):
Looks good! Thanks for the catch!

Issue #477, opened 2022-03-23
`Model.predict()` generates unexpected out-of-sample predictions for a mixed effects model


LeonieMei:
Hello,

first of all, thanks for the great work on this project, I've been using bambi a lot and it has been super helpful!

I'm currently facing a (potential) issue when trying to make out-of-sample predictions for a logistic regression model built with the following formula:

`y ~ x1 + x2 + x3 + (0 + x2|x1) + (0 + x3|x1) `

where `x1` and `x2` are categorical variables with two dimensions respectively and `x3` is a continuous variable.

The out-of-sample data I'm trying to make predictions for looks like this (exemplary):

| x1 | x2 | x3  |
|----|----|-----|
| 0  | 0  | 0   |
| 0  | 0  | 0.5 |
| 0  | 0  | 1   |
| 0  | 0  | 1.5 |
| 1  | 0  | 0   |
| 1  | 0  | 0.5 |
| 1  | 0  | 1   |
| 1  | 0  | 1.5 |

There was no error when running `model.predict(iData, data=out_of_sample_data, kind='mean')`, however the spaghetti plot I generated from the posterior predictions looked off for when `x1==1`, the variance was much bigger than I expected. (I noticed this because I had manually made a plot displaying the 0.5 decision boundary for `x3`, i.e. the mean value and hdi intervals of `x3` where there is a 50% probability of a positive outcome and that didn't match what I saw in the spaghetti plot.)

I then had a look at the code and noticed that the `Z` matrix generated in the `predict` method in `models.py` looked different from what I expected.
Here's the code bit I'm referring to (last line):
```python
        if self._design.group:
            if in_sample:
                Z = self._design.group.design_matrix
            else:
                Z = self._design.group._evaluate_new_data(data).design_matrix
```

What I got for `Z` was the following:

|||||||
|---|---|---|---|-----|-----|
| 1 | 0 | 0 | 0 | 0   | 0   |
| 1 | 0 | 0 | 0 | 0.5 | 0   |
| 1 | 0 | 0 | 0 | 1   | 0   |
| 1 | 0 | 0 | 0 | 1.5 | 0   |
| 0 | 0 | 1 | 0 | 0   | 0   |
| 0 | 0 | 1 | 0 | 0   | 0.5 |
| 0 | 0 | 1 | 0 | 0   | 1   |
| 0 | 0 | 1 | 0 | 0   | 1.5 |

...but what I was expecting (after trying to make sense of it) was this:

|||||||
[...]

PR #478, opened 2022-04-02
Change how reshape is performed for categorical group specific terms.

tomicapretto:
In #477 we discuss a problem associated with how we reshape draws from the posterior of categorical group-specific effects. These are group-specific effects where the expression part (LHS of `|`) is a categorical variable.

This PR fixes that issue by performing a two-step reshape for the mentioned type of terms. The second step uses `order="F"` so the reshape varies first through the `expr` part of the term and then through the `factor` part.

In NumPy's terms, the `expr` values vary along the first index and the `factor` values vary along the second index. With `order="F"` we have that the first index (`expr`) changes fastest and the last index (`factor`) changes slowest.

**Note** This does not affect predictions for models not involving categorical group-specific terms, but I still need to double-check if this is the right approach.

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -869,17 +869,23 @@ class Model:
                 # 1-dimensional expr (the great majority: a single slope or intercept)
                 if len(shape) == 3:
                     shape = (shape[0] * shape[1], shape[2])
+                    values = values.reshape(shape)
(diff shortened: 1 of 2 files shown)


tomicapretto:
Updated to Black 22.3.0. See https://github.com/psf/black/issues/2964

aloctavodia (approved):

PR #480, opened 2022-04-05
fix click version

aloctavodia:
There is a problem with the last version of click, that generates issues with black see CI errors in #479


tomicapretto:
@aloctavodia can we just update black? It is what I did in #478 for example.

PR #493, opened 2022-04-25
Extract args

tomicapretto:
This PR adds two things

* The function `c()` which stands for `concatenate` and is used to concatenate variables in multinomial regression.
* A utility function, `extract_argument_names`, which extracts the names passed to a function call, in this case `c()`. This allows us to use these names as labels.

For example

```python
import arviz as az
import bambi as bmb
import pandas as pd

data = pd.DataFrame(
    {
        "state": ["A", "B", "C"],
        "y1": [35298, 1885, 5775],
        "y2": [167328, 20731, 21564],
        "y3": [212682, 37716, 20222],
        "y4": [37966, 5196, 3277],
    }
)

# Contains group-specific effect
model = bmb.Model("c(y1, y2, y3, y4) ~ 0 + state", data, family="multinomial", noncentered=False)
idata = model.fit(random_seed=0)

az.summary(idata)
```

image

Thanks to that we have the `y2`, `y3`, etc... in the InferenceData object. What do you think @aloctavodia ?

~**TODO**~

* ~What happens when we compute predictions? Somehow we need access to `y1` too. I need to think about  better ways to store this in the response term.~ Done!


bambi/families/multivariate.py
```diff
@@ -6,40 +6,25 @@
...
-        """Predict mean response"""
-        mean = self.link.linkinv(linear_predictor)
-        obs_n = mean.shape[-1]
-        name = model.response.name + "_mean"
-        coord_name = name + "_dim_0"
-
-        # Drop var/dim if already present
-        if name in posterior.data_vars:
-            posterior = posterior.drop_vars(name)
+        return NotImplemented
```
canyon289, line 9: If your removing logic I would suggest a deprecation warning, or a more informative why notimplemented exception

bambi/models.py
```diff
@@ -811,93 +811,139 @@ def predict(
...
+                    )
+                elif coords == {"chain", "draw"}.union(coord_expr + coord_factor):
+                    transpose_coords = ["samples", "coefs"]
+                    values = (
+                        term_posterior.stack(samples=("chain", "draw"))
+                        .stack(coefs=tuple(coord_factor + coord_expr))
+                        .transpose(*transpose_coords)
+                        .values
+                    )
+                elif isinstance(self.family, (multivariate.Categorical, multivariate.Multinomial)):
```
canyon289, line 903: Comments on each of these blocks would help folks understand why each exists

bambi/utils.py
```diff
@@ -19,3 +24,55 @@ def multilinify(sequence, sep=","):
     """Make a multi-line string out of a sequence of strings."""
     sep += "\n"
     return "\n" + sep.join(sequence)
+
+
+def c(*args):  # pylint: disable=invalid-name
```
canyon289, line 29: I would argue for a longer more descriptive name

bambi/families/multivariate.py
tomicapretto: Actually, this was never used. There was only one child of `MultivariateFamily` (`Categorical`) and it always implemented its own `predict()` method so users will never see a problem with this `NotImplemented` stuff.

I do agree that maybe I can handle it in a better way. I haven't thought very well of other multivariate families. The only extra thing I have in mind is 'MultivariateNormal'. 

On top of that, with the addition of `MutableData` to PyMC, it does make a lot of sense to rely on PyMC for predictions instead of computing them ourselves. For example, this would simplify things a lot when Bambi includes what's known as "distributional models" (e.g. see https://cran.r-project.org/web/packages/brms/vignettes/brms_distreg.html)

bambi/utils.py
tomicapretto: I know that `c()` may not be the most self-describing name, but it's a convention I borrowed from the R language where `c()` is used a lot. On the other side, I'm 100% open to considering other alternatives and seeing if they work.

tomicapretto: @aloctavodia do you have any opinion about it?

PR #496, opened 2022-04-30
[WIP] update backend to PyMC v4

aloctavodia:


canyon289:
Maybe we should disable the linters for now so we can see if tests pass. Im happy to help out as well

aloctavodia:
Many test are still not passing

canyon289:
Thats fine, if we skip linting we can see which tests fail

tomicapretto:
I'm running a couple of models and getting this

```
/home/tomas/anaconda3/envs/bambiv4/lib/python3.9/site-packages/pymc/aesaraf.py:1005: UserWarning: The parameter 'updates' of aesara.function() expects an OrderedDict, got <class 'dict'>. Using a standard dictionary here results in non-deterministic behavior. You should use an OrderedDict if you are using Python 2.7 (collections.OrderedDict for older python), or use a list of (shared, update) pairs. Do not just convert your dictionary to this type before the call as the conversion will still be non-deterministic.
```

Are you familiar with it?

canyon289:
for the dictionary one yes but newer python dicts are ordered by default I believe. we may be able to change that in aesara. I can check later

bambi/tests/test_built_models.py
```diff
@@ -377,6 +376,7 @@ def dicts_close(a, b):
     assert all([dicts_close(priors0[x], priors1[x]) for x in priors0.keys()])
 
 
+@pytest.mark.skip(reason="Standard deviation results are different and I don't know why")
```
tomicapretto, line 379: I'm adding this line to skip this test. There's a difference in the standard deviation. I tried to fix it on my side but I'm not familiar with how this works. I guess that PyMC changed something in its `find_MAP` or `find_hessian` and that's affecting this comparison.

tomicapretto:
Some of our tests are failing because `.squeeze()` does not seem to work as before on tensor variables.

See the following example

```python
import aesara
import pymc as pm

coords = {
    "A": list("abc"), 
    "B": list("x")
}

with pm.Model(coords=coords) as pm_model:
    b = pm.Normal("b", dims=("A", "B"))
    c = b.squeeze()
    
print(b.shape.eval())
print(c.shape.eval()) # Expect this to be just 3

print("Aesara version", aesara.__version__)
print("PyMC version", pm.__version__)
```

```
[3 1]
[3 1]
Aesara version 2.5.1
PyMC version 4.0.0b6
```

I'm going to check with Ricardo if this is intended or unintended behavior.

aloctavodia:
The Op i1 error is also something that should be fixed on the aesara side.

tomicapretto:
Ricardo responded this is because of PyMC, not Aesara. And it should be working when https://github.com/pymc-devs/pymc/issues/5751 is addressed. In the meantime, we could fix it on our side without relying on `.squeeze()` but I'm not sure if we want to do it.

aloctavodia:
I think it makes sense to wait.

aloctavodia:
Yes we merge and then we keep updating the v4 branch until is ready.

tomicapretto:
Merging now. This will be in `v4` branch.

PR #499, opened 2022-05-08
Use dims instead of coords

tomicapretto:
I realized we were using the term `coords` for things that are actually `dims` in the `xarray` world.

This PR updates several places where we were wrongly using `coords`.

On top of that, I renamed the dimensions we use for group-specific terms:

* `factor + "_coord_group_factor"` is now `factor + "__factor_dim"`
* `expr + "_coord_group_expr"` is now `expr + "__expr_dim"`

which are  clearer, shorter, and safer.


aloctavodia (approved):
LGTM

PR #514, opened 2022-06-01
Use PyMC v4 

aloctavodia:


tomicapretto:
Great to see all green here! Sorry I've not been able to check before. Is there anything I could check or do to help?

aloctavodia:
Don't worry!  I think this is ready to merge, nevertheless there is some weird "extra wigllines" in some of the trace plots (the KDEs), even when summaries and other plots looks fine.

aloctavodia:
We no longer need that branch. This PR includes the changes in that branch.

canyon289:
🎉 I vote merging this one

aloctavodia:
I can confirm that the "extra wiglinness" is not due to a chance introduce now and it was already there. So it is most likely related to kde in arviz than bambi.

tomicapretto:
Sorry for being too conservative, but I would vote to merge in main there's an official PyMC release (non-beta). We could also leverage their announcement message (which I think there will be a good one) to say that Bambi also supports v4 now.

I imagine people working with both Bambi and PyMC (which is PyMC3 at the moment). If we make our main branch dependent on PyMC (v4) it may break their installation and it can result in unexpected problems

aloctavodia:
If users work from main they should expect breaks. Only a release should be stable. We will make the next Bambi release once we have PyMC 4.0, but in the meantime we can merge this and even merge other PR, if any that should increase the chance of a stable bambi release.

tomicapretto:
Ok, let's merge then.

twiecki:
You should be able to pin pymc 4.0.0 now and cut a release.

Issue #515, opened 2022-06-02
Spline regression by group in bambi


edoardovivo:
Hello, I am trying to model some data using a spline regression by group (in this case, by day of week), and I get a 'wrong number of dimension' error. 

I am using bambi version 0.8.0.

Here is a reproducible example:
 ```
x_check = pd.DataFrame({'x': [  82.,  143.,  426.,  641., 1156.,  986.,  365.,  187.,  254.,
        550.,  101.,  661.,  327.,  119.],
                         'day': ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat']*2 ,
                        'y': [ 571.,  684., 1652., 2130., 2455., 1874., 1288., 1011., 1004.,
       1993.,  593., 1986., 1503.,  711.]
                        })
iknots = np.array([191. , 297. , 512.5])

model = bmb.Model("y ~ (bs(x, knots=iknots, intercept=False, degree=1)|day)", data=x_check)
model.build()
```

And here is the full traceback:

```
TypeError                                 Traceback (most recent call last)
<command-2966464070710328> in <module>
      1 model = bmb.Model("y ~ (bs(x, knots=iknots, intercept=False, degree=1)|day)", data=x_prueba)
----> 2 model.build()

/databricks/python/lib/python3.8/site-packages/bambi/models.py in build(self)
    282         """
    283         self.backend = PyMC3Model()
--> 284         self.backend.build(self)
    285         self.built = True
    286 

/databricks/python/lib/python3.8/site-packages/bambi/backend/pymc.py in build(self, spec)
     66             self._build_intercept(spec)
     67             self._build_common_terms(spec)
---> 68             self._build_group_specific_terms(spec)
     69             self._build_response(spec)
     70             self._build_potentials(spec)

/databricks/python/lib/python3.8/site-packages/bambi/backend/pymc.py in _build_group_specific_terms(self, spec)
    170 
    171             # Build
--> 172             coef, predictor = group_specific_term.build(spec)
    173 
    174             # Add to the linear predictor

/databricks/python/lib/python3.8/site-packages/bambi/backend/terms.py in build(self, spec)
    108         # Squeeze ensures we don't have a shape of (n, 1) when we mean (n, )
    109         # This happens with categorical predictors with two levels and intercept.
--> 110         coef = self.build_distribution(dist, label, dims=dims, **kwargs).squeeze()
    111         coef = coef[self.term.group_index]
    112 

[...]

PR #516, opened 2022-06-02
Group specific terms now work with numeric of multiple columns

tomicapretto:
See #515


aloctavodia:
This seems to be ready, except for a test (we can add it later if we open an issue)

aloctavodia:
This seems to be ready, except for a test (we can add it later if we open an issue)

PR #520, opened 2022-06-06
use HDI in plot_cap as the default

aloctavodia:


bambi/plots/plot_cap.py
```diff
@@ -151,15 +155,21 @@ def plot_cap(model, idata, covariates, hdi_prob=0.94, legend=True, ax=None):
     cap_data = create_cap_data(model, covariates)
     idata = model.predict(idata, data=cap_data, inplace=False)
 
+    if hdi_prob is None:
+        hdi_prob = az.rcParams["stats.hdi_prob"]
+
     if not 0 < hdi_prob < 1:
         raise ValueError(f"'level' must be greater than 0 and smaller than 1. It is {hdi_prob}.")
```
tomicapretto, line 162: 'level' here should be 'hdi_prob' now (my bad)!

aloctavodia: good catch!

Issue #523, opened 2022-06-07
Feature request: Robust regression


markgoodhead:
Hello,

It'd be great to add the Laplace family to Bambi to enable robust regression. If this is potentially a good task for a beginner to the bambi codebase I'd be happy to give it a try if given some pointers as to what steps are required to do this?

tomicapretto:
Happy to see you already opened a PR!

PR #490, opened 2022-04-19
Add multinomial family

tomicapretto:
With Bambi

```python
import arviz as az
import bambi as bmb
import numpy as np
import pandas as pd

az.style.use("arviz-darkgrid")
rng = np.random.default_rng(0)

def c(*args):
    return np.column_stack(args)

n = 15
size = 10
data = pd.DataFrame(
    {
        "y1": rng.binomial(n, 0.3, size),
        "y2": rng.binomial(n, 0.5, size),
        "y3": rng.binomial(n, 0.7, size),
        "x": rng.normal(size=size)
    }
)

bmb_model = bmb.Model("c(y1, y2, y3) ~ x", data, family="multinomial")
idata = bmb_model.fit(random_seed=0)

bmb_model.predict(idata)
az.plot_trace(idata);
```

image

With PyMC3

```python
import arviz as az
import bambi as bmb
import numpy as np
import pandas as pd
import pymc3 as pm
import theano.tensor as tt

from formulae import design_matrices

az.style.use("arviz-darkgrid")

rng = np.random.default_rng(0)

def c(*args):
    return np.column_stack(args)

n = 15
size = 10
data = pd.DataFrame(
    {
        "y1": rng.binomial(n, 0.3, size),
        "y2": rng.binomial(n, 0.5, size),
        "y3": rng.binomial(n, 0.7, size),
        "x": rng.normal(size=size)
    }
)

dm = design_matrices("c(y1, y2, y3) ~ x", data)
Y = dm.response.design_matrix
X = dm.common.design_matrix

coords = {"x_dim": ["b", "c"]}
with pm.Model(coords=coords) as model:
    
    b0 = pm.Normal("b0", dims="x_dim")
    b1 = pm.Normal("b1", dims="x_dim")
    
    # Prepend one dimension bc response multicategorical and predictor 1 dimensional
    b = tt.concatenate([b0[np.newaxis, :], b1[np.newaxis, :]])
    
    nu = pm.math.dot(X, b)
    
    nu = tt.concatenate([np.zeros((Y.shape[0], 1)), nu], axis=1)
    
    # Must add up to 1 in the last axis
    p = pm.Deterministic("p", tt.nnet.softmax(nu))
    
    response = pm.Multinomial("response", n=Y.sum(axis=1)[:, None], p=p, observed=Y)

with model:
    idata = pm.sample(return_inferencedata=True, random_seed=0)

az.plot_trace(idata);
```
image

PR #524, opened 2022-06-07
Initial attempt to add support for the Laplace family

markgoodhead:
This is to address https://github.com/bambinos/bambi/issues/523

Note this is a WIP as I attempt to learn more of the bambi internals.

I'm not very familiar with the codebase but I attempted to copy the format of the Normal family and modify it accordingly. One thing I couldn't quite see is how this interacts with pymc - there's no "pm.Normal" calls which I was expecting to modify to the Laplace equivalent. Please let me know if there's something I'm missing and I'll attempt to fix this up.

### DIFF
--- a/bambi/families/link.py
+++ b/bambi/families/link.py
@@ -41,12 +41,12 @@ def invcloglog(eta):
 def probit(mu):
     """Probit function that ensures the input is in (0, 1)"""
     mu = force_within_unit_interval(mu)
-    return 2 ** 0.5 * special.erfinv(2 * mu - 1)  # pylint: disable=no-member
+    return 2**0.5 * special.erfinv(2 * mu - 1)  # pylint: disable=no-member
 
 
 def invprobit(eta):
     """Inverse of the probit function that ensures result is in (0, 1)"""
-    result = 0.5 + 0.5 * special.erf(eta / 2 ** 0.5)  # pylint: disable=no-member
+    result = 0.5 + 0.5 * special.erf(eta / 2**0.5)  # pylint: disable=no-member
     return force_within_unit_interval(result)
 
 
@@ -70,7 +70,7 @@ def softmax(eta, axis=None):
 
 
 def inverse_squared(mu):
-    return 1 / mu ** 2
+    return 1 / mu**2
 
 
 def inv_inverse_squared(eta):
--- a/bambi/priors/scaler.py
+++ b/bambi/priors/scaler.py
@@ -32,7 +32,7 @@ class PriorScaler:
         if self.priors:
             sigmas = np.hstack([prior["sigma"] for prior in self.priors.values()])
             x_mean = np.hstack([self.model.terms[term].data.mean(axis=0) for term in self.priors])
-            sigma = (sigma ** 2 + np.dot(sigmas ** 2, x_mean ** 2)) ** 0.5
+            sigma = (sigma**2 + np.dot(sigmas**2, x_mean**2)) ** 0.5
 
         return mu, sigma
 
--- a/bambi/tests/test_built_models.py
+++ b/bambi/tests/test_built_models.py
@@ -131,7 +131,9 @@ def test_many_common_many_group_specific(crossed_data):
         dropna=True,
     )
     model0.fit(
-        tune=10, draws=10, chains=2,
+        tune=10,
+        draws=10,
+        chains=2,
     )
 
     model1 = Model(
@@ -140,7 +142,9 @@ def test_many_common_many_group_specific(crossed_data):
         dropna=True,
     )
     model1.fit(
-        tune=10, draws=10, chains=2,
+        tune=10,
+        draws=10,
+        chains=2,
     )
     # check that the group specific effects design matrices have the same shape
     X0 = pd.concat([pd.DataFrame(t.data) for t in model0.group_specific_terms.values()], axis=1)
@@ -396,7 +400,11 @@ def test_laplace():
 def test_prior_predictive(crossed_data):
     crossed_data["count"] = (crossed_data["Y"] - crossed_data["Y"].min()).round()
     # New default priors are too wide for this case... something to keep investigating
-    model = Model("count ~ threecats + continuous + dummy", crossed_data, family="poisson",)
+    model = Model(
+        "count ~ threecats + continuous + dummy",
+        crossed_data,
+        family="poisson",
+    )
     model.build()
     print(model)
     pps = model.prior_predictive(draws=500)
@@ -496,7 +504,12 @@ def test_potentials():
     ]
 
     model = Model(
-        "w ~ 1", data, family="bernoulli", link="identity", priors=priors, potentials=potentials,
+        "w ~ 1",
+        data,
+        family="bernoulli",
+        link="identity",
+        priors=priors,
+        potentials=potentials,
     )
     model.build()
     assert len(model.backend.model.potentials) == 2
--- a/bambi/tests/test_model_construction.py
+++ b/bambi/tests/test_model_construction.py
@@ -17,7 +17,12 @@ from bambi.priors import Prior
 
 @pytest.fixture(scope="module")
 def data_numeric_xy():
-    data = pd.DataFrame({"y": np.random.normal(size=100), "x": np.random.normal(size=100),})
+    data = pd.DataFrame(
+        {
+            "y": np.random.normal(size=100),
+            "x": np.random.normal(size=100),
+        }
+    )
     return data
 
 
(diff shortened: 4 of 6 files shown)


tomicapretto:
@markgoodhead thanks for opening this PR!

Initially this looks good! The only issue I would address is the name of the scale parameter in the Laplace distribution. In some places you're using `b` while in others you're using `scale`. From [the docs](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.Laplace.html) I see the correct name is `b`.

bambi/defaults/defaults.py
```diff
@@ -117,6 +119,17 @@
         "link": "log",
         "family": NegativeBinomial,
     },
+    "laplace": {
+        "likelihood": {
+            "name": "Laplace",
+            "args": {
+                "scale": "HalfNormal"
```
tomicapretto, line 126: `"scale"` should be `"b"`

bambi/families/likelihood.py
```diff
@@ -15,6 +15,7 @@
     "Multinomial": DistSettings(params=("p",), parent="p", args=None),
     "Normal": DistSettings(params=("mu", "sigma"), parent="mu", args=("sigma",)),
     "NegativeBinomial": DistSettings(params=("mu", "alpha"), parent="mu", args=("alpha",)),
+    "Laplace": DistSettings(params=("mu", "scale"), parent="mu", args=("scale",)),
```
tomicapretto, line 18: `"scale"` should be `"b"`

bambi/families/univariate.py
```diff
@@ -90,6 +90,17 @@ def posterior_predictive(self, model, posterior, linear_predictor):
         return np.random.negative_binomial(n, p)
 
 
+class Laplace(UnivariateFamily):
+    SUPPORTED_LINKS = ["identity", "log", "inverse"]
+
+    def posterior_predictive(self, model, posterior, linear_predictor):
+        "Sample from posterior predictive distribution"
+        mean = self.link.linkinv(linear_predictor)
+        scale = posterior[model.response.name + "_scale"].values
```
tomicapretto, line 99: Here you also have to rename `"_scale"` with `"_b"`

tomicapretto (changes requested):

tomicapretto:
@markgoodhead for the missing `pm.Normal` call, you're right. Bambi implements a very general solution that avoids direct calls to pm.Normal, pm.Beta, etc. etc... It uses the name of the Likelihood function within the Family to select the appropiate PyMC distribution. Once that distribution is selected, it's a `dist(...)` call. All that happens in the following chunk

https://github.com/bambinos/bambi/blob/762f30a995a383f7588110f4343b765ed939ec70/bambi/backend/terms.py#L267-L295

markgoodhead:
Thanks very much @tomicapretto ! I'll have a look at those changes now. I also spotted there's some interaction with the centered/non-centered functionality going on here for the Normal link case:

https://github.com/bambinos/bambi/blob/762f30a995a383f7588110f4343b765ed939ec70/bambi/backend/terms.py#L134-L150

From my experiments before the Laplace link has similar trade-offs to the Normal in terms of when to use the centered/non-centered parameterisation so perhaps this would also need modifying to switch between Normal and Laplace to implement both the centered/non-centered versions of both?

markgoodhead:
> @markgoodhead for the missing `pm.Normal` call, you're right. Bambi implements a very general solution that avoids direct calls to pm.Normal, pm.Beta, etc. etc... It uses the name of the Likelihood function within the Family to select the appropiate PyMC distribution. Once that distribution is selected, it's a `dist(...)` call. All that happens in the following chunk
> 
> https://github.com/bambinos/bambi/blob/762f30a995a383f7588110f4343b765ed939ec70/bambi/backend/terms.py#L267-L295

On this point - it looks like Laplace won't need any special casing so I wouldn't need to do any changes here? In which case I think the outstanding changes would be a) the example you requested (I assume you mean a notebook in the examples folder?) and b) whatever changes you suggest to ensure compatibility with the `self.noncentered` parameter?

markgoodhead:
> > Thanks very much @tomicapretto ! I'll have a look at those changes now. I also spotted there's some interaction with the centered/non-centered functionality going on here for the Normal link case:
> > https://github.com/bambinos/bambi/blob/762f30a995a383f7588110f4343b765ed939ec70/bambi/backend/terms.py#L134-L150
> > 
> > From my experiments before the Laplace link has similar trade-offs to the Normal in terms of when to use the centered/non-centered parameterisation so perhaps this would also need modifying to switch between Normal and Laplace to implement both the centered/non-centered versions of both?
> 
> Good catch!
> 
> This is not an adjustment that happens only for the Normal family (i.e. using Normal likelihood function). This happens for all group-specific effects (i.e. the ones you get with `1 | group` and `variable | group` syntax) no matter the family. So when you add the laplace family (i.e. Laplace likelihood), we will have this centered/non-centered parametrization for its group-specific effects as well.

Ah yes of course - my mistake! 

In that case I think the only change outstanding is the example you requested. I took a look at the notebooks in the example folder and they each seem like quite detailed explanations of a particular type of statistical technique, whereas this is sort of more 'minor' (just another family) so I'm not sure if it's appropriate for me just to add a very small notebook here or if there's a better place the example should live?

tomicapretto:
> > > Thanks very much @tomicapretto ! I'll have a look at those changes now. I also spotted there's some interaction with the centered/non-centered functionality going on here for the Normal link case:
> > > https://github.com/bambinos/bambi/blob/762f30a995a383f7588110f4343b765ed939ec70/bambi/backend/terms.py#L134-L150
> > > 
> > > From my experiments before the Laplace link has similar trade-offs to the Normal in terms of when to use the centered/non-centered parameterisation so perhaps this would also need modifying to switch between Normal and Laplace to implement both the centered/non-centered versions of both?
> > 
> > 
> > Good catch!
> > This is not an adjustment that happens only for the Normal family (i.e. using Normal likelihood function). This happens for all group-specific effects (i.e. the ones you get with `1 | group` and `variable | group` syntax) no matter the family. So when you add the laplace family (i.e. Laplace likelihood), we will have this centered/non-centered parametrization for its group-specific effects as well.
> 
> Ah yes of course - my mistake!
> 
> In that case I think the only change outstanding is the example you requested. I took a look at the notebooks in the example folder and they each seem like quite detailed explanations of a particular type of statistical technique, whereas this is sort of more 'minor' (just another family) so I'm not sure if it's appropriate for me just to add a very small notebook here or if there's a better place the example should live?

No need to add a notebook at this time.. Just something like the example in this PR https://github.com/bambinos/bambi/pull/490

aloctavodia:
I agree that there is no need to add a notebook as part of this PR, opening an issue as a remainder that the example is missing should be enough. Additionally, instead of writing a new notebook, we can maybe extend this one https://bambinos.github.io/bambi/main/notebooks/t_regression.html

markgoodhead:
OK I've got a similar example that (appears to me anyway) to work on this branch - is this what you were expecting?

```
import arviz as az
import bambi as bmb
import numpy as np
import pandas as pd

size = 1000
x = rng.normal(loc=10., scale=5., size=size)
data = pd.DataFrame(
    {
        "x": x,
        "y": rng.laplace(loc=x, size=size)
    }
)

bmb_model = bmb.Model("y ~ x", data, family="laplace")
idata = bmb_model.fit(random_seed=0)

bmb_model.predict(idata)
az.plot_trace(idata);
```
image

tomicapretto (approved):

tomicapretto:
I forgot to add that it would be good to have a test for this new family. Something like the examples attached would be good for now

https://github.com/bambinos/bambi/blob/762f30a995a383f7588110f4343b765ed939ec70/bambi/tests/test_built_models.py#L546-L548

https://github.com/bambinos/bambi/blob/762f30a995a383f7588110f4343b765ed939ec70/bambi/tests/test_built_models.py#L461-L463

markgoodhead:
> Thanks a lot for this contribution! I hope it was a nice experience to you too.
> 
> If I can ask one more thing it would be comments about how easy/hard it was to work through this PR and what we could do to help you. Thanks again!

Sure - I can't think of anything big that'd change to make the process easier. There's always friction learning a new codebase but this wasn't too bad - the code is well laid out and generally quite readable.

Only two quite minor things I can think of - you can add the guidelines (which mention black and pylint) to be displayed on a PR as you open it (or a link to them) as I didn't spot those until quite late. Also, I got caught when my version of black was too old and did some bad formatting, perhaps the versions of pylint and black used in github actions could be specified in the PR guidelines?

Issue #525, opened 2022-06-07
Add example of PyMC Jax and Black Jax


canyon289:
creating per conversation with @zwelitunyiswa in this repository https://github.com/bambinos/bambi/issues/522

To make a PR you'll need to
- [ ] Use git to clone this repository
- [ ] Follow the setup instructions
- [ ] Add a documentation and compile it locally to check it worked
- [ ] Make a PR on github

zwelitunyiswa:
@canyon289  Is this still necessary given that Jax has been implemented with #522  ?

aloctavodia:
Yes, we still need an example. @zwelitunyiswa do you want to do it?

zwelitunyiswa:
Yes, was just checking . Work on it tomorrow.

speco29:
I can add those examples, would you mind telling me where to add these?
I'm a bit confused.

tomicapretto:
We already have this https://bambinos.github.io/bambi/notebooks/alternative_samplers.html

Issue #522, opened 2022-06-06
PyMC Jax support


markgoodhead:
Hello, 

Are there plans to add/support the Jax (and hence GPU) backend for PyMC 4 now that that's been released? I just tried with the latest release (admittedly only out for a couple of hours!) and it looks to be sampling with the CPU still (unsure which backend is being used)?

Thanks very much!

tomicapretto:
@zwelitunyiswa if I recall correctly you did some with JAX? Was it using Bambi or just PyMC?

zwelitunyiswa:
@tomicapretto @canyon289   I was able to get Jax (numpyro and blackjax) working for Bambi (I have an M1 Mac, btw). I installed PYMC 4 first, then Bambi. Then conda install jaxlib, jax and numpyro. I then built the formula and model. Then, using the trick that you and @aloctavodia taught me: I used model.build() and then passed that to PYMC4 and it ran with no problem. I attached a screenshot.

<img width="1368" alt="Screen Shot 2022-06-07 at 9 14 04 AM" src="https://user-images.githubusercontent.com/32966743/172389237-aba3f90f-326b-4daf-9946-9ff7629b26d3.png">

canyon289:
Very cool and thank you for sharing! Would you be willing to add some documentation showing how to do it? if so we can provide some more guidance as well!

zwelitunyiswa:
Here is blackjax:

<img width="1133" alt="Screen Shot 2022-06-07 at 9 29 08 AM" src="https://user-images.githubusercontent.com/32966743/172392521-9479eeb3-ffd1-46c8-9938-1ffc05752a5b.png">

zwelitunyiswa:
> Very cool and thank you for sharing! Would you be willing to add some documentation showing how to do it? if so we can provide some more guidance as well!

Do you mean for the docs on Bambi or just on this thread? Also, do you mean the installation process + model building or just model building? I would be happy to document the process but have no idea how to use Github (lol, I know). If you guys could help me, I would be happy to put together the guide.

canyon289:
@zwelitunyiswa official docs for bambi! We'd be happy to talk you throught github and whatnot. I'll open an issue and tag you

zwelitunyiswa:
@canyon289 Ok. That works. Thanks for the help and opportunity.

aloctavodia:
Thanks @zwelitunyiswa for contributing with this example. 

[...]

PR #526, opened 2022-06-08
Add support for numpyro and blackjax PyMC samplers

markgoodhead:
This is to address https://github.com/bambinos/bambi/issues/522 and https://github.com/bambinos/bambi/issues/525 inspired by @zwelitunyiswa's example

I decided to add a single new value to the fit() method which allows switching in of numpyro/blackjax samplers instead of the pymc default. I decided against some cpu/gpu flags because it's mostly decided by whatever Jax can find and the methods I saw to disable GPUs are [quite hacky](https://github.com/martiningram/mcmc_runtime_comparison/blob/e4a0782835a8921435f0699c84e2038960370389/fit_pymc_jax.py#L16-L18) involving playing with your environment variables which I felt is out of scope for a library to be fiddling with so I've just noted this in the documentation instead. 

I've tested the samplers locally and they work on one of my personal projects, but I'll try and knock up a simple example shortly which demonstrates them all.

One note: The [PyMC 4 release blog post](https://www.pymc.io/blog/v4_announcement.html) says: 

> These samplers live in a different submodule sampling_jax but the plan is to integrate them into pymc.sample(backend="JAX").

So we should expect the implementation here to change pretty soon, so I think it's worth keeping the implementation in bambi simple so it's easy to port-over when this happens.

### DIFF
--- a/bambi/__init__.py
+++ b/bambi/__init__.py
@@ -1,7 +1,5 @@
 import logging
 
-from pymc import math
-
 from .data import clear_data_home, load_data
 from .families import Family, Likelihood, Link
 from .models import Model
--- a/bambi/__init__.py
+++ b/bambi/__init__.py
@@ -1,5 +1,7 @@
 import logging
 
+from pymc import math
+
 from .data import clear_data_home, load_data
 from .families import Family, Likelihood, Link
 from .models import Model
--- a/bambi/backend/pymc.py
+++ b/bambi/backend/pymc.py
@@ -250,13 +250,14 @@ class PyMCModel:
                     else:
                         raise
             elif sampler_backend == "nuts_numpyro":
-                import pymc.sampling_jax  # Lazy import to not force users to install Jax
+                # Lazy import to not force users to install Jax
+                import pymc.sampling_jax  # pylint: disable=import-outside-toplevel
 
                 if not chains:
                     chains = (
                         4  # sample_numpyro_nuts does not handle chains = None like pm.sample does
                     )
-                idata = pm.sampling_jax.sample_numpyro_nuts(
+                idata = pymc.sampling_jax.sample_numpyro_nuts(
                     draws=draws,
                     tune=tune,
                     chains=chains,
@@ -264,13 +265,14 @@ class PyMCModel:
                     **kwargs,
                 )
             elif sampler_backend == "nuts_blackjax":
-                import pymc.sampling_jax  # Lazy import to not force users to install Jax
+                # Lazy import to not force users to install Jax
+                import pymc.sampling_jax  # pylint: disable=import-outside-toplevel
 
                 if not chains:
                     chains = (
                         4  # sample_blackjax_nuts does not handle chains = None like pm.sample does
                     )
-                idata = pm.sampling_jax.sample_blackjax_nuts(
+                idata = pymc.sampling_jax.sample_blackjax_nuts(
                     draws=draws,
                     tune=tune,
                     chains=chains,
--- a/bambi/backend/pymc.py
+++ b/bambi/backend/pymc.py
@@ -257,7 +257,11 @@ class PyMCModel:
                         4  # sample_numpyro_nuts does not handle chains = None like pm.sample does
                     )
                 idata = pm.sampling_jax.sample_numpyro_nuts(
-                    draws=draws, tune=tune, chains=chains, random_seed=random_seed, **kwargs,
+                    draws=draws,
+                    tune=tune,
+                    chains=chains,
+                    random_seed=random_seed,
+                    **kwargs,
                 )
             elif sampler_backend == "nuts_blackjax":
                 import pymc.sampling_jax  # Lazy import to not force users to install Jax
@@ -267,7 +271,11 @@ class PyMCModel:
                         4  # sample_blackjax_nuts does not handle chains = None like pm.sample does
                     )
                 idata = pm.sampling_jax.sample_blackjax_nuts(
-                    draws=draws, tune=tune, chains=chains, random_seed=random_seed, **kwargs,
+                    draws=draws,
+                    tune=tune,
+                    chains=chains,
+                    random_seed=random_seed,
+                    **kwargs,
                 )
             else:
                 raise ValueError(
--- a/bambi/backend/pymc.py
+++ b/bambi/backend/pymc.py
@@ -3,7 +3,6 @@ import traceback
 
 import numpy as np
 import pymc as pm
-import pymc.sampling_jax
 
 import aesara.tensor as at
 
@@ -84,12 +83,11 @@ class PyMCModel:
         chains=None,
         cores=None,
         random_seed=None,
-        sampler_backend="default",
         **kwargs,
     ):
         """Run PyMC sampler."""
         # NOTE: Methods return different types of objects (idata, approximation, and dictionary)
-        if method.lower() == "mcmc":
+        if method.lower() in ["mcmc", "mcmc-numpyro", "mcmc-blackjax"]:
             result = self._run_mcmc(
                 draws,
                 tune,
@@ -101,7 +99,7 @@ class PyMCModel:
                 chains,
                 cores,
                 random_seed,
-                sampler_backend,
+                method.lower(),
                 **kwargs,
             )
         elif method.lower() == "vi":
@@ -212,11 +210,11 @@ class PyMCModel:
         chains=None,
         cores=None,
         random_seed=None,
-        sampler_backend="default",
+        sampler_backend="mcmc",
         **kwargs,
     ):
         with self.model:
-            if sampler_backend == "default":
+            if sampler_backend == "mcmc":
                 try:
                     idata = pm.sample(
                         draws=draws,
@@ -248,7 +246,8 @@ class PyMCModel:
                         )
                     else:
                         raise
-            elif sampler_backend == "numpyro":
+            elif sampler_backend == "mcmc-numpyro":
+                import pymc.sampling_jax  # Lazy import to not force users to install Jax
                 idata = pm.sampling_jax.sample_numpyro_nuts(
                     draws=draws,
                     tune=tune,
@@ -256,7 +255,8 @@ class PyMCModel:
                     random_seed=random_seed,
                     **kwargs,
                 )
-            elif sampler_backend == "blackjax":
+            elif sampler_backend == "mcmc-blackjax":
+                import pymc.sampling_jax  # Lazy import to not force users to install Jax
                 idata = pm.sampling_jax.sample_blackjax_nuts(
                     draws=draws,
                     tune=tune,
--- a/bambi/backend/pymc.py
+++ b/bambi/backend/pymc.py
@@ -271,7 +271,7 @@ class PyMCModel:
             else:
                 raise ValueError(
                     f'sampler_backend value {sampler_backend} is not valid. Please choose one of'
-                    f'``default``, ``numpyro`` or ``blackjax``'
+                    f'``mcmc``, ``mcmc-numpyro`` or `mcmc-blackjax``'
                 )
 
         idata = self._clean_mcmc_results(idata, omit_offsets, include_mean)
--- a/bambi/backend/pymc.py
+++ b/bambi/backend/pymc.py
@@ -248,6 +248,8 @@ class PyMCModel:
                         raise
             elif sampler_backend == "mcmc-numpyro":
                 import pymc.sampling_jax  # Lazy import to not force users to install Jax
+                if not chains:
+                    chains = 4  # sample_numpyro_nuts does not handle chains = None like pm.sample does
                 idata = pm.sampling_jax.sample_numpyro_nuts(
                     draws=draws,
                     tune=tune,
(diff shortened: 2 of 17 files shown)


markgoodhead:
OK so I've hit a snag I don't quite understand. My personal project works fine but my test example fails:

`import arviz as az
import bambi as bmb
import numpy as np
import pandas as pd
import time

size = 1000
x = rng.normal(size=size)
print(x)
data = pd.DataFrame(
    {
        "x": x,
        "y": rng.normal(loc=x, size=size)
    }
)
print(data)

bmb_model = bmb.Model("y ~ x", data)
bmb_model_numpyro = bmb.Model("y ~ x", data)
bmb_model_blackjax = bmb.Model("y ~ x", data)
t0 = time.time()
idata = bmb_model.fit()
t1 = time.time()
idata_numpyro = bmb_model_numpyro.fit(chains=4, tune=1000, draws=1000, sampler_backend="numpyro", chain_method="vectorized")
t2 = time.time()
idata_blackjax = bmb_model_blackjax.fit(chains=4, tune=1000, draws=1000, sampler_backend="blackjax", chain_method="vectorized")
t3 = time.time()

print(f"Default: {t1-t0} Numpyro: {t2-t1} Blackjax: {t3-t2}")`

It fails on line https://github.com/bambinos/bambi/blob/762f30a995a383f7588110f4343b765ed939ec70/bambi/backend/pymc.py#L322 during the numpyro run because it appears that one of the idata attributes is read-only... must admit this is quite deep into the weeds of bambi/pymc internals and I'm a little stuck as to how to proceed.

bambi/backend/pymc.py
```diff
@@ -3,6 +3,7 @@
 
 import numpy as np
 import pymc as pm
+import pymc.sampling_jax
```
aloctavodia, line 6: This will call Jax and we want to keep Jax optional, so move this down.

bambi/models.py
```diff
@@ -239,6 +240,17 @@ def fit(
             in the system unless there are more than 4 CPUs, in which case it is set to 4.
         random_seed : int or list of ints
             A list is accepted if cores is greater than one.
+        sampler_backend: str
+            If ``default`` uses the standard PyMC fit() method. Is only valid if method == "mcmc".
```
aloctavodia, line 244:
```suggestion
            If ``default`` uses the standard PyMC sampling method. Is only valid if method == "mcmc".
```

bambi/models.py
```diff
@@ -239,6 +240,17 @@ def fit(
...
+            Other valid sampling methods are ``numpyro`` and ``blackjax`` which both use JAX (see
+            https://www.pymc.io/blog/v4_announcement.html#new-jax-backend-for-faster-sampling)
+            and which offers both performance improvements on the CPU due to JIT compilation
+            and also GPU use (which will happen automatically if a GPU is detected - see the
+            JAX documentation https://jax.readthedocs.io/en/latest/index.html). Also, both
```
aloctavodia, line 249:
```suggestion
            Other valid sampling methods are ``numpyro`` and ``blackjax`` which both use JAX. Also, both
```
In my opinion, this is too verbose for the docstring. But it will be very nice to have as part of an example.

aloctavodia (changes requested):
@markgoodhead Thanks for your contribution. I have a few comments.

bambi/models.py
```diff
@@ -239,6 +240,17 @@ def fit(
             in the system unless there are more than 4 CPUs, in which case it is set to 4.
         random_seed : int or list of ints
             A list is accepted if cores is greater than one.
+        sampler_backend: str
```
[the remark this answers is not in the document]
markgoodhead: Yes I think that's better - I'll add this now

aloctavodia:
> ```python
> chain_method="vectorized"
> ```

This is weird, this runs for me with both numpyro and blackjax

markgoodhead:
How odd! Perhaps my environment isn't setup correctly and I'm behind on the latest versions. What versions of pymc/jax/arviz/xarray etc are you using?

I have modified my example script to work now with the new method arg approach (and fixed a bug in chains handling I spotted)

`import arviz as az
import bambi as bmb
import numpy as np
import pandas as pd
import time

size = 1000
x = rng.normal(size=size)
data = pd.DataFrame(
    {
        "x": x,
        "y": rng.normal(loc=x, size=size)
    }
)

bmb_model = bmb.Model("y ~ x", data)
bmb_model_numpyro = bmb.Model("y ~ x", data)
bmb_model_blackjax = bmb.Model("y ~ x", data)
t0 = time.time()
idata = bmb_model.fit()
t1 = time.time()
idata_numpyro = bmb_model_numpyro.fit(method="mcmc-numpyro", chain_method="vectorized")
t2 = time.time()
idata_blackjax = bmb_model_blackjax.fit(method="mcmc-blackjax", chain_method="vectorized")
t3 = time.time()

print(f"Default: {t1-t0} Numpyro: {t2-t1} Blackjax: {t3-t2}")`

bambi/models.py
```diff
@@ -208,6 +208,11 @@ def fit(
             using the ``fit`` function.
             Finally, ``"laplace"``, in which case a Laplace approximation is used and is not
             recommended other than for pedagogical use.
+            To use the PyMC numpyro and blackjax samplers, use ``mcmc-numpyro`` or ``mcmc-blackjax``
+            respectively. Both methods will only work if you can use NUTS sampling, so your model must
+            be differentiable. It is also recommended, if using ``mcmc-numpyro`` or
+            ``mcmc-blackjax``, to set the kwarg of ``chain_method`` to either ``parallel`` or
+            ``vectorized`` for optimal performance.
```
aloctavodia, line 215:
```suggestion
            To use the numpyro and blackjax samplers, use ``nuts_numpyro`` or ``nuts_blackjax``
            respectively. Both methods will only work for differentiable models. If using ``nuts_numpyro`` or
            ``nuts_blackjax``, it is recomended to set the kwarg ``chain_method`` to either ``parallel`` or
            ``vectorized`` for optimal performance.
```

aloctavodia: Not clear to me what users should do with the chain_method argument. How should they pick between ``vectorized`` or ``parallel``?

zwelitunyiswa:
Vecrorized does work for both blackjack and numpyro(
https://www.pymc.io/projects/docs/en/stable/api/samplers.html) but I found
vectorized was slower on cpu (and my gpu won’t work since I have an M1).

bambi/__init__.py
```diff
@@ -1,7 +1,5 @@
 import logging
 
-from pymc import math
```
aloctavodia, line 3: is this right?

bambi/models.py
markgoodhead: I can add some explanation if you like here if it won't make the docs a bit too verbose? It's whether they use pmap or vmap in Jax (https://jax.readthedocs.io/en/latest/jax-101/06-parallelism.html). It depends a lot on whether you are running on a GPU or CPU, how many GPUs you have etc... maybe better for users to experiment themselves?

bambi/backend/pymc.py
```diff
@@ -209,40 +210,69 @@ def _run_mcmc(
         chains=None,
         cores=None,
         random_seed=None,
+        sampler_backend="mcmc",
```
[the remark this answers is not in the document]
markgoodhead: There is no 'method' variable in this call, I could rename it to method if you prefer for consistency?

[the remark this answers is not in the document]
ricardoV94: shouldn't the name be pymc_nuts?

bambi/models.py
markgoodhead: The default of 'sequential' is generally slower in all cases I can think of - both CPUs and GPUs can benefit from vectorisation and parallelisation (unless you're on a one core machine from a few decades ago!) it's kind of a crazy default I think.

aloctavodia: We can have some explanation as part of the notebook example. I made another suggestion trying to provide some explanation but keeping concise, feel free to improve it. Notice that in PyMC the default is parallel

aloctavodia:
Looks good, the only missing part is a test

aloctavodia:
Do you mind adding the test anyway?

markgoodhead:
> with

Tests added... Fingers crossed they actually work!

aloctavodia:
Be sure to run black and pylint https://github.com/bambinos/bambi/blob/main/CONTRIBUTING.md#pull-request-checklist

markgoodhead:
> Be sure to run black and pylint https://github.com/bambinos/bambi/blob/main/CONTRIBUTING.md#pull-request-checklist

Done 👍

markgoodhead:
One small issue is that pylint isn't happy with the import within the code itself - I assume you're happy to ignore the error here?

tomicapretto:
> One small issue is that pylint isn't happy with the import within the code itself - I assume you're happy to ignore the error here?

You could add `# pylint: disable=import-outside-toplevel` next to the import, I think that should work.

markgoodhead:
I just tried updating my version of xarray (which was 0.21.1 before) to the latest on pip (2022.3.0) and I still get the same xarray error... otherwise my versions are all compatible with the pymc 4.0.0 release on pip. Does anyone else get this error? If not, what versions of xarray etc are you using?

markgoodhead:
Tests look to be failing due to Jax not being installed (`ModuleNotFoundError: No module named 'jax'`) - @tomicapretto please can you assist?

tomicapretto:
I think we could have something like `requirements-optional.txt` like what you can find in ArviZ https://github.com/arviz-devs/arviz. I think `requirements-dev.txt` should be only for development dependencies. 

Below, you will need to add another line saying `pip install -r requirements-optional.txt`.

https://github.com/bambinos/bambi/blob/ecfdbf28c8ccdcfefa06c4a50ba75f0f65321325/.github/workflows/test.yml#L35-L36

aloctavodia:
Agreed, that's cleaner.

canyon289:
Please also ensure the optional dependencies in `setup.py` is setup correctly for optional requirements

https://stackoverflow.com/a/43090648/414104

markgoodhead:
I've added the optional requirements files and hopefully done the setup.py changes @canyon289 requested correctly (all a bit new to me so I could well have done it wrong!). I wasn't sure what versions to specify in the file so I tried to find the equivalents in pymc to align with what they have... and was a bit surprised when I [couldn't find any](https://github.com/pymc-devs/pymc/blob/main/requirements.txt)! Perhaps something similar should be added to pymc and then by depending on a specific pymc version this would flow naturally upstream to bambi?

markgoodhead:
OK it looks like the tests are failing for the same reason my local environment doesn't work which I've no idea how to fix! Anyone got any advice what I should try/do here?

```Python
2022-06-09T09:04:21.8572705Z bambi/models.py:265: in fit
2022-06-09T09:04:21.8572923Z     return self.backend.run(
2022-06-09T09:04:21.8573156Z bambi/backend/pymc.py:91: in run
2022-06-09T09:04:21.8573373Z     result = self._run_mcmc(
2022-06-09T09:04:21.8573593Z bambi/backend/pymc.py:288: in _run_mcmc
2022-06-09T09:04:21.8573880Z     idata = self._clean_mcmc_results(idata, omit_offsets, include_mean)
2022-06-09T09:04:21.8574228Z bambi/backend/pymc.py:363: in _clean_mcmc_results
2022-06-09T09:04:21.8574631Z     idata.posterior[intercept_name] -= np.dot(X.mean(0), coefs).reshape(shape)
2022-06-09T09:04:21.8575105Z /usr/share/miniconda/envs/test/lib/python3.8/site-packages/xarray/core/_typed_ops.py:290: in __isub__
2022-06-09T09:04:21.8575466Z     return self._inplace_binary_op(other, operator.isub)
2022-06-09T09:04:21.8575943Z /usr/share/miniconda/envs/test/lib/python3.8/site-packages/xarray/core/dataarray.py:3121: in _inplace_binary_op
2022-06-09T09:04:21.8576273Z     f(self.variable, other_variable)
2022-06-09T09:04:21.8576690Z /usr/share/miniconda/envs/test/lib/python3.8/site-packages/xarray/core/_typed_ops.py:480: in __isub__
2022-06-09T09:04:21.8577039Z     return self._inplace_binary_op(other, operator.isub)
```

bambi/tests/test_built_models.py
```diff
@@ -338,6 +338,40 @@ def test_laplace_regression():
...
+
+
+def test_logistic_regression_blackjax():
+    y = pd.Series(np.random.choice(["a", "b"], 50), dtype="category")
+    data = pd.DataFrame({"y": y, "x": np.random.normal(size=50)})
+    model = Model("y ~ x", data, family="bernoulli")
+    model.fit(method="nuts_blackjax", chain_method="vectorized")
+
+
+def test_regression_blackjax():
```
canyon289, line 355: Nit: These two tests `test_regression_blackjax` and  `test_regression_nunpyro` could be parameterized to reduce amount of code that needs to be read or maintained

canyon289:
@markgoodhead thanks for doing this! this is a great capability add for bambi

bambi/tests/test_built_models.py
aloctavodia: I agree with @canyon289 here, but if you want @markgoodhead you can open an issue fix this later.

PR #532, opened 2022-06-11
Remove retry logic in mcmc as pymc already does this

markgoodhead:
This came up in a recent PR (https://github.com/bambinos/bambi/pull/526#discussion_r893729628) but I think PyMC already handles this retry logic so bambi doesn't need to any more (I'm having trouble finding the exact code in pymc, although I've definitely seen it happen in my non-bambi pymc models!)

### DIFF
--- a/bambi/backend/pymc.py
+++ b/bambi/backend/pymc.py
@@ -215,40 +215,17 @@ class PyMCModel:
     ):
         with self.model:
             if sampler_backend == "mcmc":
-                try:
-                    idata = pm.sample(
-                        draws=draws,
-                        tune=tune,
-                        discard_tuned_samples=discard_tuned_samples,
-                        init=init,
-                        n_init=n_init,
-                        chains=chains,
-                        cores=cores,
-                        random_seed=random_seed,
-                        **kwargs,
-                    )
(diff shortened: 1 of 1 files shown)


aloctavodia:
Hi, not sure this is done by PyMC. Something similar that pymc now does is jittering a few times until the starting values produce a finite logprobability. We should not remove this until we are certain that this routine, or something equivalent is  implemeted upstream.

markgoodhead:
OK I'll have a proper dive into the pymc code later and try and find this - I'd swear I've seen my pymc model explicitly logging it was changing to

aloctavodia:
I can confirm that pymc is not doing this. I checked the code and double checked by asking the other pymc-devs.

tomicapretto:
If PyMC is not doing it, I think we could keep this code in Bambi. See https://github.com/bambinos/bambi/issues/381 where it was asked. I think it can prevent a headache for those taking their first steps with Bambi and Bayesian modeling in general.

markgoodhead:
Ah OK apologies - I'm curious what exactly it was I was misinterpreting in pymc now! PR closed

Issue #541, opened 2022-06-26
Censored Data in Bambi


don-jil:
First, I'm delighted to find babmi and formulae, and very much appreciate all the work you are doing in this space. I'm working with some censored data.  I have used brms in R to build models in STAN, but I'd like to transition to python where the data preparation and visualization work occurs.  As a feature request, I would very much like to have the ability to use bambi (and formulae) to build the pymc model for censored data and survival analysis. I am currently working on this in pymc without bambi (and formulae), but I love the Wilkinson formula approach.  If this should be posted on the formulae repo instead, I'd be happy to do that.

ipa:
I was just about to open a similar request... I've built a generic survival model using the pymc-learn API and was thinking about contributing it to bambi. The model uses the Weibull distribution (I don't have a Cox model). However, it has only been tested with 'right' censored data. 

My proposed syntax for the formula would be: `time | censor ~ x1 + x2 + x3`.
`censor` would contain information about censoring as ('left', 'none', 'right', and 'interval') or (-1, 0, 1, and 2) analogous to brms. 

Let's discuss. Would be happy to contribute on this.

ipa:
This requires changes in the formulae package. Instead of `time | censor` it would be cleaner to implement this as a transformation `Surv(time, event)` similar to R.

I opened another issue bambinos/formulae#79

Work in progress on my fork

don-jil:
Just a ping, here.  I love that this is moving along & that there is a really pull-request @ipa!  Is there anything that is missing @tomicapretto?  Anything I can do to help?

tomicapretto:
@ipa @don-jil I haven't had much time to work in Bambi lately. But I'm slowly starting to gain momentum again.

The thing with the PR opened is that it will just work for that particular case. I think we need something more general and that requires working both in formulae and bambi internals if I recall correctly. 

I think I'll have something in my hands in the next 15 days. Would you be interested in helping to test/document stuff?

don-jil:
Sorry for the delayed response, but yes happy to help test and or document.  :-) Will respond in the pull request to that end.

tomicapretto:
I'm working on a proposal for models for censored data in Bambi. The problem is that I'm not very familiar with interval censoring (we can decide not to support this in a first shot), but I would still like to understand how it works.

@ipa @don-jil do you know of any references to interval censoring? Thanks!

tomicapretto:
Closed since #577

PR #543, opened 2022-07-06
Adding support for survival analysis

ipa:
This PR adds support for survival analysis using the Exponential distribution. It adds support for the exponential family and for censored data. 

There is a notebook with examples using R's veteran dataset. 

Issue: #541 
Depends on: bambinos/formulae/issues/79


canyon289:
I want to say this is one of the best PRs ive ever seen in open source. Between the well formatted and readable code, tests, and docs I dont even know what more to ask for. Thanks so much for creating this

tomicapretto:
@ipa thanks a lot for this fantastic PR!

As I mentioned in the [formulae PR](https://github.com/bambinos/formulae/pull/80#issuecomment-1178160492)  it's not needed to add a new transformation in there (unless we need to treat involved variables in a very special way). You can just add a `surv` function into the `extra_namespace` dictionary that's defined here

https://github.com/bambinos/bambi/blob/236021fc0860317fdeb365da2c7540f4442ea8c3/bambi/utils.py#L78

that namespace is made available when formulas are parsed:

https://github.com/bambinos/bambi/blob/236021fc0860317fdeb365da2c7540f4442ea8c3/bambi/models.py#L147

I'm adding more comments as I review the code :)

docs/notebooks/survival_model.ipynb
```diff
@@ -0,0 +1,1587 @@
+{
```
tomicapretto, line 1: <p>It would be good to add something telling how <code>Surv</code> works in the model formula. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/543/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

tomicapretto, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    lam = 1/np.exp(az.summary(idata)["mean"]["Intercept"])
</pre><p><code>az.summary()</code> is being called multiple times. We could call it once and store the result in a variable such as <code>idata_summary</code>   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/543/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

tomicapretto, line 1: <pre class="ql-syntax" spellcheck="false">Line #6.    S0 = sp.stats.expon.sf
</pre><p>Could you add a little about the <code>sf</code> function? From SciPy docs it says it's the survival function. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/543/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

tomicapretto, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.        "Surv(time,status) ~ 1 + B(trt_str, success='Test') + T(celltype_str, ref='Squamous')",
</pre><p>I really like you're using <code>B()</code> and <code>T()</code> in the model formula. Again, one sentence saying what they do would be great. I could help here if you want since I was the author of these shortcuts.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/543/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

tomicapretto, line 1: <p>This is a fantastic chart. Looks very nice! Do you think we could think about a way to make it work with plot_cap? https://github.com/bambinos/bambi/pull/517</p><p> </p><p>That implies some modifications to the predict function. I'm not 100% sure if this is possible or not. But it's worth considering it since it would make things much easier.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/543/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

tomicapretto:
I've been reading a little about the documentation of the [`Surv`](https://www.rdocumentation.org/packages/survival/versions/2.11-4/topics/Surv) function in the `survival` R package. By default, it results in right-censored data. 

```r
> library(survival)
> x <- Surv(veteran$time, veteran$status)
> attributes(x)
$dim
[1] 137   2

$dimnames
$dimnames[[1]]
NULL

$dimnames[[2]]
[1] "time"   "status"


$type
[1] "right"

$class
[1] "Surv"
```

As implemented in this PR, using the `exponential` family will always try to result in a survival model, but this shouldn't be the case. It should result in a survival model only when `Surv` is in the LHS of the model formula (I prefer lowercase `surv` or even other alternatives such as `censored(y, status)`). This requires more thinking of course.

ipa:
@tomicapretto thanks for the great review and comments.

I was not aware of the `extra_namespace` but it makes perfect sense to add it there. I'll change that and also implement that a survival model is only results when `Surv` is in the LHS. We can use `censored` as you suggested to keep a more general terminology. 

I will also work on the example notebook and add some extra information.

docs/notebooks/survival_model.ipynb
ipa: I'll check this out and see if it's possible.

tomicapretto:
> 

@ipa feel free to ping me if you want help with something or have a question :)

ipa:
@tomicapretto An easy but not really nice way to implement this is to check for `if "censored" in self.term.name:` while building the response or something similar. 

After thinking about this a bit more I think it would be cleaner if we implement it similar to `Proportion` in the formulae package (introducing a Censored type). This would implement it again in the formulae package though. However, we could then use something like this while building the response:

    if isinstance(self.term.data, Censored):
        __build_censored_response_distribution(...)
 
I tried to do this via `extra_namespace` but then the formulae package throws errors because the type is neither numeric nor any of the other supported types like `Proportion`, `CategoricalBox` etc.

donaldbraman:
Just want to say how excited I am to see such rapid progress on this -- it will be a great addition, and will simplify a lot of work for folks (like me!) running survival analyses.  If I can do anything besides cheer from the sidelines, just let me know. :-)

don-jil:
Happy to help test and or document. Very much appreciate all the work going into this. :-)

donaldbraman:
So, just reviewing the various related issues, it looks like we need a way to let formulae know whether and how the data is left censored, right censored, or both.  Is that the main barrier to implementation?

tomicapretto:
@donaldbraman this is already implemented and it's possible to work with survival models in Bambi since #697 :)
I think I commented this in some issue but somehow it missed this.

You can have a look at this test

https://github.com/bambinos/bambi/blob/169564fe640818525049d7d946636266e4a7b968/tests/test_built_models.py#L1002-L1049

donaldbraman:
Amazing -- you just made my day!  And thank you again for all your work on this. 

> @donaldbraman this is already implemented and it's possible to work with survival models in Bambi since #697 :) I think I commented this in some issue but somehow it missed this.
> 
> You can have a look at this test
> 
> https://github.com/bambinos/bambi/blob/169564fe640818525049d7d946636266e4a7b968/tests/test_built_models.py#L1002-L1049

tomicapretto:
I'm closing this PR because it's already implemented (see this previous comment https://github.com/bambinos/bambi/pull/543#issuecomment-1731239830)

PR #553, opened 2022-08-10
add vi_kwargs to fit

aloctavodia:
`model.fit` has an argument called  `method` as well as `pymc.fit` to avoid those name clashing, this PR introduces a `vi_kwargs`argument that is passed to `pymc.fit`. As `method="mcmc"` is the most common method, we allow passing arguments as kwargs, instead of adding a `mcmc_kwargs argument`. This also add a missing test.

### DIFF
--- a/bambi/backend/pymc.py
+++ b/bambi/backend/pymc.py
@@ -84,6 +84,7 @@ class PyMCModel:
         chains=None,
         cores=None,
         random_seed=None,
+        vi_kwargs=None,
         **kwargs,
     ):
         """Run PyMC sampler."""
@@ -104,7 +105,9 @@ class PyMCModel:
                 **kwargs,
             )
         elif method.lower() == "vi":
-            result = self._run_vi(**kwargs)
+            if vi_kwargs is None:
+                vi_kwargs = {}
+            result = self._run_vi(**vi_kwargs)
         else:
             result = self._run_laplace()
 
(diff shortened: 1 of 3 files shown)


bambi/backend/pymc.py
```diff
@@ -104,7 +105,9 @@ def run(
                 **kwargs,
             )
         elif method.lower() == "vi":
-            result = self._run_vi(**kwargs)
```
canyon289, line 107: Nit: add exception somewhere above that if vi_kwargs is passed but method is not VI either a warning or exception is provided

Also nit: It feels like if theres vi kwargs, there should also be mcmc kwargs to mirror the API

tomicapretto: Agree with the first comment. Not so sure about the second. I understand that having `vi_kwargs` makes us think why not `mcmc_kwargs` as well. But it also makes me wonder whether we need `mcmc_kwargs`, `vi_kwargs`, and `laplace_kwargs`, which makes the signature much longer, or if it is better to replace `method` with a better name that does not overlap with parameters in `pymc.fit()` and keep a single `**kwargs`.

aloctavodia: As method="mcmc" is the most common method, is very convenient to allow for kwargs. Having mcmc_kwargs will turn super annoying. Instead I see vi (and laplace) as unusual methods so is ok if they deviate a little from that pattern. Having said that changing the method argument is an option. I just wanted to avoid breaking backward compatibility, but given that mcmc is the most common method probably that change will most likely unnoticed by most users.

aloctavodia: OK, let's do that

PR #554, opened 2022-08-11
Deprecate method argument in favor of inference_method

aloctavodia:
Following discussion in #553

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -174,7 +174,7 @@ class Model:
         discard_tuned_samples=True,
         omit_offsets=True,
         include_mean=False,
-        method="mcmc",
+        inference_method="mcmc",
         init="auto",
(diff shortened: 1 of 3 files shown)


bambi/models.py
```diff
@@ -247,9 +248,19 @@ def fit(
 
         Returns
         -------
-        An ArviZ ``InferenceData`` instance if method  ``"mcmc"`` (default).
+        An ArviZ ``InferenceData`` instance if inference_method  ``"mcmc"`` (default).
```
tomicapretto, line 251: Nitpick: Can we also add its InferenceData for "mcmc", "nuts_numpyro" and "nuts_blackjax"?

tomicapretto:
Apart from the small comment, looks good to me!

PR #555, opened 2022-08-13
Improve Laplace computation and integration

aloctavodia:
* Perform computations on the unbounded space and return results on the original scale.
* Return InferenceData object

### DIFF
--- a/bambi/backend/pymc.py
+++ b/bambi/backend/pymc.py
@@ -473,11 +473,22 @@ class PyMCModel:
 
         samples = np.random.multivariate_normal(modes, cov, size=draws)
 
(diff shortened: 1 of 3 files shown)


tomicapretto:
This also makes the code in `_run_laplace()` much clearer, that's great.

One extra thing we could have is a docstring in `_posterior_to_idata()`. Maybe it also needs a better name, like `_laplace_posterior_to_idata()`, `_posterior_samples_to_idata()`, or something that is more explicit about the circumstances it's used. The previous name is too general.

aloctavodia:
This is ready to merge from my side

PR #557, opened 2022-08-15
unpin pymc version

aloctavodia:


bambi/tests/test_built_models.py
```diff
@@ -590,6 +590,7 @@ def test_binomial_regression():
     model.fit(draws=10, tune=10)
 
 
+@pytest.mark.skip(reason="this example no longer trigger the fallback to adapt_diag")
```
tomicapretto, line 593: I guess this is good for PyMC and the users, but bad for our tests right? Nevertheless, I think it's not a blocker for the PR. It can be fixed later.

aloctavodia: Yes, I think we should open an issue and investigate if the initialization problem has been solved in general or just for this particular example.

PR #562, opened 2022-08-25
remove upper bound numpy requirement

aloctavodia:


tomicapretto:
We need to remove Python 3.7 here

https://github.com/bambinos/bambi/blob/505fbb3d8fe3e13d4035dd87756be2561bc2a989/.github/workflows/test.yml#L14

should we officially drop Python 3.7 as well?

aloctavodia:
We should if we follow https://numpy.org/neps/nep-0029-deprecation_policy.html that in pymc and arviz we agree to follow a while ago.  But then stuff like colab don't care about this as they are slower to do updates. I think recently pymc drop python 3.7 and then @canyon289 undropit in order to have pymc4 running on colab. Is that right?

canyon289:
that is right. if you drop 3.7 Bambi won't work on colab

aloctavodia:
We can keep, 3.7 and then numpy < 1.22. But that will not solve the dependanbot alerts.

canyon289:
We can just ignore the alerts for now

>  But then stuff like colab don't care about this as they are slower to do updates

To defend the colab team, they do care and theyre really nice people! Whats important to remember is they are responsible for what is likely the **most used** python environment in the entire world so things like changing a python version is a big deal, hence the caution they take when upgrading

aloctavodia:
I mean they don't care about NEP29, that is an attempt to sync the support of Python and NumPy versions across the Scientific Python ecosystem.

canyon289:
> I mean they don't care about NEP29, that is an attempt to sync the support of Python and NumPy versions across the Scientific Python ecosystem.

Ah, got it. I dont want to be pedantic I assure you they do care, just need to be slower about it.

Nonetheless we could take a similar strategy here that we took in PyMC. Release a specific version that supports Python 3.7 and then drop it from testing. When Colab updates to Python 3.8 this issue will be moot!

aloctavodia:
By being  colloquial i sounded like I was giving my personal opinion, but that's was not my intention. I don't have any personally opinion on the Colab team. What I was trying to convey is that the colab team does not enforce NEP29. Enforcing it implies, among other things , to have dropped python 3.7 last December. NEP29 is all about time, so being slow about it is not enforcing it.

Regarding a new release. The last release is 0.9 and that includes pymc 4 but not other stuff like jax sampling. So not sure if that should be the last version compatible with python 3.7 or we instead should release a new one. I am fine with any of those options.

PR #573, opened 2022-10-15
New implementation of predictions

tomicapretto:
This PR doesn't add any new functionality, nor impacts any user-visible change. It modifies how we compute predictions in the `.predict()` method. The computation is more similar to the math now (y = Xb + Zu), and we rely more on xarray to handle the very complicated arrays that we work with. 

The goal is to simplify the codebase, and remove all the hacky operations reshaping numpy arrays so many times.

I'm not sure if this will be the final solution that will be used in the long-term. I imagine that this won't be usable anymore when we get to more complex models (e.g. distributional models). But that's not a problem for us now.


tomicapretto:
I ran all the example notebooks using this branch and everything went fine. I think this can be reviewed @aloctavodia if you have time.

Issue #574, opened 2022-10-15
Make it possible to pass arguments to link functions in PyMC backend


tomicapretto:
We have a warning every time we use a model that uses `aesara.tensor.softmax`. This is because it asks to pass the `axis` explicitly. The fix is not too straightforward because we don't have how to pass arguments to link functions in the PyMC backend.

It's it very doable, but it requires a little of thought.

tomicapretto:
This is the warning

```
/home/tomas/anaconda3/envs/bambi/lib/python3.10/site-packages/aesara/tensor/nnet/basic.py:1116: FutureWarning: Softmax now accepts an axis argument. For backwards-compatibility it defaults to -1 when not specified, but in the future the default will be `None`
```

zwelitunyiswa:
I ran into this warning earlier this week. Question: does this mean that
the softmax accepts both long and wide data for the simplex in aesara. If
so, would it not make sense to switch the default to long format for the
simplex in the dataframe, which is the default for the other other families?

tomicapretto:
This warning is not something that impacts how we use data at a high-level. It's mostly an internal thing that won't change any result. It will just remove the warning and make sure Bambi still works the right way when aesara goes with `axis=None` as default.

tomicapretto:
This is a non-issue anymore since we already fixed the issue when it was needed.

Issue #544, opened 2022-07-11
Design how we're going to extend Bambi


tomicapretto:
The following is a list of features we're missing (or covering only partially) in Bambi

* Distributional models (we model more than the mean parameter of the response)
* Multivariate models (ie the response is a multivariate distribution)
* Non-linear models.
* Survival models/Models with censored data.
* Ordinal models.
* Zero and Zero-One inflated models.

The last three points (survival/censored, ordinal, and zero/zero-one inflated) are covered by the first points (distributional and multivariate) if we implement them appropriately. The third point, non-linear models, is a separate problem. I'll try to add a couple of things I've been thinking about lately. 

---

## Distributional models

Some API proposals

```python
formula = bmb.formula(
    "y ~ a + b",
    "sigma ~ a",
)
priors = {
    "a": bmb.Prior("Normal", mu=0, sigma=1),
    "b": bmb.Prior("Normal", mu=0, sigma=1),
    "sigma_Intercept": bmb.Prior("Normal", mu=0, sigma=1),
    "sigma_x": bmb.Prior("Normal", mu=0, sigma=1)
}
link = {"mu": "identity", "sigma": "log"}
model = bmb.Model(formula, data, priors, link)
```

* We need a formula object where we can have multiple formula parts. I propose to call it `bmb.formula()`. There's an open discussion in #423.
* We need a name for the terms associated with the auxiliary parameters. I propose to use `{param}_{term}` such as `sigma_x`.
* We need a transformation of the linear predictor of the auxiliary parameters into something that makes sense. I propose we have defaults for the built-in families that can be overridden with a dictionary. Note a dictionary is not supported by the `link` argument in `Model` now.

I haven't thought much more about the implementation details, where other concerns may appear. For the moment, I think it's good to discuss about the API we want. Any objections, any suggestions, any drawbacks I'm not seeing?

## Multivariate models

We currently support some multivariate families, such as `"categorical"` and `"multinomial"`. I feel we should think more about the implementation. I think we could make it more general so we don't need to handle all cases as special cases. With that said, I think there are other things to discuss.

* What do we use to indicate a multivariate response?

```python
"c(y1, y2, ..., yn) ~ ..."
"mvbind(y1, y2, ..., yn) ~ ..."
bmb.formula("y1 ~ ...", "y2 ~ ...", "y3 ~ ...")
```
note the last alternative allows for different predictors to be included in each case.

* How much do we want to support multivariate families? 

[...]

PR #582, opened 2022-10-29
Refactor terms

tomicapretto:
Generally speaking, this PR aims to make Bambi internals more modular and extensible. To do that, I removed any logic specific to a given object from another given object. For example, there was Family-specific logic within response terms or the PyMC backend. See 

<details>

https://github.com/bambinos/bambi/blob/e1503d267265c50bd085456b920021e37e28f5cc/bambi/terms.py#L52-L57 

https://github.com/bambinos/bambi/blob/e1503d267265c50bd085456b920021e37e28f5cc/bambi/terms.py#L65-L77 

https://github.com/bambinos/bambi/blob/e1503d267265c50bd085456b920021e37e28f5cc/bambi/backend/pymc.py#L213-L214  

https://github.com/bambinos/bambi/blob/e1503d267265c50bd085456b920021e37e28f5cc/bambi/backend/terms.py#L273-L295

</details>

This approach has been working so far, but if we plan to extend Bambi according to https://github.com/bambinos/bambi/issues/544 without making it messier, we should stop handling specific cases in the same way we've been doing so far. 

In addition, I implemented a couple of extra features/improvements that I'm going to describe below.

**Refactor**

* Moved the `terms.py` module to its own directory, called `terms`.
    * Here we have one Python file per term type. They all inherit from the `BaseTerm` class.
* Many attributes that were computed at creation time, and handled in very specific cases, are not computed at access time. In many cases, the determination of the value is determined by the Family, and not the term (see for example [here](https://github.com/bambinos/bambi/blob/367caadfb4dacfc645f5532ee202536ee5334f16/bambi/terms/response.py#L47-L51)) This allows us to add more family-specific logic without having to modify how terms work.
* Terms require fewer objects to be initialized. Previously we were passing redundant information. Now we pass the formulae term and the prior. Responses also require the family.
* Removed manipulation of distribution parameters from the backend. This is delegated to the family. As result, we removed family-specific logic from the backend, and we can extend Bambi in a much richer way. See [here](https://github.com/bambinos/bambi/blob/367caadfb4dacfc645f5532ee202536ee5334f16/bambi/backend/terms.py#L221-L222) and [here](https://github.com/bambinos/bambi/blob/367caadfb4dacfc645f5532ee202536ee5334f16/bambi/backend/terms.py#L260-L263).
    * As a consequence, families gained optional methods. When these are present, they modify how the backend manipulates the parameters of the likelihood function or the linear predictor.

**Features/improvements**

* The `softmax` function from `aesara` is now called with `axis=-1`. There's a warning saying that one needs to call it explicitly in the future. I opened  #574 because of this. This PR fixes it, and I don't see the need to handle more general use cases.
* The `Likelihood` class now accepts custom distribution functions. Until now, what users could do was pass the name of a distribution that would be looked up in the PyMC namespace ([here](https://github.com/bambinos/bambi/blob/e1503d267265c50bd085456b920021e37e28f5cc/bambi/backend/utils.py#L5-L12) and [here](https://github.com/bambinos/bambi/blob/e1503d267265c50bd085456b920021e37e28f5cc/bambi/backend/terms.py#L271)). Now, you can create a `Likelihood` instance with a custom `dist` that will be used as the likelihood function. See [here](https://github.com/bambinos/bambi/blob/367caadfb4dacfc645f5532ee202536ee5334f16/bambi/backend/terms.py#L255-L258) and [here](https://github.com/bambinos/bambi/blob/367caadfb4dacfc645f5532ee202536ee5334f16/bambi/families/likelihood.py#L49).


tomicapretto:
I've run all the examples and they work well. Just updated a couple of them to fix some latex issues. Will merge after test suit finishes.

PR #589, opened 2022-11-23
Add common numpy transforms to extra_namespace

### LINKED ISSUE #587: Make common transformations available in Bambi
It would be good to make some transformations available directly in Bambi, so it's not needed to use the NumPy counterparts directly.

This way, we would be able to have `"y ~ log(x)"` instead of `"y ~ np.log(x)"`.

tomicapretto:
Adds the following transformations

* `log()`
* `log2()`
* `log10()`
* `exp()`
* `exp2()`
* `abs()`

Previously

```python
import numpy as np
model = bmb.Model("y ~ np.log(x)", data)
```

Now

```python
model = bmb.Model("y ~ log(x)", data)
```

Closes #587

### DIFF
--- a/bambi/utils.py
+++ b/bambi/utils.py
@@ -133,4 +133,13 @@ def censored(*args):
 
 censored.__metadata__ = {"kind": "censored"}
 
-extra_namespace = {"c": c, "censored": censored}
+extra_namespace = {
+    "c": c,
+    "censored": censored,
+    "log": np.log,
+    "log2": np.log2,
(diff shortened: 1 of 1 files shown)


tomicapretto:
I think the failing test is because https://github.com/pymc-devs/pymc/pull/6291 in PyMC, which is included in the v4.4 release of PyMC. Will fix it in another PR but need to think about it since Bambi requests PyMC > 4 and this would bring problems with PyMC < 4.4

PR #591, opened 2022-11-24
add AsymmetricLaplace family (Quantile Regression)

aloctavodia:
This is useful for performing quantile regression. To do so we have to fix the value of the kappa parameter to a suitable value. For example, to fit the 0.1 quantile we do the following:

```python
q = 0.1
κ = (q/(1-q))**0.5
Model("y ~ x", data, family="asymmetriclaplace", priors={"kappa":κ})
```

The following figure shows the fit to the 0.1 and 0.9 quantiles

quantile_regression

Having to transform from q to κ can be annoying, So I will add an alternative parametrization to the AssymetricLaplace in PyMC, so we can directly set the quantiles.

### DIFF
--- a/bambi/defaults/defaults.py
+++ b/bambi/defaults/defaults.py
@@ -2,6 +2,7 @@ from bambi.families import Likelihood
 from bambi.priors import Prior
 
 from bambi.families.univariate import (
+    AsymmetricLaplace,
     Bernoulli,
     Beta,
     Binomial,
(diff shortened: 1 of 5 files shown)


tomicapretto:
See this for why tests are failing https://github.com/bambinos/bambi/pull/589#issuecomment-1325451149

do you think we have to constrain either to PyMC >= 4.4 or PyMC < 4.4?

Another option is to determine the name of the argument at runtime, checking what's the PyMC version being used.

aloctavodia:
I think it is better to just restraint the pymc version

tomicapretto:
Do you plan to merge this only after the parametrization is implemented in PyMC, or do you plan to merge this and then open a new PR to update?

aloctavodia:
I was planning on doing a follow-up PR, so we don't need to wait. But I am open to alternatives

tomicapretto:
> I was planning on doing a follow-up PR, so we don't need to wait. But I am open to alternatives

I think that's fine. I don't think many people are going to use this feature right now before you open the follow-up PR. Let's do the following:

* Merge this
* Open PR to restrict the version of PyMC to >= 4.4 (I can do it now)
* Follow up PR on this issue

Sounds good?

aloctavodia:
sounds good to me. I will also add an example in a future PR.

Issue #590, opened 2022-11-23
`plot_cap()` does not work with transformations


tomicapretto:
```python
model = bmb.Model("y ~ f(x)", data)
idata = model.fit()

plot_cap(model, idata, covariates="x")
```

won't work.

tomicapretto:
Fixed in #594 and #596

PR #594, opened 2022-11-25
Add 'transforms' argument to plot_cap

### LINKED ISSUE #588: Add transforms argument to plot_cap
Sometimes we model transformed data and we want plots in the untransformed scale.

We could add a `transforms` argument to `plot_cap` where we allow to pass transformations that are applied either to variables on the x or the y axis.

I imagine something like

```python
plot_cap(model, idata, "log_x", transforms={"x": np.exp})
# or
plot_cap(model, idata, "log_x", transforms={"log_x": np.exp})
```

tomicapretto:
This PR adds an optional argument called `transforms` to `plot_cap()`. This is useful when the model is fitted on transformed data and we want the visualization in the original scale. See the example

```python
import bambi as bmb
import numpy as np
import pandas as pd

rng = np.random.default_rng(1234)
x = np.abs(rng.normal(size=100, scale=5)) + 5
y = np.exp(2 + 0.3 * x + rng.normal(size=100, scale=0.5))

data = pd.DataFrame({"x": x, "y": y})
data["log_x"] = np.log(x)
data["log_y"] = np.log(y)
model = bmb.Model("log_y ~ 1 + log_x", data)
idata = model.fit()

# Plot CAP on transformed scale. Result is a straight line
fig, ax = plt.subplots()
ax.scatter(data["log_x"], data["log_y"], color="C4", alpha=0.6);
plot_cap(model, idata, "log_x", ax=ax);
```

image

```python
# Plot CAP on the original scale, not necessarily a straight line
transforms = {"log_y": np.exp, "log_x": np.exp}
fig, ax = plt.subplots()
ax.scatter(data["x"], data["y"], color="C4", alpha=0.6);
plot_cap(model, idata, "log_x", transforms=transforms, ax=ax);
```
image

**Update** We can also use models that contain inline transformations. Notice the plot is created in the untransformed space. If we want the transformed space, we need to ask for it. While it may look counter-intuitive in the beginning, this is the most sensible approach I think because it works well when the predictor on the horizontal scale is included in more than a single term.

```python
fig, ax = plt.subplots()
ax.scatter(np.log(data["x"]), np.log(data["y"]), color="C4", alpha=0.6);
plot_cap(model, idata, "x", transforms={"x": np.log}, ax=ax);
```
image

```python
transforms = {"log(y)": np.exp}
fig, ax = plt.subplots()
ax.scatter(data["x"], data["y"], color="C4", alpha=0.6);
plot_cap(model, idata, "x", transforms=transforms, ax=ax);
```
image

Closes #588 and #590

### DIFF
--- a/bambi/plots/plot_cap.py
+++ b/bambi/plots/plot_cap.py
@@ -1,5 +1,6 @@
 # pylint: disable = protected-access
 # pylint: disable = too-many-function-args
+# pylint: disable = too-many-nested-blocks
 from statistics import mode
 
 import arviz as az
@@ -7,6 +8,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
 
+from formulae.terms.call import Call
 from matplotlib.lines import Line2D
 from matplotlib.patches import Patch
 from pandas.api.types import is_categorical_dtype, is_numeric_dtype, is_string_dtype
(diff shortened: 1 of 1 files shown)


tomicapretto:
I think this can be merged. Since it's not that trivial I would like a second pair of eyes @aloctavodia. If you're OK with it, let's merge.

canyon289:
If I understand correctly this PR has been replaced by https://github.com/bambinos/bambi/pull/596?

tomicapretto:
> If I understand correctly this PR has been replaced by #596?

This is myself still not knowing git very well haha! Being on this branch, I did `git checkout -b the_other_branch` because I wanted to have these changes applied as well. Let's say this PR is Part 1, and the other PR is meant to be Part 2. But because of how I used git, the other branch is Part 1 + Part 2. Do you know how one can do what I wanted to do in the beginning?

canyon289:
it looks like you did this correctly. when you merge this first PR then
only the part 2 changes remain

PR #596, opened 2022-11-26
Add panel covariates to `plot_cap()` and make it more flexible

tomicapretto:
~~This is still WIP, but I'm excited about this PR since it makes `plot_cap()` much more flexible~~

Now we can do something like 

```python
data = pd.read_csv("https://gist.githubusercontent.com/seankross/a412dfbd88b3db70b74b/raw/5f23f993cd87c283ce766e7ac6b329ee7cc2e1d1/mtcars.csv")
idata = model.fit(draws=500, target_accept=0.9, random_seed=1234)

fig, axes = plot_cap(
    model, 
    idata, 
    {"horizontal": "hp", "color": "cyl","panel": "gear"},
    use_hdi=False, 
    fig_kwargs={"figsize": (15, 5), "dpi": 120, "sharey": True}
);
```

image

~~All the previous examples still work. I still need to update the the internal function `_plot_cap_categoric()` and clean some details in `_plot_cap_numeric()`. In particular I need to handle legends using the figure and not the axes.~~

**Edit:** This is done. I decided to create a notebook to show the new features of `plot_cap()`. Basically, it still works for all the cases it used to work, but now it's more flexible. We can map groups to panels.

https://gist.github.com/tomicapretto/1388437357d8520ff5ff2fcee39b7d1b

### DIFF
--- a/bambi/plots/plot_cap.py
+++ b/bambi/plots/plot_cap.py
@@ -1,5 +1,6 @@
 # pylint: disable = protected-access
 # pylint: disable = too-many-function-args
+# pylint: disable = too-many-nested-blocks
 from statistics import mode
 
 import arviz as az
@@ -7,6 +8,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
 
+from formulae.terms.call import Call
 from matplotlib.lines import Line2D
 from matplotlib.patches import Patch
 from pandas.api.types import is_categorical_dtype, is_numeric_dtype, is_string_dtype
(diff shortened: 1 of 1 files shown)


bambi/plots/plot_cap.py
```diff
@@ -133,9 +150,12 @@ def plot_cap(model, idata, covariates, use_hdi=True, hdi_prob=None, legend=True,
         Changing the global variable ``az.rcParam["stats.hdi_prob"]`` affects this default.
     legend : bool, optional
         Whether to automatically include a legend in the plot. Defaults to ``True``.
+    transforms : dict, optional
+        Transformations that are applied to each of the variables being plotted. The keys are the
+        name of the variables, and the values are functions to be applied. Defaults to ``None``.
```
canyon289, line 155:
```suggestion
        name of the variables, and the values are functions to be applied. Defaults to `None`.
```

canyon289: Actually, just curious why are the double ticks needed? I see it in other docstrings too.

canyon289:
This is incredible, you basically rewrote this whole module, and now it can do so much more

canyon289 (approved):

bambi/plots/plot_cap.py
tomicapretto: If I remember correctly it's because it doesn't render properly if it only has a single backtick. Maybe something changed and now it works with a single backtick, but previously I remember trying a single backtick without working

canyon289:
you're probably right just ignore me

PR #598, opened 2022-11-28
Remove check to convert prior parameter values into arrays

tomicapretto:
I don't know why, but we had a check that made sure the values of the prior parameters were all numpy arrays. This was problematic when one tried to use configs such as `aesara.config.floatX = 'float32'` because `np.array()` would convert to int64 or float64 by default (this just happened to me).

I'm opening this mainly to see if our test suit still works. If that's the case, it means we don't need to convert single values to numpy arrays

### DIFF
--- a/bambi/priors/prior.py
+++ b/bambi/priors/prior.py
@@ -29,7 +29,6 @@ class Prior:
         kwargs : dict
             Optional keyword arguments to add to prior args.
         """
-        # The backend expect numpy arrays, so make sure all numeric values are represented as such.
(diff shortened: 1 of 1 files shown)


tomicapretto:
The tests failing are failing because they relied on these parameters being equal to numpy arrays (using the `.ndim` attribute). 

Two ideas

* Always convert to numpy array, considering the type in `aesara.config.floatX `
* Only convert when there's the assumption they're arrays
    *  I see this is "more efficient", but could introduce bugs or many manual checks

PR #603, opened 2022-12-03
Remove auto prior

### LINKED ISSUE #599: Remove the `automatic_priors` argument from `bmb.Model()`
It can only be set to a `"default"` value. I don't see a reason to keep it, since I don't think we're planning to add more types of default priors.

tjburch:
Removes automatic_priors from bmb.Model. Adds the note that automatics are based on rstanarm to the docstring for `priors`. Closes #599.


tomicapretto (approved):

tomicapretto:
Thanks a lot! :D

Issue #601, opened 2022-11-28
Don't allow `data` in `bmb.Model` to be a string anymore 


tomicapretto:
Currently, we can pass a string that is then passed to `pd.read_csv()`. It could have been an interesting feature at some point, but now the library is more mature I don't think we want to maintain these little tricks/details -- especially since `bmb.Model` is quite a large class.

PR #604, opened 2022-12-04
Remove string option for data input

tjburch:
Closes issue #601 by removing the string option for data. 

The type check is left in (line 124), since I think it's better feedback to the user rather than failing on a later TypeError trying to run pandas operations on a string.

The docstring is updated, I added a note that the column names must match the formula just so there's some explanation on expected format.

Code coverage will likely decrease since the associated test is also removed.


tomicapretto (approved):
Thanks again! I love well describe PRs

PR #605, opened 2022-12-04
Add build check function, on prior predictive and plot prior

### LINKED ISSUE #595: `Model.prior_predictive()` does not check if the model was built
If you build the model using a certain set of priors, and then update the priors, `Model.prior_predictive()` will work with the first set of priors, ignoring the updated version. This is because the underlying PyMC model still keeps the original priors. I think `prior_predictive()` should check whether `Model.built` is `True` as it happens with `Model.plot_priors()`.

tjburch:
Added a check that the model has been built in order to call `prior_predictive`. Since the same check is being called for this and `plot_priors`, I moved it out to a function that raises a `ValueError` if unbuilt.

Also added a single test that creates an unbuilt model and looks for that `ValueError`. 

Closes #595.

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -546,6 +546,14 @@ class Model:
             else:
                 raise KeyError(f"The name {group} is not a group in any group-specific term.")
 
+    def _check_built(self):
+        # Checks if model is built, raises ValueError if not
+        if not self.built:
+            raise ValueError(
+                "Model is not built yet! "
+                "Call .build() to build the model or .fit() to build and sample from the posterior."
+            )
+
     def plot_priors(
         self,
         draws=5000,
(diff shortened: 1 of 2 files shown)


tomicapretto:
https://github.com/bambinos/bambi/blob/7824d72e7125f6d4c96c57ad1a8494b4938c9bf5/bambi/models.py#L883-L887

The above is not exactly the same type of check but it's very similar. There will be cases where changing the priors won't change the graph, if you still use the same family of distributions for example, but there will be cases where it would change the graph. So I think we should consider the check here as well. What do you think?

tjburch:
Good catch.

That checks on a backend, which is set within the build function _only_, so it looks to me like in all cases where the model is built, the backend should exist, so doing the check on being built be identical to checking if the backend being built. So I'll just run the same check. If I'm not thinking of some edge case let me know.

tomicapretto:
Does it look like the failing test is due to numerical issues?

tjburch:
Yeah, looks like some spurious divergences. Thanks for triggering it again.

tomicapretto:
Thanks @tjburch :D

PR #609, opened 2022-12-08
Prepare release

tomicapretto:
* Update changelog
* Bump version to 0.9.2

I'm using `0.9.2` because I would like to make `0.10.0` the one we introduce distributional models. Is that ok?


aloctavodia:
Makes sense

PR #613, opened 2022-12-18
Make imports compatible with PyTensor/PyMC v5

### LINKED ISSUE #611: bambi depends on aesara, but pymc forked it to pytensor
Recently the `pymc` team announced that they forked `aesara` to `pytensor` [announcement here](https://www.pymc.io/blog/pytensor_announcement.html). They already moved ahead and ported their code to use `pytensor` instead of `aesara`.

Now when I install `bambi` in a clean environment, and try to import it without installing aesara as well, I get a `ModuleNotFoundError`.

Steps to reproduce:
```sh
pyenv virtualenv 3.11.0 bambi_test
pyenv activate bambi_test
pip install -U pip
pip install bambi ipython
```

Now trying to import `bambi` in the new environment (in ipython in this case) will fail with:
```python
In [1]: import bambi as bmb
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[1], line 1
----> 1 import bambi as bmb

File ~/.pyenv/versions/3.11.0/envs/bayes_test/lib/python3.11/site-packages/bambi/__init__.py:5
      1 import logging
      3 from pymc import math
----> 5 from .backend import PyMCModel
      6 from .data import clear_data_home, load_data
      7 from .families import Family, Likelihood, Link

File ~/.pyenv/versions/3.11.0/envs/bayes_test/lib/python3.11/site-packages/bambi/backend/__init__.py:1
----> 1 from .pymc import PyMCModel
      3 __all__ = ["PyMCModel"]

File ~/.pyenv/versions/3.11.0/envs/bayes_test/lib/python3.11/site-packages/bambi/backend/pymc.py:10
      7 import numpy as np
      8 import pymc as pm
---> 10 import aesara.tensor as at
     12 from bambi import version
     14 from bambi.backend.links import cloglog, identity, inverse_squared, logit, probit, arctan_2

ModuleNotFoundError: No module named 'aesara'
```

michaelosthege:
I changed the `aesara` imports to `try` PyTensor first, and fall back to `aesara` otherwise.

Mentions of `Aesara` were changed to `PyTensor`, and in `pymc.py` I had to change where `softmax` comes from because `aesara.tensor.nnet` was removed.

Closes #611

### DIFF
--- a/bambi/backend/links.py
+++ b/bambi/backend/links.py
@@ -1,13 +1,17 @@
 import numpy as np
-import aesara.tensor as at
+
+try:
+    import pytensor.tensor as pt
+except ModuleNotFoundError:
+    import aesara.tensor as pt
 
 
 def probit(x):
     """Probit function that ensures result is in (0, 1)"""
     eps = np.finfo(float).eps
-    result = 0.5 + 0.5 * at.erf(x / at.sqrt(2))
-    result = at.switch(at.eq(result, 0), eps, result)
-    result = at.switch(at.eq(result, 1), 1 - eps, result)
+    result = 0.5 + 0.5 * pt.erf(x / pt.sqrt(2))
+    result = pt.switch(pt.eq(result, 0), eps, result)
+    result = pt.switch(pt.eq(result, 1), 1 - eps, result)
 
     return result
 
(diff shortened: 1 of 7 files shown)


michaelosthege:
cc @tomicapretto @aloctavodia

bambi/backend/links.py
```diff
@@ -1,33 +1,37 @@
 import numpy as np
-import aesara.tensor as at
+
+try:
+    import pytensor.tensor as pt
+except ModuleNotFoundError:
```
canyon289, line 5: IMO we should just hard cut over. Theres no reason to stay on Aesara and it just means that we may get user bug reports from two backends rather than one.

michaelosthege: I would recommend to keep compatibility for at least one release and print a warning.

In the four packages I patched in the past few days I maintained compatibility with v4 because it simplifies updating my environments and Docker images.

In any case please take over from here - I'm not even using bambi myself 🙊

canyon289: Ill merge it. Hopefully no one gets angry at me

PR #614, opened 2022-12-18
Pytensor only

canyon289:
**Thank your for opening a PR!**

Before you proceed, please check the following notes.

* Make sure you describe the changes introduced in the PR.
* If this PR addresses an existing issue, please mention it here.
* Check our [Guidelines for Contributing](https://github.com/bambinos/bambi/blob/main/CONTRIBUTING.md).
* In particular, pay attention to the [pull request checklist](https://github.com/bambinos/bambi/blob/main/CONTRIBUTING.md#pull-request-checklist).
* The most important points are to
    + Install all the needed requirements (`requirements.txt`, `requirements-dev.txt`, etc.)
    + Make sure all test pass.
    + Make sure your code passes `black`.
    + Make sure your code passes `pylint`.

### DIFF
--- a/bambi/backend/links.py
+++ b/bambi/backend/links.py
@@ -1,9 +1,5 @@
 import numpy as np
-
-try:
-    import pytensor.tensor as pt
-except ModuleNotFoundError:
-    import aesara.tensor as pt
+import pytensor.tensor as pt
 
 
 def probit(x):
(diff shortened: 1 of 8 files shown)


tomicapretto:
I think the CI failing is a problem with numerical precision that we have seen previously, so that's not a big issue. 

But I think we should pin `pymc` to be at least version 5 in our `requirements.txt` file. Right now, it would install with PyMC 4.X.X, but that would install Aesara instead of PyTensor, and then we would have dependency issues.

canyon289:
Good point

tomicapretto:
Thanks! :D

PR #627, opened 2023-01-14
`plot_cap()` can have different model parameters as output

### LINKED ISSUE #622: Allow `plot_cap()` to work with any distribution parameter
Currently `plot_cap()` only works with the mean parameter in the family distribution. 
It would be great to be able to plot other parameters as well.

tomicapretto:
The function `plot_cap()` gains a new argument, `target`. This determines what parameter of the distribution is plotted as a function of the covariates. This is useful in distributional models. See the example

```python
import bambi as bmb
import numpy as np
import pandas as pd

rng = np.random.default_rng(121195)
N = 200
a, b = 0.5, 1.1
x = rng.uniform(-1.5, 1.5, N)
shape = np.exp(0.3 + x * 0.5 + rng.normal(scale=0.1, size=N))
y = rng.gamma(shape, np.exp(a + b * x) / shape, N)
data_gamma = pd.DataFrame({"x": x, "y": y})

formula = bmb.Formula("y ~ x", "alpha ~ x")
model = bmb.Model(formula, data_gamma, family="gamma")
idata = model.fit(random_seed=1234)
```

The default behavior still shows the mean parameter.

```python
plot_cap(model, idata, "x")
```

image

But we can have other targets as well 

```python
plot_cap(model, idata, "x", target="alpha")
```

image

Closes #622

### DIFF
--- a/tests/test_plots.py
+++ b/tests/test_plots.py
@@ -161,9 +161,9 @@ def test_multiple_outputs():
     y = rng.gamma(shape, np.exp(a + b * x) / shape, N)
     data_gamma = pd.DataFrame({"x": x, "y": y})
 
-    
+
     formula = Formula("y ~ x", "alpha ~ x")
     model = Model(formula, data_gamma, family="gamma")
-    idata = model.fit()
+    idata = model.fit(tune=100, draws=100, random_seed=1234)
     plot_cap(model, idata, "x")
     plot_cap(model, idata, "x", "alpha")
--- a/tests/test_plots.py
+++ b/tests/test_plots.py
@@ -5,7 +5,7 @@ import pandas as pd
 import matplotlib.pyplot as plt
 import pytest
 
-from bambi.models import Model
+from bambi.models import Model, Formula
 from bambi.plots import plot_cap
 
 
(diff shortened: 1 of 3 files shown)


canyon289 (approved):

tests/test_plots.py
```diff
@@ -150,3 +150,20 @@ def test_transforms(mtcars):
 
     transforms = {"mpg": np.log, "hp": np.log}
     plot_cap(model, idata, ["hp"], transforms=transforms)
+
+
+def test_multiple_outputs():
```
canyon289, line 155:
```suggestion
def test_multiple_outputs():
"""Test plot cap default and specified values for target argument"""
```

tests/test_plots.py
```diff
@@ -150,3 +150,20 @@ def test_transforms(mtcars):
...
+    x = rng.uniform(-1.5, 1.5, N)
+    shape = np.exp(0.3 + x * 0.5 + rng.normal(scale=0.1, size=N))
+    y = rng.gamma(shape, np.exp(a + b * x) / shape, N)
+    data_gamma = pd.DataFrame({"x": x, "y": y})
+
+
+    formula = Formula("y ~ x", "alpha ~ x")
+    model = Model(formula, data_gamma, family="gamma")
+    idata = model.fit(tune=100, draws=100, random_seed=1234)
+    plot_cap(model, idata, "x")
```
canyon289, line 168:
```suggestion
   # Test default target 
    plot_cap(model, idata, "x")
```

tests/test_plots.py
```diff
@@ -150,3 +150,20 @@ def test_transforms(mtcars):
...
+    plot_cap(model, idata, "x", "alpha")
```
canyon289, line 169:
```suggestion
    # Test user supplied target argument
    plot_cap(model, idata, "x", "alpha")
```

PR #655, opened 2023-03-22
imported bambi as bmb in tests

Amarjit0511:


tests/test_aliases.py
```diff
@@ -1,6 +1,8 @@
 import pytest
 
-from bambi import load_data, Formula, Model
+import bambi as bmb
```
canyon289, line 3: Can you explain what this line is doing in all tests? I undestand that its importing bambi as bmb, but how is the bambi top level being used within the test?

canyon289:
looks like were getting some testing failures! Try running the tests locally to debug. There also is a conflict with the main branch thatll need to be resolved as well

tomicapretto:
@Amarjit0511 thanks for the contribution!

Can you add a description to the PR? Also, if we do `import bambi as bmb` in the tests, we need to use the `bmb` namespace to have access to Bambi related classes and functions such as `bmb.Model`, `bmb.Formula`, `bmb.Prior`, etc.

Issue #644, opened 2023-02-19
Bambi FAQ could be more helpful


canyon289:
1. Remove boilerplate text
2. Add actual FAQs

https://bambinos.github.io/bambi/faq.html
![image](https://user-images.githubusercontent.com/7213793/219982182-14aec621-9db0-4461-9d1d-be2122cb28fb.png)

tomicapretto:
This was created because someone once offered to write content for the FAQ, but that never happened. And we never wrote it either. We could remove it if there's nobody who can write it

abuzarmahmood:
I'd be happy to contribute if I can get some pointers on what to include @tomicapretto @canyon289

tomicapretto:
Thanks @abuzarmahmood and @YuvrajSinghJadon for the initiative! And thanks Ravin for answering!
Following Ravin's comment, I think you could start with a list of questions that you have about the project (how to do X, or is Y supported, or I have Z problem what does it mean?) and we can help with the answers to those questions, while you create the appropriate markdown/notebook file in a PR.

abuzarmahmood:
Thank you for the pointers @tomicapretto and @canyon289. I've copied below a starter list of Q&A for your feedback. Full disclosure, I have only very recently started using Bambi so I apologize for inaccuracy/naivete in this list.

## General

- Why have a Bayesian regression library?
	- Bayesian modelling allows flexible (read 'bespoke') model specification and also provides an
		estimation of uncertainty in the model parameters. Both of these are wildly useful in 
		practice, in particular in a business context where the model is used to make decisions,
		and where a complex model may be needed to capture the underlying relationships. Further,
		Bayesian modelling allows graceful handling of small sample sizes by judicious use of
		prior distributions.

- I'm not familiar with Bayesian modelling, where can I learn more?
	- There are a number of excellent resources available online, including:
		- [Bambi examples](https://bambinos.github.io/bambi/examples.html)
		- [Bayesian Modeling and Computation](https://www.amazon.com/Bayesian-Modeling-Computation-Statistics-Science/dp/0124058884)
		- [Bayesian Data Analysis](https://www.amazon.com/Bayesian-Analysis-Chapman-Statistical-Science/dp/1439840954)
		- [Bayesian Methods for Hackers](https://github.com/CamDavidsonPilon/Probabilistic-Programming-and-Bayesian-Methods-for-Hackers)

- What is the difference between Bambi and PyMC3?
[...]

PR #657, opened 2023-03-23
FAQ page first draft

abuzarmahmood:
Adding questions for FAQ page per Issue #644

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -4,53 +4,19 @@ Frequently Asked Questions
 General Questions
 ===================
 
-Why have a Bayesian regression library?
+What is the difference between Bambi and PyMC?
 ----------------------------
-
-Bayesian modelling allows flexible (read 'bespoke') model specification and also provides an
-estimation of uncertainty in the model parameters. Both of these are wildly useful in 
-practice, in particular in a business context where the model is used to make decisions,
-and where a complex model may be needed to capture the underlying relationships. Further,
-Bayesian modelling allows graceful handling of small sample sizes by judicious use of
-prior distributions.
-
-
-And finally, see `this issue <https://github.com/bambinos/bambi/issues/389>`_.
-
-
-I'm not familiar with Bayesian modelling, where can I learn more?
-----------------------------
-
-There are a number of excellent resources available online, including:
-    * [Bambi examples](https://bambinos.github.io/bambi/examples.html)
-    * [Bayesian Modeling and Computation](https://bayesiancomputationbook.com/welcome.html)
-    * [Bayesian Data Analysis](https://www.amazon.com/Bayesian-Analysis-Chapman-Statistical-Science/dp/1439840954)
-    * [Bayesian Methods for Hackers](https://github.com/CamDavidsonPilon/Probabilistic-Programming-and-Bayesian-Methods-for-Hackers)
-
-What is the difference between Bambi and PyMC3?
-----------------------------
-    * Bambi is a wrapper around PyMC3, NumPyro, and BlackJax. It provides a simple interface for
-            specifying Bayesian models, and allows for easy inference using MCMC or 
-            variational inference.
-    * PyMC3 is a library for Bayesian modelling, and is the backend used by Bambi. It is a very
-            powerful library, but can be difficult to use for beginners. Bambi provides a simple
-            interface for specifying models, and allows for easy inference using MCMC or variational
-            inference.
+    * Bambi is a regression library built on top of PyMC. It provides a simple 
+        interface for specifying Bayesian models, and allows for easy inference using MCMC or 
+        variational inference.
+    * PyMC is a library for Bayesian modelling, and is the backend used by Bambi. It is a very
+            powerful library, but can be challenging to use for beginners. Bambi provides a simple
+            interface for specifying models, and allows for easy inference via MCMC or variational
+            inference using PyMC.
 
 
 Inference Questions
 ====================
-
-My sampler through errors/indicating divergences, what should I do?
-----------------------------
-    * Divergences are a common issue in Bayesian modelling, and are usually not a problem as long as
-            they are not prevalent. However, if you are seeing a lot of divergences, you may want 
-            to try 1) respecifying your model, 2) a different sampler.
-    * If the sampler fails, this is likely an issue with model specification. Make sure you are using
-            the correct priors for your model, and that you are not specifying a prior that is too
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -1,28 +1,93 @@
...
 
-Bambi not working on Linux
+I'm not familiar with Bayesian modelling, where can I learn more?
 ----------------------------
 
-* First
-* Second
+There are a number of excellent resources available online, including:
+    * [Bambi examples](https://bambinos.github.io/bambi/examples.html)
+    * [Bayesian Modeling and Computation](https://www.amazon.com/Bayesian-Modeling-Computation-Statistics-Science/dp/0124058884)
```
canyon289, line 26:
```suggestion
    * [Bayesian Modeling and Computation](https://bayesiancomputationbook.com/welcome.html)
```

canyon289 (commented):
This is look good to me. I'll let Tomas review as well!

Thank you for the PR

docs/faq.rst
```diff
@@ -1,28 +1,93 @@
...
+estimation of uncertainty in the model parameters. Both of these are wildly useful in 
+practice, in particular in a business context where the model is used to make decisions,
+and where a complex model may be needed to capture the underlying relationships. Further,
+Bayesian modelling allows graceful handling of small sample sizes by judicious use of
+prior distributions.
 
-* First
-* Second
 
 And finally, see `this issue <https://github.com/bambinos/bambi/issues/389>`_.
```
tomicapretto, line 18: Can we remove this line?

tomicapretto:
Hi @abuzarmahmood thanks a lot for this contribution! 

I think the following questions fit really well in the FAQ

* What is the difference between Bambi and PyMC3?
* What sampling methods are available?
* Can infernece in Bambi be sped up using GPUs/TPUs?
* My data has a non-normal distributions, can I still use Bambi?
    * Although I would phrase this in a different way.
    * I want to use a non-normal likelihood model, can I still use Bambi?
* How do I find out what priors are available?
* Does bambi come with pre-specified regression models?


I'm not so sure about these other questions

* Why have a Bayesian regression library?
* I'm not familiar with Bayesian modelling, where can I learn more?
* My sampler through errors/indicating divergences, what should I do?

Those are more conceptual/philosophical and modeling-related questions. I don't think we should try to answer them in the FAQ, since I thing the FAQ is for questions about the library.

docs/faq.rst
```diff
@@ -1,28 +1,93 @@
...
-* Second
+There are a number of excellent resources available online, including:
+    * [Bambi examples](https://bambinos.github.io/bambi/examples.html)
+    * [Bayesian Modeling and Computation](https://www.amazon.com/Bayesian-Modeling-Computation-Statistics-Science/dp/0124058884)
+    * [Bayesian Data Analysis](https://www.amazon.com/Bayesian-Analysis-Chapman-Statistical-Science/dp/1439840954)
+    * [Bayesian Methods for Hackers](https://github.com/CamDavidsonPilon/Probabilistic-Programming-and-Bayesian-Methods-for-Hackers)
+
+What is the difference between Bambi and PyMC3?
+----------------------------
+    * Bambi is a wrapper around PyMC3, NumPyro, and BlackJax. It provides a simple interface for
```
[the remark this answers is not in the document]
abuzarmahmood: Yes, calling it a wrapper is incorrect. Thank you for the correction. I also took out mention of numpyro and blackjax here because they are mentioned as backends in multiple other questions.

abuzarmahmood:
> Hi @abuzarmahmood thanks a lot for this contribution!
> 
> I think the following questions fit really well in the FAQ
> 
> * What is the difference between Bambi and PyMC3?
> * What sampling methods are available?
> * Can infernece in Bambi be sped up using GPUs/TPUs?
> * My data has a non-normal distributions, can I still use Bambi?
>   
>   * Although I would phrase this in a different way.
>   * I want to use a non-normal likelihood model, can I still use Bambi?
> * How do I find out what priors are available?
> * Does bambi come with pre-specified regression models?
> 
> I'm not so sure about these other questions
> 
> * Why have a Bayesian regression library?
> * I'm not familiar with Bayesian modelling, where can I learn more?
> * My sampler through errors/indicating divergences, what should I do?
> 
> Those are more conceptual/philosophical and modeling-related questions. I don't think we should try to answer them in the FAQ, since I think the FAQ is for questions about the library.

That makes sense. I've updated it to include the questions you suggested.

docs/faq.rst
```diff
@@ -1,28 +1,51 @@
...
+----------------------------
+Yes, Bambi supports a wide range of distributions which can be specified using the "family"
+argument to the "Model". You can find examples of how to specify these distributions 
+in the [Bambi examples](https://bambinos.github.io/bambi/examples.html).
 
-Miscellaneous how to
-====================
+How do I find out what priors are available?
+----------------------------
+You can find a list of available priors in the [Bambi Models API](https://bambinos.github.io/bambi/api_reference.html#module-bambi.models) under the "family" argument.
```
tomicapretto, line 45: The family argument tells you what are the available families. But for priors, you can use any PyMC distribution that is valid as a prior.

tomicapretto:
> @tomicapretto Got it! That was my bad for confusing families with priors. I've update the response to the priors question ([67f52fb](https://github.com/bambinos/bambi/commit/67f52fb6dfb0a635f066409177f8df554a6aa15a)).

Last nit: It's PyMC instead of PyMC3 (we're using PyMC version 5, but we write just PyMC)

abuzarmahmood:
> > @tomicapretto Got it! That was my bad for confusing families with priors. I've update the response to the priors question ([67f52fb](https://github.com/bambinos/bambi/commit/67f52fb6dfb0a635f066409177f8df554a6aa15a)).
> 
> Last nit: It's PyMC instead of PyMC3 (we're using PyMC version 5, but we write just PyMC)

Sounds good. Latest commit uses "PyMC" instead of "PyMC3".

docs/faq.rst
abuzarmahmood: Updated as of 67f52fb6df

tomicapretto:
The test failure is known. It's due to an update of the variable names in PyMC/PyTensor, which uses full names now. It's already updated in #668. I'm merging.

Thanks @abuzarmahmood !

abuzarmahmood:
Lovely. Thank you @tomicapretto @canyon289 for your guidance

PR #659, opened 2023-03-27
Add compound families

tomicapretto:
This adds two new families

* [x] The `"beta_binomial"` family which uses the `BetaBinomial` distribution from PyMC.
* [x] The `"dirichlet_multinomial"` family which uses the `DirichletMultinomial` distribution from PyMC.

Also, I'm adding a `BinomialBaseFamily` which implements the common transformations needed for posterior predictive sampling and the transformation of backend kwargs.

~~For the dirichlet_multinomial I first need to improve the implementation of posterior predictive of the multinomial family~~ Done!

### DIFF
--- a/bambi/families/univariate.py
+++ b/bambi/families/univariate.py
@@ -55,6 +55,11 @@ class Bernoulli(UnivariateFamily):
 
 
 class Beta(UnivariateFamily):
+    """Beta Family
+
+    It uses the mean (mu) and sample size (kappa) parametrization of the Beta distribution.
+    """
+
     SUPPORTED_LINKS = {"mu": ["logit", "probit", "cloglog"], "kappa": ["log"]}
 
     @staticmethod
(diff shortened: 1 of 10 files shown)


bambi/families/univariate.py
```diff
@@ -46,27 +66,22 @@ def transform_backend_kwargs(kwargs):
...
-        if data is None:
-            trials = model.response_component.response_term.data[:, 1]
-        else:
-            trials = model.response_component.design.response.evaluate_new_data(data)
-
-        response_name = get_aliased_name(model.response_component.response_term)
-        mean = posterior[response_name + "_mean"]
-        return xr.apply_ufunc(np.random.binomial, trials.squeeze(), mean)
+class BetaBinomial(BinomialBaseFamily):
+    # We use the 'mu' and 'kappa' parametrization of the Beta
```
canyon289, line 70: Might as well make this a proper class docstring. It would be a good addition for everything else in here as well

bambi/families/univariate.py
```diff
@@ -46,27 +66,22 @@ def transform_backend_kwargs(kwargs):
...
+    SUPPORTED_LINKS = {"mu": ["logit", "probit", "cloglog"], "kappa": ["log"]}
 
     @staticmethod
     def transform_backend_kwargs(kwargs):
-        observed = kwargs.pop("observed")
-        kwargs["observed"] = observed[:, 0].squeeze()
-        kwargs["n"] = observed[:, 1].squeeze()
-        return kwargs
+        # Beta specific parameters
```
canyon289, line 75: Same with this docstring

bambi/families/univariate.py
```diff
@@ -150,31 +165,12 @@ class Wald(UnivariateFamily):
     SUPPORTED_LINKS = {"mu": ["inverse", "inverse_squared", "identity", "log"], "lam": ["log"]}
 
 
-class ZeroInflatedBinomial(UnivariateFamily):
+class ZeroInflatedBinomial(BinomialBaseFamily):
```
canyon289, line 168: Nice

PR #660, opened 2023-03-28
import bambi and bmb

Amarjit0511:
Made required changes after importing bambi as bmb

### DIFF
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -42,30 +42,30 @@ def no_remote_data(monkeypatch, tmpdir):
 def test_clear_data_home():
     resource = DATASETS["test_remote"]
     assert os.path.exists(resource.filename)
-    clear_data_home(data_home=os.path.dirname(resource.filename))
+    bmb.clear_data_home(data_home=os.path.dirname(resource.filename))
     assert not os.path.exists(resource.filename)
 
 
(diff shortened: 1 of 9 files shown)


tests/test_data.py
```diff
@@ -5,8 +5,10 @@
 import pandas as pd
 import pytest
 
-from bambi import clear_data_home, load_data
-from bambi.data.datasets import DATASETS, FileMetadata
+import bambi as bmb
+
+from bmb import clear_data_home, load_data
```
tomicapretto, line 10: same as above, not needed to import these now

tomicapretto (changes requested):
Thanks for the updates!

There are a couple of changes that need to be made.

Amarjit0511:
> Ahhh. I see we are doing some duplicated work @Amarjit0511. I just realised this since this PR didn't link the issue. I'll admit, that's also part my fault for not searching through the open PR's - lesson learned. However, I have just completed my PR #662 with all tests passing for this issue. Therefore, I think we can close this PR. Cheers!


No, it's my fault actually for not linking the issue in the PR. Actually I am very new to these contributions on GitHub and I often end up making silly mistakes. 
Btw I am happy to know that your PR #662 passed all the tests and helped solve this issue.

PR #662, opened 2023-03-29
import bambi as bmb in tests

### LINKED ISSUE #653: Use `import bambi as bmb` in tests
I don't know why we are importing specific objects only. Tests don't look like regular Bambi code because of this.

GStechschulte:
This draft PR resolves #653 to avoid importing specific objects only in the test files of the `test/` directory  . For directories treated as packages in the [__init__.py](https://github.com/bambinos/bambi/blob/main/bambi/__init__.py) I use `import bambi as bmb`. Directories not treated as packages, I import the submodule individually. For example, `from bambi.terms import CommonTerm`

Tests are completed and passed locally:

- [x] Make sure all test pass.
- [x] Make sure your code passes `black`.
- [x] Make sure your code passes `pylint`.

### DIFF
--- a/docs/notebooks/getting_started.ipynb
+++ b/docs/notebooks/getting_started.ipynb
@@ -662,7 +662,6 @@
    ]
   },
   {
-   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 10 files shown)


tomicapretto:
Thanks Gabriel! :D

PR #666, opened 2023-04-07
Fix name of arviz traceplot function

mbjoseph:
This PR fixes a small typo for the traceplot method from arviz, which is [`plot_trace`](https://python.arviz.org/en/latest/api/generated/arviz.plot_trace.html) rather than [`plot_traces`](https://github.com/arviz-devs/arviz/search?q=plot_traces).


tomicapretto:
Good catch! Thanks @mbjoseph for the contribution!

Issue #623, opened 2023-01-06
Allow `plot_cap()` to show predictions at the observation level


tomicapretto:
`plot_cap()` show the adjusted mean. After #622 it will be able to show other adjusted parameters. But it doesn't show the adjusted posterior predictive distribution, and it would be really nice to have that.

tomicapretto:
Closed with #668

PR #668, opened 2023-04-16
Allow `plot_cap()` to show predictions at the observation level

GStechschulte:
This draft PR addresses issue #623 to allow `plot_cap()` to show predictions at the observation level, i.e., to use the adjusted posterior predictive distribution. 

This is achieved by adding an additional boolean argument `pps` (posterior predictive sample) to the `plot_cap()` function. If true, then the adjusted posterior predictive distribution is plotted for the response variable by using `model.predict(..., kind="pps")`, else the adjusted posterior is used for plotting, i.e., `model.predict(..., kind="mean")`. 

Below are some examples:

```python3
import arviz as az
import bambi as bmb
import numpy as np
import pandas as pd

data = pd.read_csv("data/mtcars.csv")

idata = model.fit(draws=1000, target_accept=0.95, random_seed=1234)
```

```python3
fig, ax = plt.subplots(figsize=(7, 3), dpi=120)
plot_cap(model, idata, "hp", pps=False, ax=ax);
```
output1

```python3
# posterior predictive distribution (forward sampling)
# i.e., plot_cap at observation level
fig, ax = plt.subplots(figsize=(7, 3), dpi=120)
plot_cap(model, idata, "hp", pps=True, ax=ax);
```
output2

```python3
fig, ax = plt.subplots(figsize=(7, 3), dpi=120)
plot_cap(model, idata, ["hp", "wt"], pps=False, ax=ax);
```
output3

```python3
fig, ax = plt.subplots(figsize=(7, 3), dpi=120)
plot_cap(model, idata, ["hp", "wt"], pps=True, ax=ax);
```
output4

This is a work in progress and I still need to test on other examples, add tests, and perform linting. However, if anyone has any initial comments or criticisms, let me know. Thanks!


bambi/plots/plot_cap.py
```diff
@@ -155,6 +156,8 @@ def plot_cap(
         are the names of the variables.
     target : str
         Which model parameter to plot. Defaults to 'mean'.
```
tomicapretto, line 158: I would add this only has effect when `pps` is `False`

bambi/plots/plot_cap.py
```diff
@@ -194,22 +197,30 @@ def plot_cap(
     cap_data = create_cap_data(model, covariates)
     idata = model.predict(idata, data=cap_data, inplace=False)
 
-    if hdi_prob is None:
-        hdi_prob = az.rcParams["stats.hdi_prob"]
-
-    if not 0 < hdi_prob < 1:
-        raise ValueError(f"'hdi_prob' must be greater than 0 and smaller than 1. It is {hdi_prob}.")
-
```
tomicapretto, line 202: This was here so we avoid computations when we raise an error.  Now I think it should be even before 

```python
cap_data = create_cap_data(model, covariates)
```

bambi/plots/plot_cap.py
```diff
@@ -194,22 +197,30 @@ def plot_cap(
     cap_data = create_cap_data(model, covariates)
     idata = model.predict(idata, data=cap_data, inplace=False)
```
tomicapretto, line 198: I think this line could be removed because it now depends on `pps`, right?

bambi/plots/plot_cap.py
```diff
@@ -194,22 +197,30 @@ def plot_cap(
...
         transforms = {}
 
     response_name = get_aliased_name(model.response_component.response_term)
     response_transform = transforms.get(response_name, identity)
 
-    y_hat = response_transform(idata.posterior[f"{response_name}_{target}"])
-    y_hat_mean = y_hat.mean(("chain", "draw"))
+    if pps:
+        idata = model.predict(idata, data=cap_data, inplace=False, kind="pps")
+        y_hat = response_transform(idata.posterior_predictive[f"{response_name}"])
```
tomicapretto, line 208: the f string is not needed here

bambi/plots/plot_cap.py
```diff
@@ -244,7 +255,7 @@ def plot_cap(
     for ax in axes.ravel():  # pylint: disable = redefined-argument-from-local
         ax.set(xlabel=main, ylabel=ylabel)
 
-    return fig, axes
+    return (fig, axes), idata, cap_data
```
tomicapretto, line 258: I'm not so sure about returning `idata` and `cap_data` as well. I see it would be nice because it allows you to do more things with that, but i think it could be surprising for users who are not familiar with these data structures. So perhaps we can have an optional `return_data` argument?

docs/notebooks/data/mtcars.csv
tomicapretto, line 1: Later I would like to upload this to figshare and make it available through `bmb.load_data("mtcars")` so we don't keep the data in the repo (although we have some data in the repo). I can help with this if you want :)

tomicapretto (changes requested):
Thanks a lot for the nice PR! I added some suggestions for changes. Let me know if you have any questions :)

bambi/plots/plot_cap.py
GStechschulte: Ahhh, yes I forgot to delete this code. I was returning the data objects to inspect myself. Thanks for that!

docs/notebooks/data/mtcars.csv
GStechschulte: Good idea. Would we also like to move existing data in the repo also to figshare? I am not too familiar with figshare, so I will do some research first and then communicate back here.

bambi/plots/plot_cap.py
GStechschulte: I was also thinking about this and agree. I will also add error handling so that if users do attempt to use `pps=True` and `target` is not None, it will raise an informative error.

docs/notebooks/data/mtcars.csv
tomicapretto: Don't worry about it, I can do it for you and update the PR appropriately

bambi/plots/plot_cap.py
tomicapretto: Good! But I don't think it should be an error, because it's not something we cannot recover from. A warning would be more appropriate. However, the default is `target="mean"` so it would cause the warning unless users pass `target=None` explicitly... So finally, I wouldn't make it either an error or a warning. Just the note in the docstring is fine.

tomicapretto (approved):

tomicapretto:
@GStechschulte I just updated the PR with the mtcars dataset, now you can do `bmb.load_data("mtcars")`. I also bumped PyMC to 5.3.0. Let me know if you want me to review the example :)

GStechschulte:
> @GStechschulte I just updated the PR with the mtcars dataset, now you can do `bmb.load_data("mtcars")`. I also bumped PyMC to 5.3.0. Let me know if you want me to review the example :)

Great, thanks! I am going to perform a few more examples (other than the mtcars dataset and model) first, and then you can review the example(s). I plan to be finished with this (and move from a draft PR to PR) by the end of this weekend 23.04.

As far as adding tests in the `test_plots.py` of the `tests/` directory goes, what do you think? I don't have much experience with writing tests, but I think with the added functionality, it would be useful to add some. Looking at the code in `test_plots.py`, I have a pretty good idea of what the tests should be. Or...I can ask CoPilot ;)

tomicapretto:
Sounds great! Thanks for the extra examples.

It's not trivial to test plotting functions. That's why the code in `plot_tests.py` only checks things work, without checking the actual plot. If you can do something close to that, that would be awesome.


See the test code also uses mtcars, but it's loaded from the tests module. You could remove that and load it with `bmb.load_data()` as well :)

GStechschulte:
> Sounds great! Thanks for the extra examples.
> 
> It's not trivial to test plotting functions. That's why the code in `plot_tests.py` only checks things work, without checking the actual plot. If you can do something close to that, that would be awesome.
> 
> See the test code also uses mtcars, but it's loaded from the tests module. You could remove that and load it with `bmb.load_data()` as well :)

Hey @tomicapretto I have been testing this functionality with a variety of different models. The examples can be found in this [Gist](https://gist.github.com/GStechschulte/345a700cb5b9fb702878c5c98fe222fe). Most of them look good besides categorical regression (which I opened an issue), and models with binary responses that are represented by the proportion alias. 

The problem with `p(y, n) ~ x` is that when using `pps=True`, to create the `cap_data`, both `n` and `x` need to be passed in order to compute the proportion `y / n` of the outcome `y`. I provide an example of what a general solution to this problem looks like in the _Regression for binary responses_ section.

tomicapretto:
@GStechschulte thanks for the thorough notebook. It's a great piece of work! Here my answers

* For binary responses the posterior predictive distribution is in a different scale than the mean parameter. The former is in the binary scale (set {0, 1}) and the latter is on the probability scale (interval (0, 1)). So a barchart or point and line plot would be more appropriate. **However** I think it's not needed to handle that case in this PR. I would leave it as it is, or add an error when the family is Bernoulli/Binomial.
* For categorical responses:
    * We could plot the mean. I can help to fix the fill between.
    * Similar issues than before with the draws from the posterior predictive distribution. It doesn't need to be solved now.

What do you think?

tomicapretto:
@GStechschulte Yes, you can do something like this:

```python
@pytest.mark.parametrize("pps", [False, True])
def test_basic(mtcars, pps):
    model, idata = mtcars

    # Using dictionary
    # Horizontal variable is numeric
    plot_cap(model, idata, {"horizontal": "hp"}, pps=pps)

    # Horizontal variable is categorical
    plot_cap(model, idata, {"horizontal": "gear"}, pps=pps)

    # Using list
    plot_cap(model, idata, ["hp"], pps=pps)
    plot_cap(model, idata, ["gear"], pps=pps)
```

Havent' tested locally but I think that's it. Let me know if you need help :)

tomicapretto:
@GStechschulte looks like it can be merged? :D

GStechschulte:
> @GStechschulte looks like it can be merged? :D

Indeed, it can be merged :) thanks for your code reviews and quick feedback!

Issue #667, opened 2023-04-16
Add explicit `plot_cap` example in docs


GStechschulte:
After looking for an example of the `plot_cap` function, I realised there is no explicit example in the docs that describes what this function is doing and its capabilities. The only examples that exist are from related `plot_cap` [closed PRs](https://github.com/bambinos/bambi/pulls?q=is%3Apr+is%3Aclosed+plot_cap), this [Gist](https://gist.github.com/tomicapretto/1388437357d8520ff5ff2fcee39b7d1b), and the Gaussian Process (HSGP) [example](https://bambinos.github.io/bambi/notebooks/hsgp_1d.html).

This is unfortunate because the work by @tomicapretto on this is great, and the functionality provides a lot of value-add for the users. Thereby, users should only have to look into the docs to understand the usage of this function. I think we could compile the examples from the closed PRs and the Gist linked above into one notebook with explanations, etc. and add this as an example to the docs.

I would be happy to take this issue on. Any suggestions?

PR #670, opened 2023-04-24
plot_cap example for docs

GStechschulte:
This draft PR addresses #667 by adding an explicit `plot_cap` notebook in the `docs/notebooks/` directory. The objective of this notebook is to show how to use, and the capabilities, of the `plot_cap` function. Additionally, there will be sections on why such tooling is needed for GLMs (generally speaking), and what is happening on the backend to produce the `plot_cap` plots. 

This is a work in progress, and at the moment, I only have a draft outline for the doc example. Nonetheless, examples will be taken from this [Gist](https://gist.github.com/GStechschulte/345a700cb5b9fb702878c5c98fe222fe).


tomicapretto:
@GStechschulte feel free to let me know when you need a review :)

docs/notebooks/plot_cap.ipynb
```diff
@@ -0,0 +1,796 @@
+{
```
tomicapretto, line 1: <p>Typo: it says g^-1(n) instead of g^-1(eta) </p><p> </p><blockquote>The distribution can also depend on a dispersion parameter&nbsp;...</blockquote><p>While it's correct that Bambi can handle this, the resulting model does not fall under the GLM family -- we call these distributional models.</p><p> </p><p><s>What's more, for these models to be GLMs the conditional distribution of the response must be a member of the exponential family of distributions (however we usually omit this detail). </s> Just saw you also added it, lol.</p><p> </p><blockquote>the link function often bounds the mean outcome between some range or to make it more plausible when used as the parameter in some outcome distribution</blockquote><p>Not sure about the "or to make it more plausible...", maybe you were modifying this and forgot to remove something there?</p><p> </p><p>What do you mean with the following?</p><p> </p><blockquote>there are two ways to interpret the model parameters of a GLM: (1) on the link scale, or (2) on the response scale</blockquote><p> </p><p>I'm not sure to which parameters you refer to (if it's the conditional mean or it's regression coefficients). </p><p> </p><p>The way I see it, you usually have three quantities you can plot with plot_cap</p><p> </p><ul><li>The linear predictor eta.  This always ranges between -inf and inf no matter the nature of the response.</li><li>The mean. Which is mu = g^-1(eta). This is <em>sometimes</em> in the scale of the response variable (e.g. Normal or Poisson likelihood), but other times it's not in the scale of the response (e.g. Binomial likelihood).</li><li>Response values. This are values in the same scale than the response because, well, they are draws from the conditional response distribution for given values of the covariates.</li></ul><p> </p><p>But maybe you tried to say the same thing and I'm just not understanding. Let's open the discussion. </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/670/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

tomicapretto, line 1: <p>Ha, I should read it completely before writing because above I basically repeated what you say here. I still have some comments thought.</p><p> </p><p>Regarding the first sentence. I think the "adjusted" in "adjusted predictions" comes from the fact that you're "adjusting for covariates" when computing the predictions. This is also seen as "controlling for the covariates" but the latter is a bit more misleading because it may be interpreted that you controlled the covariates and that's not always the case. </p><p> </p><p>I think this is what you say in the last sentence.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/670/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

tomicapretto, line 1: <p>"how a model parameter varies..." I think it would be slightly better to say "how a parameter of the (conditional) response distribution varies...". It's not incorrect to say "model parameter", but it can make people think of other model parameters such as regression coefficients. </p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/670/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

tomicapretto, line 1: <p>I would add just one or two sentences about the dataset. Something along the lines of "this is the famous mtcars dataset which contains..." bla bla bla. </p><p> </p><p>Or you can simply copy the description</p><p> </p><p>The data was extracted from the 1974 Motor Trend US magazine, and comprises fuel consumption and 10</p><p>aspects of automobile design and performance for 32 automobiles (1973--74 models).&nbsp;</p><p>The following is a description of the variables:</p><p> </p><p>* mpg: Miles/(US) gallon</p><p>* cyl: Number of cylinders</p><p>* disp: Displacement (cu.in.)</p><p>* hp: Gross horsepower</p><p>* drat: Rear axle ratio</p><p>* wt: Weight (1000 lbs)</p><p>* qsec: 1/4 mile time</p><p>* vs: Engine (0 = V-shaped, 1 = straight)</p><p>* am: Transmission (0 = automatic, 1 = manual)</p><p>* gear: Number of forward gear</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/670/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

tomicapretto, line 1: <p>The description is very well done in general. The only nitpick is that this is not showing the response, but the mean of the response. Since both are in the same scale the difference is subtle.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/670/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

tomicapretto, line 1: <p>Fantastic explanation. </p><p> </p><p>I'm also considering the comment below the figure. I would add in one case we're visualizing the posterior of a parameter (pps=False) and in the other case we're visualizing the posterior predictive which is over the response variable. As you say, it incorporates  an additional source of uncertainty. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/670/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

tomicapretto, line 1: <p>Maybe a sentence explaining how the list is mapped onto the dictionary you described above. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/670/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

tomicapretto, line 1: <p>nit: passing a parameter into target -&gt; passing a parameter into <code>target</code></p><p> </p><p> <code>target</code>  doesn't matter when <code>pps</code> is True because <code>pps=True</code> means to plot the posterior predictive distribution and that can only refer to the outcome variable (instead of any of the parameters of the response distribution).</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/670/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='34'/>

tomicapretto:
@GStechschulte the example is in very very good shape. I love the amount of detail. We just need to round out some explanations and perhaps add a couple more. But in general this is great and almost ready to be merged. Thanks a lot!

docs/notebooks/plot_cap.ipynb
GStechschulte: <blockquote>Regarding the first sentence. I think the "adjusted" in "adjusted predictions" comes from the fact that you're "adjusting for covariates" when computing the predictions</blockquote><p>This is also my interpretation also. I have adjusted the wording to include  "adjusted predictions refers to the outcome predicted by a fitted model on a specified scale for a given combination of values of the predictor variables, such as their observed values, their means, or some user specified grid of values". Regarding the "conditional" part, I now explicitly state it is the "specific predictor(s) we would like to condition on...."</p>

GStechschulte: <p>This is a good point. Model parameter is ambiguous. </p><p> </p>

GStechschulte: <p>Thanks! And I have added this except instead of saying "posterior of a parameter" to "mean outcome parameter" to avoid the unambiguous meaning of "parameter". This way, we stay in line with the docstring of plot_cap (target="mean") and the GLM explanation in the beginning sections.</p>

GStechschulte: <p>Yeah, I like the further clarification here.</p>

GStechschulte:
@tomicapretto Thank you and thanks for the quick and detailed feedback. You can find my comments to your reviews in the [ReviewNB](https://app.reviewnb.com/bambinos/bambi/pull/670/discussion/) (they haven't been mapped back to this PR I noticed).

docs/notebooks/plot_cap.ipynb
```diff
@@ -0,0 +1,856 @@
...
```
tomicapretto, line 1: <p>About the third point: I would say Y | mu, theta ~ D(mu, theta) or something like that. "mu" is the mean parameter and "theta" is (possibly) a vector that contains all the other "nuisance" parameters. </p><p> </p><p>As it is written now the second and third points both talk about the mean.</p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/670/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

tomicapretto, line 1: <p>"Adjusted predictions refers to the outcome predicted by a fitted model on a specified scale for a given combination of values of the predictor variables, such as their observed values, their means, or some user specified grid of values" as complex as it reads... it's exactly what it is. So good work here!</p><p> </p><p>Edited to add a question. What do you mean with "link scale"? Is it the scale of the linear predictor, or the scale of mu = g^-1(eta)? I think you mean the scale of "eta" because of the logistic regression example, but I want to confirm.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/670/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

GStechschulte: <p>Yes, good point. I used the following notation: Y ~ D(mu, theta). </p>

GStechschulte: <p>Thanks, and yeah it is a "mouth full" to say. Yes, I mean the scale of eta because of the logistic regression example. </p><p> </p>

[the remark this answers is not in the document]
GStechschulte: <p>Good point. I will explicitly state "the Gaussian linear model uses an identity link function that results in no transformation of the linear predictor to the mean of the outcome variable" so the user knows the use of the identity link in the case of the Gaussian linear model has this property.</p>

GStechschulte:
Again, thanks for the detailed comments @tomicapretto. I have incorporated these points into the latest commit.

tomicapretto:
@GStechschulte it's good to be merged, thanks for the great work! 

But there's a final to do :sweat_smile:.  See the conflict with the notebook file. Do you know how to fix it?

GStechschulte:
Thanks @tomicapretto! Looks like we are good to go now 😄

PR #684, opened 2023-06-12
Plot comparisons

GStechschulte:
This draft PR introduces `plot_comparisons`, a function for comparing the predictions made by a fitted model for different contrasts while holding all other covariates constant or at a user defined value. Inspiration was taken from the great [marginaleffects](https://vincentarelbundock.github.io/marginaleffects/) R package. 

At a high level, `plot_comparisons` allows the modeller to define a contrast `contrast_predictor` and the covariate values to condition on `conditional`. If a user does not pass specific values (for either the contrast_predictor or conditional), then a default grid of values is computed. Thus, the comparison in predictions allows a modeller to "see through the eyes of the model", i.e., the comparison on the scale of the outcome. The comparison of predictions is computed using all chains and draws of the posterior.

Currently, `plot_comparisons` only allows a user to compare the predictions for 1 contrast level, i.e., how does the probability of survival change if a person moves from 1st to 3rd class given `Age = 50` and `Sex = [0, 1]`.

```python3
fig, ax = plot_comparison(
    model=titanic_model,
    idata=titanic_idata,
    contrast_predictor={"PClass": [1, 3]},
    conditional={"Age": [50], "SexCode": [0, 1]}
)
```
image

In the above example, the user defined a value for each covariate. However, default values are computed if the user does not provide any for `conditional`:

```python3
fig, ax = plot_comparison(
    model=titanic_model,
    idata=titanic_idata,
    contrast_predictor={"PClass": [1, 3]},
    conditional=["Age", "SexCode"]
)
```
image

Another example of default values being computed for a categorical contrast predictor and numerical conditional covariate:

```python3
fig, ax, comparisons_df, contrast_df, idata = plot_comparison(
    model=fish_model,
    idata=fish_idata,
    contrast_predictor="livebait",
    conditional="persons"
) 
```
image

These examples, and further explanations can be found in the following [notebook](https://github.com/GStechschulte/bambi/blob/plot-comparisons/docs/notebooks/plot_comparisons.ipynb).


GStechschulte:
There is still some work to do here before moving to a normal PR. Namely:
- [ ] shape handling for comparisons where a user passes > 1 level
- [ ] allow plotting of subplots (panels)
- [ ] docstrings and type hints
- [ ] optional return of `contrast_df` (a dataframe containing descriptive statistics of the contrast comparison)
- [ ] refactor `plot_cap` code so it works
- [ ] plot comparisons of other model parameters?

GStechschulte:
To plot subplots, I have added an additional argument `subplot_kwargs` where the user can specify the `main`, `group`, and `panel` covariates. I believe this additional arg. is needed because with `plot_comparisons`, the user can pass their own values for the covariates in `conditional` as a dict. If we allowed the same subplot figure control as `plot_cap`, this would require the user to pass a nested dictionary and would result in more code needed to "unnest" the dictionary to access the desired key, value pairs.

For example:

```python3
fig, ax = plot_comparison(
    model=penguin_model,
    idata=penguin_idata,
    contrast_predictor=["flipper_length_mm"],
    conditional=["bill_length_mm", "species"],
    subplot_kwargs={"main": "bill_length_mm", "group": "species", "panel": "species"},
    fig_kwargs={"figsize": (10, 3), "sharey": True},
    legend=False
)
```
image

Now this raises the question if we should add the same `subplot_kwargs` to `plot_cap` to stay consistent?

GStechschulte:
Commit 41d4565bcf29727bc1254b0e1519254e2b3a9ebb adds the ability to compute multi-level contrast comparisons. This is achieved by first computing all pairwise orderings of the contrast value. Then, the `xr.DataArray` is indexed using each pair. 

For example, `contrast=prog` has 3 values (General, Vocational, Academic). Each pairwise contrast is shown in the output below. 

```python3
# if the user wants to compare > 2 levels. Use the comparison function directly
comparisons(
    model=model_interaction,
    idata=idata_interaction,
    contrast="prog",
    conditional="math"
)
```
math | term | contrast | estimate | hdi_3% | hdi_97%
-- | -- | -- | -- | -- | --
1.000000 | prog | ['Academic', 'General'] | 5.705757 | -2.780949 | 15.583213
1.000000 | prog | ['Academic', 'Vocational'] | 5.678390 | -2.237369 | 15.903185
1.000000 | prog | ['General', 'Vocational'] | 5.651256 | -2.166023 | 15.727247
... | ... | ... | ... | ... | ...
99.000000 | prog | ['Academic', 'General'] | -4.931466 | -10.382592 | -0.191101
99.000000 | prog | ['Academic', 'Vocational'] | -4.909654 | -10.369010 | -0.098217
99.000000 | prog | ['General', 'Vocational'] | -4.887939 | -10.395197 | -0.081602

bambi/plots/__init__.py
```diff
@@ -1,3 +1,16 @@
-from .plot_cap import create_cap_data, plot_cap
+from .create_data import create_cap_data, create_comparisons_data
```
tomicapretto, line 1: Here I would do `from bambi.plots.create_data import ...` always writing `bambi.plots` before `.create_data` so it's not relative anymore.

bambi/plots/create_data.py
```diff
@@ -0,0 +1,123 @@
+import pandas as pd
+import itertools
+
+import bambi as bmb
```
tomicapretto, line 4: I wouldn't import `bambi` in a submodule. It's better to import the things we need from it as we do with, for example, `from bambi.utils import clean_formula_lhs`

bambi/plots/effects.py
```diff
@@ -0,0 +1,349 @@
...
+import pandas as pd
+import xarray as xr
+
+import bambi as bmb
+from bambi.utils import listify, get_aliased_name
+from bambi.plots.create_data import create_cap_data, create_comparisons_data
+from bambi.plots.utils import identity, Comparison
+
+
+def predictions(
```
tomicapretto, line 19: I like the name of the function. In the future, `plot_cap()` should be called `plot_predictions()` as well.

bambi/plots/effects.py
```diff
@@ -0,0 +1,349 @@
...
+@dataclass
+class Response:
+    name: str
+    target: str
+
+
+@dataclass
+class ContrastEstimate:
+    comparison: Dict[str, xr.DataArray]
+    hdi: Dict[str, xr.Dataset]
```
tomicapretto, line 137: Should these go here? I'm not 100% what are the "best practices" for dataclasses

bambi/plots/effects.py
```diff
@@ -0,0 +1,349 @@
...
+        contrast value, and conditional values.
+
+    Raises
+    ------
+    ValueError
+        When ``hdi_prob`` is not > 0 and < 1.
+    """
+
+    if isinstance(contrast, dict):
+        contrast_name, contrast_level = next(iter(contrast.items()))
```
tomicapretto, line 194: If I understand correctly, when you have a dict like `{"a": 1, "b":2, "c": 3}, this sets `contrast_name` to `"a"` and `contrast_level` to `1`. Why do we keep only the key/value of the first pair?

tomicapretto: Also, it seems you only care about `contrast_name`? I'm not sure where `contrast_level` is used.

tomicapretto: perhaps `next(iter(contrast.keys()))` is better, and I would add a comment explaining why you do that

bambi/plots/effects.py
```diff
@@ -0,0 +1,349 @@
...
+    elif isinstance(contrast, list):
+        contrast_name = contrast[0]
+    elif isinstance(contrast, str):
+        contrast_name = contrast
```
tomicapretto, line 198: what happens if it's not `dict`, `list` or `str`? We should raise something.

bambi/plots/effects.py
```diff
@@ -0,0 +1,349 @@
...
+            select_draw = select_draw.assign_coords(
+                {f"{response.name}_obs": np.arange(len(select_draw.coords[f"{response.name}_obs"]))}
+            )
+            draws[val] = select_draw
+
+        # iterate over pairwise combinations of contrast values
+        pairwise_contrasts = list(itertools.combinations(contrast.value, 2))
+
+        functions = {"diff": lambda x, y: x - y, "ratio": lambda x, y: x / y}
+        function = functions[comparison_type]
```
tomicapretto, line 259: Would be good to have a friendly error message here

bambi/plots/plot_types.py
```diff
@@ -0,0 +1,190 @@
+from matplotlib.lines import Line2D
+from matplotlib.patches import Patch
+import numpy as np
+import pandas as pd
+
+from bambi.plots.utils import identity, get_unique_levels, get_group_offset, Covariates
+
+
+def plot_numeric(
```
tomicapretto, line 9: These work for both comparisons and predictions?! Cool!!

bambi/plots/plotting.py
```diff
@@ -0,0 +1,277 @@
...
+from pandas.api.types import is_categorical_dtype, is_numeric_dtype, is_string_dtype
+
+import bambi as bmb
+from bambi.utils import listify, get_aliased_name
+from bambi.plots.effects import predictions, comparisons
+from bambi.plots.plot_types import plot_numeric, plot_categoric
+from bambi.plots.utils import get_covariates
+
+
+def plot_cap(
```
tomicapretto, line 21: So basically all the hard work you did on the other modules allowed you to get much lighter versions of `plot_cap()` and `plot_comparisons()`... great!

tomicapretto (changes requested):

bambi/plots/create_data.py
GStechschulte: Ahh good point. I was importing it for type hints. But you are right, in a submodule, I could do `from Bambi import Model` and use `Model`

bambi/plots/effects.py
GStechschulte: Should I go ahead and make the change here in this branch and PR? Or should a separate PR be made for this? (to avoid too much "other" developments being made in this PR)

bambi/plots/effects.py
GStechschulte: I thought the same thing. I debated putting them either here or in the `utils.py` module. However, these classes aren't used by the other effects functions (predictions or slopes). I decided to use a class instead of a dict to have a strongly typed object with named parameters that have an attached type. Then, I use `@dataclass` to create the constructor with params., and to store the params. into the fields, etc. Aka it lets me to not write as much boilerplate code. 

That was my thought process for using the classes and dataclass decorator.

bambi/plots/effects.py
```diff
@@ -0,0 +1,349 @@
...
+    contrast : str, dict, list
+        The predictor name whose contrast we would like to compare.
+    conditional : str, dict, list
+        The covariates we would like to condition on.
+    comparison_type : str, optional
+        The type of comparison to plot. Defaults to 'diff'.
+    target : str
+        Which model parameter to plot. Defaults to 'mean'. Passing a parameter into target only
+        works when pps is False as the target may not be available in the posterior predictive
+        distribution.
```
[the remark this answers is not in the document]
GStechschulte: I also have not seen this elsewhere. At the moment, I think we should remove this arg. and potentially add it later if time permits?

bambi/plots/effects.py
GStechschulte: Regarding:

_Comment 1_: Correct, and I only keep the key/value of the first pair since I have not added functionality for computing cross-contrasts (when a user wants to compare > 1 contrast predictor)

_Comment 2 & 3_: `contrast_level` is not used. Changed to use `.keys()`. Perhaps an error should be raised if they attempt to pass > 1 contrast predictor as this is not implemented.

bambi/plots/effects.py
```diff
@@ -0,0 +1,349 @@
...
+            )
+            select_draw = select_draw.assign_coords(
+                {f"{response.name}_obs": np.arange(len(select_draw.coords[f"{response.name}_obs"]))}
+            )
+            draws[val] = select_draw
+
+        # iterate over pairwise combinations of contrast values
+        pairwise_contrasts = list(itertools.combinations(contrast.value, 2))
+
+        functions = {"diff": lambda x, y: x - y, "ratio": lambda x, y: x / y}
```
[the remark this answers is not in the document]
GStechschulte: Should this be outside the nested `_compute_contrast_estimate` function, but inside the the `comparisons` function? Or by having all capital letters, are you implying it should be treated as a global variable and outside of the `comparisons` function?

bambi/plots/effects.py
tomicapretto: It can be a separate PR after this is done. No need to worry about it now.

bambi/plots/effects.py
```diff
@@ -0,0 +1,349 @@
...
+    response_name = get_aliased_name(model.response_component.response_term)
+    response_transform = transforms.get(response_name, identity)
+
+    if pps:
+        idata = model.predict(idata, data=cap_data, inplace=False, kind="pps")
+        y_hat = response_transform(idata.posterior_predictive[response_name])
+        y_hat_mean = y_hat.mean(("chain", "draw"))
+    else:
+        idata = model.predict(idata, data=cap_data, inplace=False)
+        y_hat = response_transform(idata.posterior[f"{response_name}_{target}"])
```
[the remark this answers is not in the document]
tomicapretto:
```python
component  = model.components[target]
if component.alias: # not sure if this is the right syntax
    component_name = get_aliased_name(component)
    idata.posterior[component_name]
else:
    idata.posterior[f"{response_name}_{target}"]
```

bambi/plots/effects.py
```diff
@@ -0,0 +1,349 @@
...
+        y_hat_bounds = az.hdi(y_hat, hdi_prob)[response_name].T
+    elif use_hdi:
+        y_hat_bounds = az.hdi(y_hat, hdi_prob)[f"{response_name}_{target}"].T
+    else:
+        lower_bound = round((1 - hdi_prob) / 2, 4)
+        upper_bound = 1 - lower_bound
+        y_hat_bounds = y_hat.quantile(q=(lower_bound, upper_bound), dim=("chain", "draw"))
+
+    cap_data["estimate"] = y_hat_mean
+    cap_data["hdi_3%"] = y_hat_bounds[0]
```
tomicapretto, line 116: we should clean the name of the lower and upper bound to reflect `hdi_prob`

bambi/plots/effects.py
tomicapretto: it's ok to leave them here

bambi/plots/effects.py
```diff
@@ -0,0 +1,349 @@
...
+        if len(contrast.value) < 3:
+            contrast_df = comparisons_df.drop_duplicates(list(conditional.values())).reset_index(
+                drop=True
+            )
+            contrast_df = contrast_df.drop(columns=contrast.term)
+            N = contrast_df.shape[0]
+            contrast_df["term"] = contrast.term
+            contrast_df["contrast"] = list(np.tile(contrast.value, N).reshape(N, 2))
+            contrast_df["estimate"] = contrast_estimate.comparison[tuple(contrast.value)].to_numpy()
+            contrast_df["hdi_3%"] = (
```
tomicapretto, line 295: flagging the label may not be always right

bambi/plots/effects.py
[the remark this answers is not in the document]
tomicapretto: Sounds good, let's remove it for now.

GStechschulte:
In addition to the requested changes by @tomicapretto, the latest commits in this PR added the following functionality:

- `average_by` argument in `comparisons` and `plot_comparisons`
- `subplot_kwargs` in `plot_cap` to follow the same design as `plot_comparisons`
- organised the `test_plots.py` file into three classes: (1) TestCommon to test common args. of both `plot_cap` and `plot_comparisons`; usually regarding Matplotlib figure args, (2) TestCap tests args. specific to `plot_cap`, and (3) TestComparisons tests args. specific to `plot_comparisons`

Here, I give a brief example of `average_by`. For example:

```python3
fish_data = pd.read_stata("http://www.stata-press.com/data/r11/fish.dta")
cols = ["count", "livebait", "camper", "persons", "child"]
fish_data = fish_data[cols]
fish_data["livebait"] = fish_data["livebait"].astype("category")
fish_data["camper"] = fish_data["camper"].astype("category")

likelihood = bmb.Likelihood("ZeroInflatedPoisson", params=["mu", "psi"], parent="mu")
links = {"mu": "log", "psi": "logit"}
zip_family = bmb.Family("zip", likelihood, links)
priors = {"psi": bmb.Prior("Beta", alpha=3, beta=3)}

fish_model = bmb.Model(
    "count ~ livebait + camper + persons + child", 
    fish_data, 
    priors=priors,
    family=zip_family
)

fish_idata = fish_model.fit(draws=1000, target_accept=0.95, random_seed=1234, chains=4)

comparisons(
    model=fish_model,
    idata=fish_idata,
    contrast="camper",
    conditional=["livebait", "child", "persons"]
)
```
term | contrast | livebait | child | persons | estimate | hdi_0.03% | hdi_0.97%
camper | (0.0, 1.0) | 0.0 | 0.0 | 1.0 | 0.185616 | 0.086099 | 0.291901
camper | (0.0, 1.0) | 0.0 | 0.0 | 2.0 | 0.443242 | 0.212235 | 0.679702
camper | (0.0, 1.0) | 0.0 | 0.0 | 4.0 | 2.542203 | 1.309796 | 3.867047
...| ...| ... | ... | ... | ... | ... | ...
camper | (0.0, 1.0) | 1.0 | 3.0 | 1.0 | 0.016145 | 0.006866 | 0.025888
camper | (0.0, 1.0) | 1.0 | 3.0 | 2.0 | 0.038481 | 0.016707 | 0.060103
camper | (0.0, 1.0) | 1.0 | 3.0 | 4.0 | 0.219881 | 0.112066 | 0.343412

A user can pass a covariate(s) they would like to average by. For example:

```python3
# marginalizes over child and persons
comparisons(
    model=fish_model,
    idata=fish_idata,
    contrast="camper",
    conditional=["livebait", "child", "persons"],
    average_by="livebait"
)
```
  | term | contrast | livebait | estimate | hdi_0.03% | hdi_0.97%
-- | -- | -- | -- | -- | -- | --
camper | (0.0, 1.0) | 0.0 | 0.445599 | 0.223418 | 0.683028
camper | (0.0, 1.0) | 1.0 | 2.442100 | 1.830418 | 3.058506

Passing `livebait` to `average_by` averages by [0, 1] which marginalises over the other covariates `child` and `persons` to get the average estimate and uncertainty of `livebait`. This can also be plotted:

```python3
plot_comparison(
    model=fish_model,
    idata=fish_idata,
    contrast="camper",
    conditional=["livebait", "child", "persons"],
    average_by="livebait"
)
```
image


I will be adding a notebook for the docs explaining the functionality and how to use `plot_comparisons` in the coming week.

GStechschulte:
> ## [Codecov](https://app.codecov.io/gh/bambinos/bambi/pull/684?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bambinos) Report
> > Merging [#684](https://app.codecov.io/gh/bambinos/bambi/pull/684?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bambinos) ([5ae78cb](https://github.com/bambinos/bambi/commit/5ae78cb34eeed62f0d015c51293b7a0fb00e783f)) into [main](https://app.codecov.io/gh/bambinos/bambi/commit/e1b0e8ccbed265d6c995aa8ed59f55e50bdc8383?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bambinos) ([e1b0e8c](https://github.com/bambinos/bambi/commit/e1b0e8ccbed265d6c995aa8ed59f55e50bdc8383)) will **decrease** coverage by `1.24%`.
> > The diff coverage is `82.63%`.
> 
> ```diff
> @@            Coverage Diff             @@
> ##             main     #684      +/-   ##
> ==========================================
> - Coverage   88.33%   87.10%   -1.24%     
> ==========================================
>   Files          40       43       +3     
>   Lines        2880     3210     +330     
> ==========================================
> + Hits         2544     2796     +252     
> - Misses        336      414      +78     
> ```
> 
> [Impacted Files](https://app.codecov.io/gh/bambinos/bambi/pull/684?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bambinos)	Coverage Δ	
> [bambi/plots/utils.py](https://app.codecov.io/gh/bambinos/bambi/pull/684?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bambinos#diff-YmFtYmkvcGxvdHMvdXRpbHMucHk=)	`76.02% <75.00%> (-23.98%)`	⬇️
> [bambi/plots/effects.py](https://app.codecov.io/gh/bambinos/bambi/pull/684?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bambinos#diff-YmFtYmkvcGxvdHMvZWZmZWN0cy5weQ==)	`81.41% <81.41%> (ø)`	
> [bambi/plots/plot_types.py](https://app.codecov.io/gh/bambinos/bambi/pull/684?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bambinos#diff-YmFtYmkvcGxvdHMvcGxvdF90eXBlcy5weQ==)	`83.48% <83.48%> (ø)`	
> [bambi/plots/plotting.py](https://app.codecov.io/gh/bambinos/bambi/pull/684?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bambinos#diff-YmFtYmkvcGxvdHMvcGxvdHRpbmcucHk=)	`91.57% <91.57%> (ø)`	
> [bambi/plots/create_data.py](https://app.codecov.io/gh/bambinos/bambi/pull/684?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bambinos#diff-YmFtYmkvcGxvdHMvY3JlYXRlX2RhdGEucHk=)	`92.50% <92.50%> (ø)`	
> [bambi/models.py](https://app.codecov.io/gh/bambinos/bambi/pull/684?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bambinos#diff-YmFtYmkvbW9kZWxzLnB5)	`79.16% <100.00%> (ø)`	
> [bambi/plots/__init__.py](https://app.codecov.io/gh/bambinos/bambi/pull/684?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bambinos#diff-YmFtYmkvcGxvdHMvX19pbml0X18ucHk=)	`100.00% <100.00%> (ø)`	
> 📣 We’re building smart automated test selection to slash your CI/CD build times. [Learn more](https://about.codecov.io/iterative-testing/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bambinos)

Mmmm. Code coverage decreased in `bambi/plots/utils.py`. However, these functions are implicitly tested by testing the functions in the `effects.py` and `plotting.py` (since these modules call the functions in `utils.py`. Thus, the reason for not adding explicit tests for the functions in `utils.py`. Does someone with more experience in performing tests have a comment on this? @tomicapretto @aloctavodia?

tomicapretto:
I'm not 100% sure what's the cause. But if we see this https://app.codecov.io/gh/bambinos/bambi/pull/684/blob/bambi/plots/utils.py it indicates `contrast_dtype()` is not tested as well as `get_model_covariates()`.

bambi/plots/__init__.py
```diff
@@ -1,3 +1,16 @@
...
+
+__all__ = [
+    "create_cap_data",
+    "create_comparisons_data",
+    "comparisons",
+    "predictions",
+    "plot_cap",
+    "plot_comparison",
+    "plot_numeric",
+    "plot_categoric",
```
tomicapretto, line 15: Do you think users will make use of these functions? Similar for `create_cap_data` and `create_comparisons_data`

bambi/plots/create_data.py
```diff
@@ -0,0 +1,128 @@
...
+
+def create_comparisons_data(
+    condition: ConditionalInfo, contrast: ContrastInfo, user_passed: bool = False
+) -> pd.DataFrame:
+    """Create data for a Conditional Adjusted Comparisons
+
+    Parameters
+    ----------
+    comparisons : ComparisonInfo
+        An dataclass instance containing the model, contrast, and conditional
```
tomicapretto, line 59: A~n~ dataclass

bambi/plots/create_data.py
```diff
@@ -0,0 +1,128 @@
...
+    data_dict = set_default_values(model, data_dict, kind="predictions")
+    return enforce_dtypes(data, pd.DataFrame(data_dict))
+
+
+def create_comparisons_data(
+    condition: ConditionalInfo, contrast: ContrastInfo, user_passed: bool = False
+) -> pd.DataFrame:
+    """Create data for a Conditional Adjusted Comparisons
+
+    Parameters
```
tomicapretto, line 56: missing docstring for the `condition` argument

bambi/plots/effects.py
```diff
@@ -0,0 +1,402 @@
...
+        lower_bound = round((1 - prob) / 2, 4)
+        upper_bound = 1 - lower_bound
+        y_hat_bounds = y_hat.quantile(q=(lower_bound, upper_bound), dim=("chain", "draw"))
+
+    lower_bound = round((1 - prob) / 2, 4)
+    upper_bound = 1 - lower_bound
+
+    cap_data["estimate"] = y_hat_mean
+    cap_data[f"lower_{lower_bound}%"] = y_hat_bounds[0]
+    cap_data[f"upper_{upper_bound}%"] = y_hat_bounds[1]
```
tomicapretto, line 119: I think these will look like `upper_0.97%` instead of `upper_97%`, right? Just need to multiply by 100 above.

bambi/plots/effects.py
```diff
@@ -0,0 +1,402 @@
...
+        using a contrast mask. Then, pairwise comparisons are computed for the
+        contrast values. Finally, the mean comparison and lower/upper bounds
+        are computed for each pairwise comparison.
+        """
+        function = comparison_functions[comparison_type]
+
+        draws = {}
+        for idx, val in enumerate(contrast.values):
+            mask = np.array(comparisons_df[contrast.name] == contrast.values[idx])
+            select_draw = idata.posterior[f"{response.name}_{response.target}"].sel(
```
tomicapretto, line 245: This will have a similar conflict as the one mentioned above when using aliases for non-parent parameters.

bambi/plots/effects.py
```diff
@@ -0,0 +1,402 @@
...
+                            .sel(hdi="higher")
+                            .values
+                        )
+                    )
+                else:
+                    lower.append(contrast_estimate.hdi[tuple(pair)].sel(quantile=lower_bound))
+                    upper.append(contrast_estimate.hdi[tuple(pair)].sel(quantile=upper_bound))
+
+            contrast_df[f"lower_{lower_bound}%"] = np.array(lower).flatten()
+            contrast_df[f"upper_{upper_bound}%"] = np.array(upper).flatten()
```
tomicapretto, line 370: Just want to mention I see many f-strings written on the fly. It's not problematic per se. But you may want to consider creating variables to hold these names and then use the variables. Makes it easier to maintain and understand what they represent.

bambi/plots/plot_types.py
```diff
@@ -0,0 +1,190 @@
...
+    """Plotting of numeric data types.
+
+    Parameters
+    ----------
+    covariates : Covariates
+        Covariates callable with attributes main, group, panel.
+    plot_data : pd.DataFrame
+        The data created by the `create_cap_data` or `create_comparisons_data`
+        function.
+
```
tomicapretto, line 25: missing docstring for `transforms`, `legend` and `axes`.

bambi/plots/plot_types.py
```diff
@@ -0,0 +1,190 @@
...
+    """Plotting of categorical data types.
+
+    Parameters
+    ----------
+    covariates : Covariates
+        Covariates callable with attributes main, group, panel.
+    plot_data : pd.DataFrame
+        The data created by the `create_cap_data` or `create_comparisons_data`
+        function.
+
```
tomicapretto, line 121: Same as above

bambi/plots/plotting.py
```diff
@@ -0,0 +1,299 @@
...
+    legend : bool, optional
+        Whether to automatically include a legend in the plot. Defaults to ``True``.
+    transforms : dict, optional
+        Transformations that are applied to each of the variables being plotted. The keys are the
+        name of the variables, and the values are functions to be applied. Defaults to ``None``.
+    ax : matplotlib.axes._subplots.AxesSubplot, optional
+        A matplotlib axes object or a sequence of them. If None, this function instantiates a
+        new axes object. Defaults to ``None``.
+    subplot_kwargs : optional
+        Keyword arguments passed to the matplotlib subplot function as a dict. This allows you
```
tomicapretto, line 65: Actually, these are not passed to `plt.subplot()`, right? I would say these kwargs determine how different variables are mapped to different aspects of the visualization, without focusing on _how_ it happens internally.

tomicapretto (changes requested):
The PR is 95% done. It's super clean and you wrote very high-quality code. Thanks a lot for that!

Most of the changes I'm requesting are "cosmetic" or docstring related changes. The only "big" thing is the updates needed to make sure aliases on non-parent parameters still work. Don't hesitate to ask if you want help here (either to build an example/test or to implement it)

bambi/plots/__init__.py
GStechschulte: No, they shouldn't. Users should only use `plot_*` along with `predictions` & `comparisons`.

bambi/plots/plotting.py
GStechschulte: You are right; good catch.

bambi/plots/effects.py
GStechschulte: I like that tip. Thanks for the feedback!

bambi/plots/effects.py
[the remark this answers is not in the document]
GStechschulte: I'm not sure I know what "non-parent parameter" means. Is this referring to if there's an alias on the response, then the aliased name will not be the same as the `response_name`? 

If `target != "mean"`, then logic needs to be put in place to parse the aliased name if aliased is true. Else, use the response name and default `targt="mean"`?

bambi/plots/effects.py
GStechschulte: I replied to your comment above regarding "non-parent parameters"

bambi/plots/effects.py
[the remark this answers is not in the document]
tomicapretto:
```python
formula = bmb.Formula("y ~ x", "sigma ~ x")
model = bmb.Model(formula)
# y_sigma
# f"{aliased_response_name}_{parameter_name}"

# However, if you assign an alias to "sigma", then that pattern does not work anymore

aliases = {"sigma": "whatever"}

# f"{aliased_response_name}_{parameter_name}" does not work anymore
# Bambi does not prepend the response name to non-parent parameters that have an alias
# it will be 'whatever', not 'y_whatever'
```

bambi/plots/effects.py
GStechschulte: Now that I remember, we previously discussed that for this first PR, `plot_comparisons` will not have the functionality to plot and compare non-parent parameters. Therefore there is no `target` arg. and no possibility of a `KeyError` being thrown since we use `get_aliased_name` for the response.

tomicapretto (approved):
Looks like this is good to go! 


@GStechschulte let me know if I can merge :)

GStechschulte:
> Looks like this is good to go!
> 
> @GStechschulte let me know if I can merge :)

Looks like we can merge 👍🏼 Again, thanks for all the code reviews and insights. Much appreciated!

PR #689, opened 2023-06-24
Update additional dependencies install command

tjburch:
Was starting to do some development and ran into the following:

```bash
pip install .[dev]
zsh: no matches found: .[dev]
```

Found that adding the single quote around it resolved it, so updated the instructions accordingly:

```bash
pip install '.[dev]'
```

And similar for the jax dependencies.

I assume this is ubiquitous, but if it's just a zsh for some reason, feel free to disregard.


tomicapretto:
I was not aware of this potential issue! Thanks for fixing it. I confirm it still works on my side.

PR #690, opened 2023-06-24
Update predict pps docstring

### LINKED ISSUE #679: Document `Model.predict(kind="pps")`
`Model.predict(kind="pps")` will also store the mean value (and any other distributional parameters) in the inference data object. This is not documented and may be surprising.

tjburch:
Resolves #679 by updating the `model.predict` docstring to make it clear that the `"pps"` kind also returns the mean posterior distribution.


tomicapretto:
Thanks a lot!!

PR #691, opened 2023-06-25
Add warning for aliases athat aren't used

### LINKED ISSUE #675: Raise an exception (or warning) when key doesn't exist in `set_alias`
If you use a key that doesn't match any term name, it silently omits it and moves on. It would be good if we had at least a warning so users are aware some of the information they passed is being omitted. 

This is common when you have interaction effects or group-specific effects.

tjburch:
Resolves #675 by throwing a warning if any aliases are not used. I didn't want to make the warnings too verbose, so if there's 5 or fewer aliases that aren't used, it explicitly tells you which aren't used. If more than 5 aren't used, it throws a generic message.

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -491,6 +491,7 @@ class Model:
             raise ValueError(f"'aliases' must be a dictionary, not a {type(aliases)}.")
 
         response_name = get_aliased_name(self.response_component.response_term)
+        missing_names = []
 
         # If there is a single distributional component (the response)
         #   * Keys are the names of the terms and the values are their aliases.
@@ -517,6 +518,17 @@ class Model:
                 for term in self.response_component.group_specific_terms.values():
                     if name in term.prior.args:
                         term.hyperprior_alias = {name: alias}
+
+                if (
+                    not any(
+                        name in term.prior.args
+                        for term in self.response_component.group_specific_terms.values()
+                    )
+                    and name not in self.response_component.terms
+                    and name not in self.constant_components
+                    and name != response_name
+                ):
+                    missing_names.append(name)
         else:
             for component_name, component_aliases in aliases.items():
                 if component_name in self.constant_components:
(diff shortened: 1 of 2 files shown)


bambi/models.py
```diff
@@ -517,6 +518,17 @@ def set_alias(self, aliases):
                 for term in self.response_component.group_specific_terms.values():
                     if name in term.prior.args:
                         term.hyperprior_alias = {name: alias}
+
+                if (
```
canyon289, line 522: Add a comment to this one and the code block below so when reading this module its more clear how the list, and the `if missing_name` block below interact

tjburch: done

canyon289: Thanks! Tomas can comment on the rest of the logic. LGTM but he knows best!

tomicapretto: If I understand correctly this is "asking" the same questions that were in the for loop again. If they are all False, it means none of the if-blocks in the previous if-statements got executed, implying the alias was not used. 

What do you think about this alternative approach (pseudo-code)?


```python
for name, alias in aliases.items():
    is_used = False
    
    if cond1:
        ....
        is_used = True

    if cond2:
        ...
        is_used = True


     # finally
     if not is_used:
          missing_names.append(name)
```

I know it introduces a new auxiliary variable. But I think it's slightly clearer since the human-reader doesn't need to parse a "complex" if statement with multiple conditions.

tomicapretto:
Thanks a lot again @tjburch for the useful contribution. And thanks @canyon289 for the review!

I think the logic is good. Still, I left a comment suggesting an approach that may be a little bit better. Let me know what you think.

The only thing I would add to the PR is a test that checks the warning is raised. It can go into `tests/test_aliases.py`

bambi/models.py
tjburch: Definitely cleaner. Added.

tjburch:
Thanks for the reviews. Made the specified changes, and just added the tests too.

PR #698, opened 2023-07-14
Fixes color mismatch in plot_cap

### LINKED ISSUE #685: Minor plotting error in `plot_cap`
Working with a toy dataset I notice that I get a great plot from `plot_cap`. This is from the model `y ~ 0 + C(group)`. Note how the line colour and shading colour match up.
Screenshot 2023-06-18 at 09 43 37

But there seems to be an issue in an example with the model `y ~ 1`. Now, the line colour and fill colour no longer match up
Screenshot 2023-06-18 at 09 44 19

Data was generated with:
```python
groups = 3
samples_per_group = 10
N = groups * samples_per_group
x = np.tile(np.linspace(0, 10, samples_per_group), groups)
group = np.repeat(np.arange(groups), samples_per_group)
y = 2 + -0.5 * group + rng.normal(0, 0.1, N)
df = pd.DataFrame({"x": x, "group": group, "y": y})
```

tjburch:
Resolves #685 by explicitly using the first color in the matplotlib cylcer when plotting with a single covariate. Implemented in both numeric and categoric plot types.

After fix, reproducing the issue yields the following:

solved_bug

### DIFF
--- a/bambi/plots/plot_types.py
+++ b/bambi/plots/plot_types.py
@@ -45,8 +45,8 @@ def plot_numeric(
     if len(covariates) == 1:
         ax = axes[0]
         values_main = transform_main(plot_data[main])
-        ax.plot(values_main, y_hat_mean, solid_capstyle="butt")
-        ax.fill_between(values_main, y_hat_bounds[0], y_hat_bounds[1], alpha=0.4)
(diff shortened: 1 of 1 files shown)


tomicapretto:
Thanks a lot @tjburch !

PR #699, opened 2023-07-20
plot_slopes and slopes

GStechschulte:
This draft PR introduces `slopes` and `plot_slopes`. Slopes can be defined as the "partial derivative of the regression equation with respect to (wrt) a regressor of interest" (definition taken from [marginaleffects](https://vincentarelbundock.github.io/marginaleffects/articles/slopes.html)). `slopes` and `plot_slopes` allow the user to view a summary dataframe and or plot the slope of the regressor of interest (wrt). This PR supports the following quantities of interest:

- unit level slopes
- average slopes
- average by group slopes
- user defined value slopes

Additionally, this PR also supports the interpretation of slopes as an elasticity (percent change in $x$ is associated with a percentage change in $y$). If the regressor of interest is categorical, computing an elasticity is not possible. For example, there is no % change in penguin species. Furthermore, if `wrt` is categorical when calling `slopes`, then `comparisons` is called to compute the difference in contrast means.

`slopes` and `comparisons` are closely related. Thus, I have refactored the code such that these two functions can use the same code when building the summary dataframe and plotting. Lastly, I simplified the code to build the summary data frames.

### DIFF
--- a/bambi/plots/__init__.py
+++ b/bambi/plots/__init__.py
@@ -1,4 +1,4 @@
-from bambi.plots.effects import comparisons, slopes, predictions
+from bambi.plots.effects import comparisons, predictions, slopes
 from bambi.plots.plotting import plot_cap, plot_comparison, plot_slopes
 
 
--- a/bambi/plots/__init__.py
+++ b/bambi/plots/__init__.py
@@ -1,5 +1,5 @@
-from bambi.plots.effects import comparisons, predictions
-from bambi.plots.plotting import plot_cap, plot_comparison
+from bambi.plots.effects import comparisons, slopes, predictions
+from bambi.plots.plotting import plot_cap, plot_comparison, plot_slopes
 
 
-__all__ = ["comparisons", "predictions", "plot_cap", "plot_comparison"]
+__all__ = ["comparisons", "slopes", "predictions", "plot_cap", "plot_comparison", "plot_slopes"]
--- a/bambi/plots/create_data.py
+++ b/bambi/plots/create_data.py
@@ -16,97 +16,93 @@ from bambi.plots.utils import (
 )
 
 
+def _grid_level(
+    condition_info: ConditionalInfo, variable_info: VariableInfo, user_passed: bool, kind: str
+):
+    """
+    Creates a "grid" of data by using the covariates passed into the
+    `conditional` argument. Values for the grid are either: (1) computed
+    using a equally spaced grid, mean, and or mode (depending on the
+    covariate dtype), and (2) a user specified value or range of values.
+    """
+    covariates = get_covariates(condition_info.covariates)
+
+    if user_passed:
+        data_dict = {**condition_info.conditional}
+    else:
+        main_values = make_main_values(condition_info.model.data[covariates.main])
+        data_dict = {covariates.main: main_values}
+        data_dict = make_group_panel_values(
+            condition_info.model.data,
+            data_dict,
+            covariates.main,
+            covariates.group,
+            covariates.panel,
+            kind=kind,
+        )
+
+    data_dict[variable_info.name] = variable_info.values
+    comparison_data = set_default_values(condition_info.model, data_dict, kind=kind)
+    # use cartesian product (cross join) to create pairwise grid
+    keys, values = zip(*comparison_data.items())
+    pairwise_grid = pd.DataFrame([dict(zip(keys, v)) for v in itertools.product(*values)])
+    # cannot enfore dtypes if 'slopes' because it may remove floating point of dy/dx
+    if kind == "comparisons":
+        pairwise_grid = enforce_dtypes(condition_info.model.data, pairwise_grid)
+
+    return pairwise_grid
+
+
+def _unit_level(variable_info: VariableInfo, kind: str) -> pd.DataFrame:
+    """
+    Creates the data for unit-level contrasts by using the observed (empirical)
+    data. All covariates in the model are included in the data, except for the
+    contrast predictor. The contrast predictor is replaced with either: (1) the
+    default contrast value, or (2) the user specified contrast value.
+    """
+    covariates = get_model_covariates(variable_info.model)
+    df = variable_info.model.data[covariates].drop(labels=variable_info.name, axis=1)
+
+    variable_vals = variable_info.values
+
+    if kind == "comparisons":
+        variable_vals = np.array(variable_info.values)[..., None]
+        variable_vals = np.repeat(variable_vals, variable_info.model.data.shape[0], axis=1)
+
+    contrast_df_dict = {}
+    for idx, value in enumerate(variable_vals):
+        contrast_df_dict[f"contrast_{idx}"] = df.copy()
+        contrast_df_dict[f"contrast_{idx}"][variable_info.name] = value
+
+    return pd.concat(contrast_df_dict.values())
+
+
 def create_differences_data(
     condition_info: ConditionalInfo, variable_info: VariableInfo, user_passed: bool, kind: str
 ) -> pd.DataFrame:
-    def _grid_level():
-        """
-        Creates a "grid" of data by using the covariates passed into the
-        `conditional` argument. Values for the grid are either: (1) computed
-        using a equally spaced grid, mean, and or mode (depending on the
-        covariate dtype), and (2) a user specified value or range of values.
-        """
-        covariates = get_covariates(condition_info.covariates)
+    """
+    Creates either unit level or grid level data for 'comparisons' and 'slopes'
+    depending if the user passed covariate values.
+    """
 
-        if user_passed:
-            data_dict = {**condition_info.conditional}
-        else:
-            main_values = make_main_values(condition_info.model.data[covariates.main])
-            data_dict = {covariates.main: main_values}
-            data_dict = make_group_panel_values(
-                condition_info.model.data,
-                data_dict,
-                covariates.main,
-                covariates.group,
-                covariates.panel,
-                kind=kind,
-            )
-
-        data_dict[variable_info.name] = variable_info.values
-        comparison_data = set_default_values(condition_info.model, data_dict, kind=kind)
-        # use cartesian product (cross join) to create pairwise grid
-        keys, values = zip(*comparison_data.items())
-        pairwise_grid = pd.DataFrame([dict(zip(keys, v)) for v in itertools.product(*values)])
-
-        # cannot enfore dtypes if 'slopes' because it may remove floating point of dy/dx
-        if kind == "comparisons":
-            pairwise_grid = enforce_dtypes(condition_info.model.data, pairwise_grid)
-
-        return pairwise_grid
-
-    def _unit_level():
-        """
-        Creates the data for unit-level contrasts by using the observed (empirical)
-        data. All covariates in the model are included in the data, except for the
-        contrast predictor. The contrast predictor is replaced with either: (1) the
-        default contrast value, or (2) the user specified contrast value.
-        """
-        covariates = get_model_covariates(variable_info.model)
-        df = variable_info.model.data[covariates].drop(labels=variable_info.name, axis=1)
-
-        variable_vals = variable_info.values
-
-        if kind == "comparisons":
-            variable_vals = np.array(variable_info.values)[..., None]
-            variable_vals = np.repeat(variable_vals, variable_info.model.data.shape[0], axis=1)
-
-        contrast_df_dict = {}
-        for idx, value in enumerate(variable_vals):
-            contrast_df_dict[f"contrast_{idx}"] = df.copy()
-            contrast_df_dict[f"contrast_{idx}"][variable_info.name] = value
-
-        return pd.concat(contrast_df_dict.values())
-
-    return _unit_level() if not condition_info.covariates else _grid_level()
+    if not condition_info.covariates:
+        return _unit_level(variable_info, kind)
+    else:
+        return _grid_level(condition_info, variable_info, user_passed, kind)
 
 
 def create_cap_data(model: Model, covariates: dict) -> pd.DataFrame:
-    """Create data for a Conditional Adjusted Predictions
-
-    Parameters
-    ----------
-    model : bambi.Model
-        An instance of a Bambi model
-    covariates : dict
-        A dictionary of length between one and three.
-        Keys must be taken from ("horizontal", "color", "panel").
-        The values indicate the names of variables.
-
-    Returns
-    -------
-    pandas.DataFrame
-        The data for the Conditional Adjusted Predictions dataframe and or
-        plotting.
+    """
+    Creates a data grid for conditional adjusted predictions using the covariates
+    passed by the user.
     """
     data = model.data
     covariates = get_covariates(covariates)
     main, group, panel = covariates.main, covariates.group, covariates.panel
 
-    # Obtain data for main variable
     main_values = make_main_values(data[main])
     data_dict = {main: main_values}
 
-    # Obtain data for group and panel variables if not None
     data_dict = make_group_panel_values(data, data_dict, main, group, panel, kind="predictions")
     data_dict = set_default_values(model, data_dict, kind="predictions")
 
--- a/bambi/plots/create_data.py
+++ b/bambi/plots/create_data.py
@@ -18,12 +18,30 @@ from bambi.plots.utils import (
 
 def _grid_level(
     condition_info: ConditionalInfo, variable_info: VariableInfo, user_passed: bool, kind: str
-):
-    """
-    Creates a "grid" of data by using the covariates passed into the
+) -> pd.DataFrame:
+    """Creates a "grid" of data by using the covariates passed into the
     `conditional` argument. Values for the grid are either: (1) computed
     using a equally spaced grid, mean, and or mode (depending on the
     covariate dtype), and (2) a user specified value or range of values.
+
+    Parameters
+    ----------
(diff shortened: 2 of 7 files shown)


GStechschulte:
A demo is shown below:

```python3
import bambi as bmb
from bambi.plots import slopes, plot_slopes

data = bmb.load_data("mtcars")
model = bmb.Model("mpg ~ hp * wt + drat", data=data, family="gaussian")
idata = model.fit(draws=1000, chains=2)
```

Unit level slopes

```python3
slopes(
    model,
    idata,
    wrt="hp",
    conditional=None
).head(10)
```
term | estimate_type | value | drat | wt | estimate | lower_3.0% | upper_97.0%
hp | dydx | (110.0, 110.0001) | 3.90 | 2.620 | -0.046077 | -0.061826 | -0.027424
hp | dydx | (110.0, 110.0001) | 3.90 | 2.875 | -0.039438 | -0.055346 | -0.024399
hp | dydx | (93.0, 93.0001) | 3.85 | 2.320 | -0.053887 | -0.074023 | -0.033815
hp | dydx | (110.0, 110.0001) | 3.08 | 3.215 | -0.030586 | -0.045647 | -0.016320
hp | dydx | (175.0, 175.0001) | 3.15 | 3.440 | -0.024728 | -0.040884 | -0.009271
hp | dydx | (105.0, 105.0001) | 2.76 | 3.460 | -0.024207 | -0.040335 | -0.008568
hp | dydx | (245.0, 245.0001) | 3.21 | 3.570 | -0.021343 | -0.039777 | -0.006217
hp | dydx | (62.0, 62.0001) | 3.69 | 3.190 | -0.031237 | -0.046218 | -0.016832
hp | dydx | (95.0, 95.0001) | 3.92 | 3.150 | -0.032278 | -0.046864 | -0.017400
hp | dydx | (123.0, 123.0001) | 3.92 | 3.440 | -0.024728 | -0.040884 | -0.009271

Average slopes

```python3
slopes(
    model,
    idata,
    wrt="hp",
    conditional=None,
    average_by=True
)
```
term | estimate_type | estimate | lower_3.0% | upper_97.0%
-- | -- | -- | -- | --
hp | dydx | -0.030527 | -0.051032 | -0.01007

Average by group slopes (and plot)

```python3
fig, ax = plot_slopes(
    model,
    idata,
    wrt="hp",
    conditional=None,
    average_by="wt"
)
fig.set_size_inches(7, 3)
```
image

Condition on weight and plot elasticity

```python3
fig, ax = plot_slopes(
    model,
    idata,
    wrt="hp",
    conditional="wt",
    slope="eyex"
)
fig.set_size_inches(7, 3)
```
image

User provided values

```python3
fig, ax = plot_slopes(
    model,
    idata,
    wrt="hp",
    conditional={"wt": [2, 3, 5], "drat": [3.5, 4, 4.5]},
)
fig.set_size_inches(7, 3)
```
image

bambi/plots/__init__.py
```diff
@@ -1,5 +1,5 @@
-from bambi.plots.effects import comparisons, predictions
-from bambi.plots.plotting import plot_cap, plot_comparison
+from bambi.plots.effects import comparisons, slopes, predictions
```
tomicapretto, line 1: nit: alphabetic order `import comparisons, predictions, slopes`

bambi/plots/__init__.py
```diff
@@ -1,5 +1,5 @@
...
+from bambi.plots.plotting import plot_cap, plot_comparison, plot_slopes
 
 
-__all__ = ["comparisons", "predictions", "plot_cap", "plot_comparison"]
+__all__ = ["comparisons", "slopes", "predictions", "plot_cap", "plot_comparison", "plot_slopes"]
```
tomicapretto, line 5: same about alphabetic order

bambi/plots/create_data.py
```diff
@@ -6,124 +6,108 @@
...
 
-        return enforce_dtypes(condition.model.data, pd.DataFrame(contrast_dict))
+        # cannot enfore dtypes if 'slopes' because it may remove floating point of dy/dx
+        if kind == "comparisons":
+            pairwise_grid = enforce_dtypes(condition_info.model.data, pairwise_grid)
 
-    def _unit_level(contrast: ContrastInfo):
+        return pairwise_grid
+
+    def _unit_level():
```
tomicapretto, line 57: I would like to avoid functions with no argument that do something depending on a value in an upper-level namespace (i.e. `variable_info` in this case). Could you add the variable_info as an argument to the function?

tomicapretto: same thing with `_grid_level()` which I've just noticed

bambi/plots/plot_types.py
```diff
@@ -45,8 +45,8 @@ def plot_numeric(
     if len(covariates) == 1:
         ax = axes[0]
         values_main = transform_main(plot_data[main])
-        ax.plot(values_main, y_hat_mean, solid_capstyle="butt", color="C0")
-        ax.fill_between(values_main, y_hat_bounds[0], y_hat_bounds[1], color="C0", alpha=0.4)
+        ax.plot(values_main, y_hat_mean, solid_capstyle="butt")
```
tomicapretto, line 48: The `color="C0"` was added in #689 to fix #685. We should keep the fix here as well (unless you have a better option, of course)

bambi/plots/plot_types.py
```diff
@@ -155,8 +155,8 @@ def plot_categoric(covariates: Covariates, plot_data: pd.DataFrame, legend: bool
 
     if len(covariates) == 1:
         ax = axes[0]
-        ax.scatter(idxs_main, y_hat_mean, color="C0")
-        ax.vlines(idxs_main, y_hat_bounds[0], y_hat_bounds[1], color="C0")
+        ax.scatter(idxs_main, y_hat_mean)
+        ax.vlines(idxs_main, y_hat_bounds[0], y_hat_bounds[1])
```
tomicapretto, line 159: same as above

tomicapretto (changes requested):
Providing a quick review. I couldn't find big issues with the code, it's already almost done. Still, I'm requesting some changes (stylistic and minor updates). As always, fantastic work. Thanks a lot!

bambi/plots/plot_types.py
GStechschulte: Definitely. I must have neglected this in my rebase when I was resolving conflicts.

bambi/plots/utils.py
```diff
@@ -312,12 +317,13 @@ def make_group_values(x: np.ndarray, groups_n: int = 5) -> np.ndarray:
...
 
 
-def get_unique_levels(x: np.ndarray) -> Union[list, np.ndarray]:
+def get_unique_levels(x: np.ndarray) -> np.ndarray:
     """
     Get unique levels of a categoric variable.
     """
     if hasattr(x, "dtype") and hasattr(x.dtype, "categories"):
-        levels = list(x.dtype.categories)
+        # levels = list(x.dtype.categories)
```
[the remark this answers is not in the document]
GStechschulte: Yup, removed.

GStechschulte:
> Providing a quick review. I couldn't find big issues with the code, it's already almost done. Still, I'm requesting some changes (stylistic and minor updates). As always, fantastic work. Thanks a lot!
> 
> **Edit** It would be good to explain what is the `value` column in the returned data frame

Thanks a lot for the review. :) Much appreciated. In the latest commits I realised I forgot to explain what the _value_ column. I will add this as an inline comment and also explain it in the docs.

GStechschulte:
To Do: 

- [ ] add tests
- [ ] resolve pylint errors
- [ ] add the following semi-elasticity slopes (currently `slopes` only supports `eyex`): `eydx` and `dyex`

GStechschulte:
The latest commits added the following functionality:

- semi-elasticities (`eydx`,  `dyex`)
- user provided multiple values with `wrt`

and added tests for `plot_slopes`. Below, I provide an example of added functionality (using the model and data from above):

User provided multiple values with `wrt`. The `value` column represents the values of the `wrt` arg. used to compute the derivative. If the user passes multiple values, a small amount $\epsilon$ `eps` is added to each value and then divided by that `eps` to obtain the "instantaneous rate of change":

```python3
slopes(
    model,
    idata,
    wrt={"hp": [150, 200, 250]},
    conditional=["wt"]
)
```
term | estimate_type | value | wt | drat | estimate | lower_3.0% | upper_97.0%
hp | dydx | (150.0, 150.0001) | 1.513000 | 3.596563 | -0.074025 | -0.104541 | -0.041595
hp | dydx | (200.0, 200.0001) | 1.513000 | 3.596563 | -0.072006 | -0.102318 | -0.041779
hp | dydx | (250.0, 250.0001) | 1.513000 | 3.596563 | -0.069988 | -0.099388 | -0.040974

`eydx`: unit increase in $x$ (`wrt`) is associated with a % change in $y$:

```python3
plot_slopes(
    model,
    idata,
    wrt={"hp": 150},
    conditional=["wt"],
    slope="eydx"
)
```
image

`dyex`: % change in $x$ (`wrt`) is associated with a unit increase in $y$:

```python3
fig, ax = plot_slopes(
    model,
    idata,
    wrt={"hp": 150},
    conditional=["wt"],
    slope="dyex"
)
```
image

bambi/plots/create_data.py
```diff
@@ -6,124 +6,106 @@
...
+
+    contrast_df_dict = {}
+    for idx, value in enumerate(variable_vals):
+        contrast_df_dict[f"contrast_{idx}"] = df.copy()
+        contrast_df_dict[f"contrast_{idx}"][variable_info.name] = value
+
+    return pd.concat(contrast_df_dict.values())
+
+
+def create_differences_data(
```
tomicapretto, line 82: Could you document the arguments in this function?

bambi/plots/create_data.py
```diff
@@ -6,124 +6,106 @@
...
+    depending if the user passed covariate values.
+    """
+
+    if not condition_info.covariates:
+        return _unit_level(variable_info, kind)
+    else:
+        return _grid_level(condition_info, variable_info, user_passed, kind)
+
+
 def create_cap_data(model: Model, covariates: dict) -> pd.DataFrame:
```
tomicapretto, line 96: Same as above, document the arguments.

bambi/plots/utils.py
```diff
@@ -12,25 +14,74 @@
 
 
 @dataclass
-class ContrastInfo:
+class VariableInfo:
```
tomicapretto, line 17: Can you document this dataclass? I'm not so sure that thorough documentation of all the arguments is needed, but it would be nice to know what's the goal of this dataclass and where it is used.

bambi/plots/utils.py
```diff
@@ -12,25 +14,74 @@
...
+                                    values = self.centered_difference(
+                                        predictor_data, self.eps, dtype
+                                    )
+                            elif component.kind == "categoric":
+                                values = get_unique_levels(predictor_data)
+
+        return values
 
 
 @dataclass
```
tomicapretto, line 87: same here

bambi/plots/utils.py
```diff
@@ -236,56 +287,14 @@ def set_default_values(model: Model, data_dict: dict, kind: str):
...
-                        if component.kind == "numeric":
-                            contrast = _numeric_difference(np.mean(model.data[name])).astype(
-                                contrast_dtype
-                            )
-                        # For categoric predictors, select the most frequent level.
-                        elif component.kind == "categoric":
-                            contrast = get_unique_levels(model.data[name])
-
-    return contrast
+    # elif kind == "predictions":
```
tomicapretto, line 296: should this be removed?

bambi/plots/utils.py
```diff
@@ -12,25 +14,74 @@
...
+        elif not isinstance(self.variable, (list, dict, str)):
+            raise TypeError("`variable` must be a list, dict, or string")
+
+    def centered_difference(self, x, eps, dtype):
+        return np.array([x - eps, x + eps], dtype=dtype)
+
+    def epsilon_difference(self, x, eps):
+        return np.array([x, x + eps])
+
+    def set_default_variable_values(self):
```
tomicapretto, line 57: could you document this to explain what are the defults in each case?

bambi/plots/utils.py
tomicapretto: I don't know where to add this, but I'm seeing the following block inside ConditionalInfo

```python
        elif isinstance(self.conditional, dict):
            self.covariates = {k: listify(v) for k, v in self.conditional.items()}
            self.covariates = dict(zip(covariate_kinds, self.conditional))
            self.user_passed = True
```

isn't the first line within the `elif` not required? You save something into `self.covariates` and then you rewrite it with something different. Or is that a bug?

bambi/plots/effects.py
```diff
@@ -1,18 +1,21 @@
...
-# pylint: disable = too-many-function-args
-# pylint: disable = too-many-nested-blocks
 from dataclasses import dataclass, field
 import itertools
 from typing import Dict, Union
 
 import arviz as az
 import numpy as np
 import pandas as pd
+from pandas.api.types import is_categorical_dtype, is_string_dtype
```
tomicapretto, line 8: Could you sort the imports so this `from ...` appears below all the `import x as y`?

bambi/plots/effects.py
```diff
@@ -41,6 +44,260 @@ def __post_init__(self):
...
+            self.higher = np.array(
+                [self.bounds[key].sel(quantile=higher) for key in self.bounds]
+            ).flatten()
+
+
+SUPPORTED_SLOPES = ("dydx", "eyex")
+SUPPORTED_COMPARISONS = {
+    "diff": lambda x, y: x - y,
+    "ratio": lambda x, y: x / y,
+}
```
tomicapretto, line 87: Constants should be in the beginning of the code (after imports, before defining dataclasses, functions, etc.)

bambi/plots/effects.py
```diff
@@ -240,210 +508,180 @@ def comparisons(
...
+def slopes(
+    model: Model,
+    idata: az.InferenceData,
+    wrt: Union[str, dict],
+    conditional: Union[str, dict, list, None] = None,
+    average_by: Union[str, list, bool, None] = None,
+    eps: float = 1e-4,
+    slope: str = "dydx",
+    use_hdi: bool = True,
+    prob=None,
```
tomicapretto, line 560: can you add the type for this one and the `transforms` argument?

bambi/plots/plotting.py
```diff
@@ -253,55 +334,169 @@ def plot_comparison(
...
+def plot_slopes(
+    model: Model,
+    idata: az.InferenceData,
+    wrt: Union[str, dict],
+    conditional: Union[str, dict, list, None] = None,
+    average_by: Union[str, list] = None,
+    eps: float = 1e-4,
+    slope: str = "dydx",
+    use_hdi: bool = True,
+    prob=None,
```
tomicapretto, line 359: types here too :)

bambi/plots/plot_types.py
```diff
@@ -46,7 +46,7 @@ def plot_numeric(
         ax = axes[0]
         values_main = transform_main(plot_data[main])
         ax.plot(values_main, y_hat_mean, solid_capstyle="butt", color="C0")
-        ax.fill_between(values_main, y_hat_bounds[0], y_hat_bounds[1], color="C0", alpha=0.4)
+        ax.fill_between(values_main, y_hat_bounds[0], y_hat_bounds[1], alpha=0.4)
```
tomicapretto, line 49: we need both `ax.plot()` and `ax.fill_between()` to have `color="C0"`

tomicapretto (changes requested):
Hi @GStechschulte! I'm requesting some changes. In terms of code, very minor things. But in general, I'm asking to add many docstrings.

GStechschulte:
> Hi @GStechschulte! I'm requesting some changes. In terms of code, very minor things. But in general, I'm asking to add many docstrings.

Thanks for the review! In regard to docstrings, I feel I must explain myself. The reason for me not adding docstrings to non-public modules was because I was following the PEP 8 section on [Documentation Strings](https://peps.python.org/pep-0008/#documentation-strings):

> Write docstrings for all public modules, functions, classes, and methods. Docstrings are not necessary for non-public methods, but you should have a comment that describes what the method does. This comment should appear after the "def" line.

Nonetheless, I recognise that there are docstrings for non-public modules in the Bambi codebase. I will add them, it is no problem 😄

bambi/plots/utils.py
GStechschulte: You are right. The first line is not required. Just removed and ran tests Tests passed :)

tomicapretto:
> > Hi @GStechschulte! I'm requesting some changes. In terms of code, very minor things. But in general, I'm asking to add many docstrings.
> 
> Thanks for the review! In regard to docstrings, I feel I must explain myself. The reason for me not adding docstrings to non-public modules was because I was following the PEP 8 section on [Documentation Strings](https://peps.python.org/pep-0008/#documentation-strings):
> 
> > Write docstrings for all public modules, functions, classes, and methods. Docstrings are not necessary for non-public methods, but you should have a comment that describes what the method does. This comment should appear after the "def" line.
> 
> Nonetheless, I recognise that there are docstrings for non-public modules in the Bambi codebase, and that they help in explainability. I will add them, it is no problem 😄

I agree it's not necessary to add docstrings to everything, especially when it starts to become redundant. However, some internal documentation is also nice to help others understand how things work. Thanks for the flexibility in adding the requested docstrings :)

tomicapretto (approved):

Issue #703, opened 2023-08-14
Posterior draws for comparisons(), predictions(), and slopes() functions


zwelitunyiswa:
@tomicapretto @GStechschulte Thanks for your hard work on this module of Bambi. I was playing around with it this past week and I am mighty impressed! I love that I can create a data frame with summaries that I can plot in the Bambi plotting module or any plotting library I prefer. That is huge for reporting.

I was wondering if it would be possible to add a function that returns the underlying posterior draws for the summaries for the comparisons(), predictions(), and slopes() functions, like you can do in marginaleffects  with the [posterior_draws()](https://vincentarelbundock.github.io/marginaleffects/articles/reference/posterior_draws.html) function?

That would be super useful for reporting probabilities, ROPE, etc.

tomicapretto:
Hi @zwelitunyiswa thanks for the nice words! All credit belongs to Gabriel :)

@GStechschulte do you think that would be possible?

GStechschulte:
Hey @zwelitunyiswa, thanks a lot for the kind words and feedback. Yeah, that is possible. I admit, I have yet to use the `posterior_draws()` function in margineffects. I will use it and see how they return the posterior draws. 

My initial thoughts are to either: (1) use a function to return an `az.InferenceData` object or (2) add an argument in comparisons(), predictions(), and slopes() such as `return_idata` that would return an `az.InferenceData` object.

I like the use of the inference data object since Bambi and PyMC users likely have experience with such data structures, and due to the n-dimensional nature of the posterior draws.

Let me think some more and do some prototyping, and I will communicate back here. Thanks!

GStechschulte:
Okay @zwelitunyiswa, so I have used the marginaleffects `posterior_draws()` function and believe the simple thing to do is to have an optional `return_posterior` arg in `comparisons()`, `predictions()`, and `slopes()` that merges the posterior draws with the corresponding observation that "produced" that draw and returns a dataframe. Below you can find an example:

```python3
fish_data = pd.read_stata("http://www.stata-press.com/data/r11/fish.dta")
cols = ["count", "livebait", "camper", "persons", "child"]
fish_data = fish_data[cols]
fish_data["livebait"] = pd.Categorical(fish_data["livebait"])
fish_data["camper"] = pd.Categorical(fish_data["camper"])

fish_model = bmb.Model(
    "count ~ livebait + camper + persons + child", 
    fish_data, 
    family='zero_inflated_poisson'
)

[...]

PR #708, opened 2023-08-20
Return posterior arg. in `predictions`, `comparisons`, and `slopes`

GStechschulte:
This PR addresses issue #703 by adding an argument `return_posterior: bool = False` in comparisons(), predictions(), and slopes() that merges the posterior draws with the corresponding observation that "produced" that draw and returns a dataframe. I describe the solution and an example in the issue, but I will copy and paste here also:

```python3
fish_data = pd.read_stata("http://www.stata-press.com/data/r11/fish.dta")
cols = ["count", "livebait", "camper", "persons", "child"]
fish_data = fish_data[cols]
fish_data["livebait"] = pd.Categorical(fish_data["livebait"])
fish_data["camper"] = pd.Categorical(fish_data["camper"])

fish_model = bmb.Model(
    "count ~ livebait + camper + persons + child", 
    fish_data, 
    family='zero_inflated_poisson'
)

fish_idata = fish_model.fit(
    draws=1000, 
    target_accept=0.95, 
    random_seed=1234, 
    chains=4
)
```

With `posterior_draws=True`, two data frames are returned. First, the summary data frame (no change here), and second, the posterior draws flattened and merged with the data used to generate the predictions.

```python3
summary_df, post_draws_df = comparisons(
    model=fish_model,
    idata=fish_idata,
    contrast={"persons": [1, 4]},
    conditional={"child": [0, 1, 2], "livebait": [0, 1]},
    return_posterior=True
)

post_draws_df
```

chain | draw | livebait_dim | camper_dim | Intercept | livebait | camper | persons | child | count_psi | count_mean | child_obs | livebait_obs | persons_obs | camper_obs
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
0 | 0 | 1.0 | 1.0 | -2.338197 | 1.592060 | 0.653525 | 0.843256 | -1.284886 | 0.595906 | 0.431100 | 0.0 | 0.0 | 1.0 | 1.0
0 | 1 | 1.0 | 1.0 | -2.305775 | 1.538601 | 0.690644 | 0.855629 | -1.326705 | 0.570563 | 0.467899 | 0.0 | 0.0 | 1.0 | 1.0
0 | 2 | 1.0 | 1.0 | -2.078147 | 1.368352 | 0.742767 | 0.821930 | -1.226769 | 0.623266 | 0.598427 | 0.0 | 0.0 | 1.0 | 1.0
0 | 3 | 1.0 | 1.0 | -2.790560 | 2.066620 | 0.542978 | 0.875392 | -1.345364 | 0.541648 | 0.253551 | 0.0 | 0.0 | 1.0 | 1.0
0 | 4 | 1.0 | 1.0 | -2.557727 | 1.797058 | 0.677152 | 0.856107 | -1.414570 | 0.632578 | 0.358988 | 0.0 | 0.0 | 1.0 | 1.0
... | ... | ... | ... | ... | ... | ... | ... | ... | ... | ... | ... | ... | ... | ...
3 | 995 | 1.0 | 1.0 | -2.796267 | 1.891494 | 0.719850 | 0.900202 | -1.393952 | 0.602064 | 1.873827 | 2.0 | 1.0 | 4.0 | 1.0
3 | 996 | 1.0 | 1.0 | -2.675291 | 1.730114 | 0.732860 | 0.886734 | -1.188741 | 0.583542 | 2.604234 | 2.0 | 1.0 | 4.0 | 1.0
3 | 997 | 1.0 | 1.0 | -2.389797 | 1.695540 | 0.602298 | 0.857744 | -1.504039 | 0.633454 | 1.392279 | 2.0 | 1.0 | 4.0 | 1.0
3 | 998 | 1.0 | 1.0 | -2.556063 | 1.641193 | 0.721424 | 0.882217 | -1.259443 | 0.579240 | 2.262654 | 2.0 | 1.0 | 4.0 | 1.0
3 | 999 | 1.0 | 1.0 | -2.663126 | 1.902570 | 0.519382 | 0.900782 | -1.361

where the columns are:
- columns with substring "_obs" indicate the data corresponding to that draw
- columns with substring "_dim" indicate categorical data (only present if you encoded data as factors / categorical)
- chain & draw are self explanatory
- the remaining columns are the estimates of the parameters for that draw

I have also updated the corresponding notebooks to include sections on how to return posterior draws. 

However, now I have a question regarding adding tests. This arg. was only added for non-plotting functions. Before, testing the plotting functionality would implicitly test predictions(), comparisons(), and slopes(). However, now tests would need to be added to explicitly test arguments not in the plotting functions. Should I create a new test file `tests/test_interpret.py` that contains tests not specific to plotting?


tomicapretto:
Thanks for this great PR @GStechschulte! 

I agree a `tests/test_interpret.py` is needed. However, I would only include those things that are not tested in `tests/test_plots.py`. 

One thought I just had. Are we aware of any drawbacks associated with returning a variable number of objects? I can imagine cases where this is desired (i.e. you want both the draws and the summary and you don't want to call the function twice) but I can also imagine cases where you get an error because you change an argument value without allocating a new variable. Right now, I need to do something like

```python
if return_posterior:
    summary_df, post_draws_df = comparisons(
        model=fish_model,
        idata=fish_idata,
        contrast={"persons": [1, 4]},
        conditional={"child": [0, 1, 2], "livebait": [0, 1]},
        return_posterior=True
    )
else:
    summary_df = comparisons(
        model=fish_model,
        idata=fish_idata,
        contrast={"persons": [1, 4]},
        conditional={"child": [0, 1, 2], "livebait": [0, 1]},
        return_posterior=False
    )
```

If it always returns one object, this would work

```python
output_df = comparisons(
    model=fish_model,
    idata=fish_idata,
    contrast={"persons": [1, 4]},
    conditional={"child": [0, 1, 2], "livebait": [0, 1]},
    return_posterior=return_posterior
)
```

aloctavodia:
Why do you prefer a DataFrame over an InferenceData? 

I agree with Tomás, it is better to return a single object. A 2-tuple, with the second element `None` by default?

GStechschulte:
> Why do you prefer a DataFrame over an InferenceData?
> 
> I agree with Tomás, it is better to return a single object. A 2-tuple, with the second element `None` by default?

I decided on a DataFrame as it was not so obvious to me on how to merge the data into the `az.InferenceData` object. 

I could [add new groups](https://python.arviz.org/en/stable/getting_started/WorkingWithInferenceData.html#add-groups-to-inferencedata-objects) to the inference data object, but then it isn't clear to me how to perform group by aggregations on the posterior dataset while taking into account this new group. 

Additionally, the data shouldn't be merged as a data variable in the `az.InferenceData.posterior` dataset because, when aggregations are performed along the coordinates, these aggregations will also be applied to the data used to generate the predictions (since they were merged as a data variable).

Lastly, the data could be merged and made as a coordinate so you can specify along which dimension(s) you want to compute the aggregation. This does indeed work.

GStechschulte:
Thanks @tomicapretto and @aloctavodia for the comments regarding the returning of a variable number of objects. If the user uses [iterable unpacking ](https://realpython.com/lessons/tuple-assignment-packing-unpacking/) then they would get an error if they changed the arg. to `False`. However, if everything is stored in a single tuple object, then no errors will be raised if you change the arg. to `False`.

```python3
output = bmb.interpret.slopes(
    model,
    idata,
    wrt={"hp": [150, 200, 250]},
    conditional=["wt"],
    return_posterior=True
)

out[0]
```

term | estimate_type | value | wt | drat | estimate | lower_3.0% | upper_97.0%
hp | dydx | (150.0, 150.0001) | 1.513000 | 3.596563 | -0.074478 | -0.107209 | -0.046854
hp | dydx | (200.0, 200.0001) | 1.513000 | 3.596563 | -0.072424 | -0.103653 | -0.045472
... | ... | ... | ... | ... | ... | ... | ...
hp | dydx | (200.0, 200.0001) | 5.344184 | 3.596563 | 0.017953 | -0.016046 | 0.053197
hp | dydx | (250.0, 250.0001) | 5.344184 | 3.596563 | 0.020007 | -0.016298 | 0.055597

```python3
out[1]
```

chain | draw | Intercept | hp | wt | hp:wt | drat | mpg_sigma | mpg_mean | wt_obs | hp_obs | drat_obs
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
0 | 0 | 55.148882 | -0.127908 | -7.656893 | 0.027188 | -1.665181 | 2.575231 | 24.559235 | 1.513 | 150.0000 | 3.596563
0 | 1 | 39.805886 | -0.128098 | -7.311933 | 0.030934 | 1.861646 | 2.373092 | 23.244213 | 1.513 | 150.0000 | 3.596563
... | ... | ... | ... | ... | ... | ... | ... | ... | ... | ... | ...
1 | 995 | 39.386338 | -0.097691 | -7.049042 | 0.023804 | 1.646870 | 2.024121 | 14.930625 | 5.424 | 250.0001 | 3.596563
1 | 996 | 42.190725 | -0.100035 | -6.823994 | 0.021666 | 1.100839 | 2.140400 | 13.506382 | 5.424 | 250.0001 | 3.596563


No error since the tuple can store a variable number of objects.

tomicapretto:
The only concern about tuples is: do users who only come for the summaries want to learn they're getting a tuple (instead of a data frame) because there's another option that allows them to work with the draws?

As far as I'm seeing, there's no approach without drawbacks. I think we should check what other libraries do in similar situations.

GStechschulte:
> The only concern about tuples is: do users who only come for the summaries want to learn they're getting a tuple (instead of a data frame) because there's another option that allows them to work with the draws?

I suppose this would only be a concern if we do `return (summary_df, None)` by default. I should have stated that my code above is without changing the default return statement to include `None` and implied that if the user does not use iterable unpacking, no error would be thrown when they change the arg. to `False`.

Agreed, I will try to find an example in another library to see what they do. Good idea 👍🏼

PR #709, opened 2023-08-20
Add Governance

tomicapretto:
Several months ago I talked to some people mentioning the idea that I wanted to add a governance structure to Bambi to make this a better project for all. Finally, we have a draft here. Highlights are

* I copied this from PyMC and ArviZ
* The only strong opinion I have is that I don't want a BDFL model
    * Please, feel free to open a discussion if you think I'm wrong
* I do have some questions
    * What's an appropriate number of members for the Council? ArviZ has 7. I don't know if we have that many people but I suspect I'm not aware of the drawbacks of having a small council.
    * How do we decide who's a Core contributor a who's not? I guess it depends on merge permissions right now?
    * Is there anything you would remove from the document?
    * Is there anything you would like to add?

**This is a very first draft, everything can be discussed**.

Tagging all of you because I mentioned this to you at some point in time:

@aloctavodia @ahartikainen @ColCarroll @canyon289 @juanitorduz @junpenglao @tyarkoni @GStechschulte @zwelitunyiswa

Anyone else seeing this and wanting to participate, feel free to comment! You're more than welcomed.

I'm leaving this as it is for 7 days to collect feedback and ideas. Then we can start iterating.


twiecki:
I wonder if a council isn't overkill. A BDFL structure with core members I think is a better fit for a project of this size, and as core members grow, BDFL can be replaced with a council. Any reason you don't want a BDFL?

Having said that, I also don't think it's problematic other than creating (unnecessary) overhead.

tomicapretto:
maybe all core contributors are part of the council? that way we don't have to maintain two separate things. I think this is similar to how R Core works (I couldn't find any proper governance online)

canyon289:
CC @OriolAbril whos great at this stuff

GOVERNANCE.md
```diff
@@ -0,0 +1,356 @@
...
+* Steering council <!-- Do we specify the number of persons?-->
+
+Anyone working with Bambi has the responsibility to personally uphold the Code of Conduct. 
+Core Contributors have the additional responsibility of _enforcing_ the Code of Conduct to maintain
+a safe community.
+
+
+#### Recurrent Contributors
+
+Recurrent Contributors are those individuals who contribute recurrently to theproject and can
```
aloctavodia, line 47:
```suggestion
Recurrent Contributors are those individuals who contribute recurrently to the project and can 
```

aloctavodia, line 47:
```suggestion
Recurrent Contributors are those individuals who contribute recurrently to the project and can 
```

GOVERNANCE.md
```diff
@@ -0,0 +1,356 @@
...
+by interpreting yes=+1, neutral=0 and no=-1. If too many candidates were confirmed, the 10 
+candidates with higher rank are elected. If not enough candidates were chosen, the 4 candidates with
+ higher rank are elected.
+* In the event of a tie there will be a runoff election for the tied candidates. To avoid further 
+ties and discriminate more among the tied candidates, this vote will be held by 
+[Majority Judgment](https://en.wikipedia.org/wiki/Majority_judgment) (MJ): for each candidate, 
+voters judge their suitability for office as either "Excellent", "Very Good", "Good", "Acceptable", 
+"Poor", or "Reject". Multiple candidates may be given the same grade by a voter. 
+The candidate with the highest median grade is the winner.
+* If more than one candidate has the same highest median-grade, the MJ winner isdiscovered by
```
aloctavodia, line 271:
```suggestion
* If more than one candidate has the same highest median-grade, the MJ winner is discovered by 
```

GOVERNANCE.md
```diff
@@ -0,0 +1,356 @@
...
+"Poor", or "Reject". Multiple candidates may be given the same grade by a voter. 
+The candidate with the highest median grade is the winner.
+* If more than one candidate has the same highest median-grade, the MJ winner isdiscovered by 
+removing (one-by-one) any grades equal in value to the shared median grade from each tied 
+candidate's total. This is repeated until only one of the previously tied candidates is currently 
+found to have the highest median-grade.
+* If ties are still present after this second round, the winner will be chosen at random. 
+Each person tied will pick an integer number in the `[1, 100]` interval and send it privately to the
+third party mediating the election. After receiving all the numbers, said third party will draw a 
+random integer from random.org. The person with theclosest circular distance, defined as
```
aloctavodia, line 278:
```suggestion
random integer from random.org. The person with the closest circular distance, defined as 
```

GOVERNANCE.md
```diff
@@ -0,0 +1,356 @@
...
+found to have the highest median-grade.
+* If ties are still present after this second round, the winner will be chosen at random. 
+Each person tied will pick an integer number in the `[1, 100]` interval and send it privately to the
+third party mediating the election. After receiving all the numbers, said third party will draw a 
+random integer from random.org. The person with theclosest circular distance, defined as 
+`min(|a-b|, 100-|a-b|)`, will be selected. This process will be repeated as many times as necessary
+as there may be ties resulting from candidates choosing the same number.
+* At the conclusion of voting, all the results will be posted. And at least 24 hours will be left to
+challenge the election result in case there were suspicions of irregularities or the process had not
+been correctly carriedout.
```
aloctavodia, line 283:
```suggestion
been correctly carried out.
```

aloctavodia:
I am also in favor of a council and against a BDFL. 

The document refers to Bambi or Bambi project and then mentions that "The Project ... hosted in public GitHub repositories under the [bambinos GitHub organization](https://github.com/bambinos). "

Is this governance only for Bambi or it will include formulae and kulprit? I think it is better to include the 3 projects as they are closely related and small.

tomicapretto:
> Is this governance only for Bambi or it will include formulae and kulprit? I think it is better to include the 3 projects as they are closely related and small.

I agree, it should apply to the three projects.And I envision that if we happen to add another project to the organization, it will be intimately related to Bambi, so it would apply for that other case too.

GOVERNANCE.md
```diff
@@ -0,0 +1,356 @@
+# Main Governance Document
+
+## The Project
+
+The Bambi Project (The Project) is an open source software project.
+The goal of The Project is to develop open source software and deploy open and public websites and
```
canyon289, line 6:
```suggestion
The goal of The Project is to develop open source software and deploy open and public content and 
```
No change needed. websites seems specific, perhaps this can be broadened

GOVERNANCE.md
```diff
@@ -0,0 +1,356 @@
...
+decisions and interacts with other institutions, we see a need for a more formal governance model. 
+Moving forward The Project leadership will consist of a **TODO NAME** Council. 
+We view this governance model as the formalization of what we are already doing, rather than a 
+change in direction.
+
+
+<!-- NOTES:  
+
+Include something about our relationship with PyMC (PyMC helped us by giving us access to GSoC three times)
+How do we start a Council if we don't have 'core' contributors yet?
```
canyon289, line 355: Identify core contributors first through some critieria. Examples are asking anyone that's contributed to self nominate, that way you can start with folks who are involved and actively interested.

OriolAbril (commented):
I also prefer a council model, but I think the most important thing is not that but setting the culture and drafting the content and its layout to aim for the council/bdfl to be as uninvolved as possible.

> I'm not aware of the drawbacks of having a small council.

We did tinker with this a bit when defining ArviZ governance and it shouldn't be a disastrous default, but yYou might want to update the threshold percentages for votes. Current line 206 says at least 60% yes and no more than 20% no. For 3 people this means a vote passes only if there are 3 yes or 2 yes and 1 neutral. Anything else is not enough. No votes have a lot of weight as they go against a proposal o nomination (which is done by a trusted contributor, hence the obligation to reasoning), so 2 yes and 1 no would not be enough for example. With 4 people, 3 yes and 1 no would also not be enough for the vote to pass. 

> How do we decide who's a Core contributor a who's not? I guess it depends on merge permissions right now?

As @canyon289 commented somewhere, not necessarly right now. You can take that as a start but also have a call for self-nominations or something similar. What really matters is that after this process, their permissions (or the governance) are updated.

> I copied this from PyMC and ArviZ

They are more different than it initially appears, both on paper and in practice. It looks like you took ArviZ one, here are some examples in case you want to further discuss one of them:

* Recurrent contributor status is public in ArviZ, secret in PyMC
* It isn't very clear who can vote on ArviZ council elections (we didn't really have the discussion about who votes after the first election but it was even considered for the whole community to vote, keeping only core contributors as eligible), in PyMC it is and only core contributors do
* All core contributors get merge permissions in ArviZ even if they don't need them, not in PyMC, it depends on their teams
*  Differences in handling institutional partners and sponsors (which is currently missing completely from bambi's governance)

tomicapretto:
After this is merged we need to open an issue ticket to add people to the recurring contributor list and emeritus team members.

GStechschulte (approved):
Thanks for initiating this @tomicapretto. My knowledge on this subject is lacking, but it looks good to me 👍🏼

aloctavodia (approved):

GOVERNANCE.md
```diff
@@ -0,0 +1,340 @@
...
+## Governance
+
+### Community Architecture
+
+* General Contributors
+* Recurrent Contributors
+* Core Contributors (of which Council members are also a part of)
+* Steering Council
+
+Anyone working with Bambi has the responsibility to personally uphold the Code of Conduct.
```
canyon289, line 40: Link code of conduct

GOVERNANCE.md
```diff
@@ -0,0 +1,340 @@
...
+to fill the position.
+* There is no limit on the number of terms a Council Member can serve
+
+##### Current Council members
+
+The current Council members are:
+
+* Gabriel Stechschulte ([@gstechschulte](https://github.com/gstechschulte))
+* Osvaldo Martin ([@aloctavodia](https://github.com/aloctavodia))
+* Ravin Kumar ([@canyon289](https://github.com/canyon289))
```
canyon289, line 152:
```suggestion
```

Removing myself so you three have less voting overhead and with votes for people that are more familiar with the package

canyon289:
The only thing I see missing now is actual enforcement action. Doesnt need to be part of this governance doc, but somewhere it should be written that council can decide to ban anyone if theyre causing issues

canyon289 (approved):

PR #712, opened 2023-08-26
Migrate documentation to Quarto

tomicapretto:
* We don't use Sphinx anymore
* We use [`quartodoc`](https://github.com/machow/quartodoc/) which uses [`griffe`](https://pypi.org/project/griffe/) to extract signatures.
* I attempted to add a PR preview feature, but it was not possible, see comment below.

With this new PR our docs page looks so much better, we don't have the formatting issues with the HTML versions of the pandas data frames, and there's a new example gallery, among other changes.


tomicapretto:
I tried to add deploy previews but it doesn't work from forks. See:

https://github.com/rossjrw/pr-preview-action/discussions/48#discussioncomment-5821256

zwelitunyiswa:
Looks really nice! @tomicapretto

PR #719, opened 2023-09-13
Ordinal regression docs

### LINKED ISSUE #583: Ordinal models example
We still don't have ordinal models, but this dataset seems to be good for an example

https://archive.ics.uci.edu/ml/datasets/Estimation+of+obesity+levels+based+on+eating+habits+and+physical+condition+

GStechschulte:
This draft PR contains a notebook with an ordinal regression model for the newly added `cumulative` family. 

First, it is explained why ordered categorical outcomes require special treatment. Secondly, the basics of ordinal regression are explained, motivated through the use of the cumulative link function. Next, an intercept only model is fit to explain how the cumulative link function "describes" an ordered distribution. Lastly, a model with predictors is developed.

I will also add a section with the `sratio` family.

Edit by Tomas Capretto: Closes #583


tomicapretto:
Hi @GStechschulte! I recommend to rebase so it runs the latest version of the CI :)

GStechschulte:
> Hi @GStechschulte! I recommend to rebase so it runs the latest version of the CI :)
> 
> Btw, is there anything I can do here to help you?

Hey, and will do! Thanks :) Not necessarily. Although, there is a problem with ordinal models and the `interpret` sub-package. Since an ordinal model prediction is a vector of probabilities of length $K$, we get shape errors when attempting to use any of the functions in `interpret` 😵‍💫.

tomicapretto:
> > Hi @GStechschulte! I recommend to rebase so it runs the latest version of the CI :)
> > Btw, is there anything I can do here to help you?
> 
> Hey, and will do! Thanks :) Not necessarily. Although, there is a problem with ordinal models and the `interpret` sub-package. Since an ordinal model prediction is a vector of probabilities of length K, we get shape errors when attempting to use any of the functions in `interpret` 😵‍💫.
> 
> I need to see if marginaleffects supports ordinal models, and see if we can adopt their solution.

Isn't it similar to the case of the `"categorical"` family?

GStechschulte:
@tomicapretto yes, it should be similar to the `categorical` family.

docs/notebooks/ordinal_regression.ipynb
```diff
@@ -0,0 +1,1844 @@
+{
```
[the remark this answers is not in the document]
GStechschulte: <p>I like the detail, no worries! :) I agree. I will include both examples.</p><p> </p>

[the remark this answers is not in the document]
GStechschulte: <p>Yes, it works. :)</p>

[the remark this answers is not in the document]
GStechschulte: <p>In the first model, when I omit the intercept <code>response ~ 0 </code> the warning is no longer raised. I would need to explicitly include a model with the intercept to explain the behavior. Instead I explain this behavior in the paragraph before fitting the "threshold-only" model.</p>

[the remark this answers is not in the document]
GStechschulte: <p>In this model, if I omit the intercept I get <code>IndexError: list index out of range </code> originating from <code>formulae</code>.</p><p> </p><pre class="ql-syntax" spellcheck="false">File ~/Documents/repos/bambi/bambi/models.py:157, in Model.__init__(self, formula, data, family, priors, link, categorical, potentials, dropna, auto_scale, noncentered, center_predictors, extra_namespace)
    153     self.formula = check_ordinal_formula(self.formula)
    154     # Notice the intercept is added so formulae constrains categorical predictors, avoiding
    155     # linear dependencies with the cutpoints.
    156     # Then the intercept is removed from the design matrix because of the cutpoints.
--&gt; 157     design = fm.design_matrices(
    158         self.formula.main + " + 1", self.data, na_action, 1, additional_namespace
    159     )
    160     design = remove_common_intercept(design)
    161 else:

File ~/miniforge3/envs/bambinos/lib/python3.11/site-packages/formulae/matrices.py:556, in design_matrices(formula, data, na_action, env, extra_namespace)
    553     else:
    554         raise ValueError(f"'data' contains {incomplete_rows_n} incomplete rows.")
--&gt; 556 design = DesignMatrices(description, data, env)
...
-&gt; 1255         encoding = encodings[term.name][0]
   1256     else:
   1257         encoding = False

IndexError: list index out of range  
</pre>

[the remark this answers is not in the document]
tomicapretto: <p>I <em>think</em> it's because you need to update formulae. I still need to to the release, but in the meantime </p><p> </p><pre class="ql-syntax" spellcheck="false">pip install git+https://github.com/bambinos/formulae.git --force-reinstall
</pre><p> </p><p>Just make sure you don't get numpy 1.26.0 because it doesn't work with pytensor now.</p>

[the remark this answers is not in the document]
GStechschulte: This resolved the error. Thanks!

GStechschulte:
I think this PR is almost there. However, I am still questioning the interpretation of the threshold coefficients for the `sratio` model. I will paste my Slack message here. 

From my understanding, for ordinal models with the ‘sratio’ family and ‘logit’ link, the interpretation of the threshold coefficients are not cumulative-logits, but rather logits. This is in part because the ordering of the $k$ thresholds does not matter since each outcome is characterised by its own latent distribution. Whereas ordinal models with `family=cumulative` and `link=logit`, the coefficients of the thresholds are cumulative-logits.

However, when I plot the logits of the coefficients for the sratio model, it appears as though they are “partially” cumulative (overall, the probability increases as category increases, but for some categories the probability decreases which is not possible under a cumulative specification).

Plotting the PyMC graph for the sratio model, there are no constraints.

docs/notebooks/ordinal_regression.ipynb
```diff
@@ -0,0 +1,1798 @@
...
```
tomicapretto, line 1: <blockquote>Moreover, the response variable can also be a discrete count, and ordered</blockquote><p> </p><p>If a variable is a count then it's in a ratio scale (nominal &lt; ordinal &lt; interval &lt; ratio), which is more powerful than ordered, so it's not needed to mention that.</p><p> </p><blockquote>The distribution of ordinal responses may be nonnormal, particularly if very low or high values are infrequently chosen.</blockquote><p> </p><p>The most important reason why the distribution will be nonnormal is because the response variable is not continuous. Still, there are cases where the variable is not continuous but a normal distribution makes sense (imagine a count with a bell-shaped distribution). But is not usually the case in ordinal variables because the number of response levels is not usually high</p><p> </p><blockquote>(1) cumulative link, and (2) sequential link.</blockquote><p> </p><p>I would call them models or families, but not link, which makes me think of link functions</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/719/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

tomicapretto, line 1: <p>Maybe I'm confused, but according to the equation above, shouldn't P(Y = 3) be equal to Phi(tau_3) - Phi(tau_2)? But then, if there's tau_3 it means there are 4 categories (because number of categories is number of thresholds + 1)</p><p> </p><p>So when you are interested in Pr(Y = k) the result is 1 - Pr(Y = k - 1). The one is because we have Phi(infty) which is 1 because Phi is a CDF.</p><p> </p><p>So I'm thinking as I'm writing, but if we want to have it very complete we should write</p><p> </p><p> </p><p>P(Y = k) = F(tau_k) - F(-infty) = F(tau_k) - 0 if k = 1</p><p>P(Y = k) = F(tau_k) - F(tau_{k-1}) if  k in {2, 3, 4, ..., K - 1}</p><p>P(Y = k) = F(+infty) - F(tau_{k-1}) if  k = K</p><p> </p><p>So if we have three categories, then the probabilities are</p><p> </p><p>P(Y = 1) = F(tau_1)</p><p>P(Y = 2) = F(tau_2) - F(tau_1)</p><p>P(Y = 3) = 1 - F(tau_2)</p><p> </p><p> </p><p>Do you think we can update this part accordingly?</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/719/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

tomicapretto, line 1: <p>"cumulative link function" -&gt; "cumulative family"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/719/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

tomicapretto, line 1: <p>I know I'm being very nitpicky, but it can help users when they're reading this. There is no "ordered distribution", so I would say "recover the parameters of an ordinal model". You don't need to add the following, but I think the explanation is that the response distribution is not ordered. It's just a categorical distribution with its parameter vector built in some special way. This special way is that there is a parameter vector that goes through some transformation to obtain an ordered parameter vector.</p><p> </p><p>Again, not cumulative link function, but cumulative family. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/719/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

tomicapretto, line 1: <p>I think Pr(Y = 3) should be P(Z_1 &gt; tau_1) * P(Z_2 &gt; tau_2) * P(Z_3 &lt;= tau_3) which is (1 - P(Z_1 &lt;= tau_1)) * (1 - P(Z_2 &lt;= tau_2)) * P(Z_3 &lt;= tau_3)</p><p> </p><p>and then if the latent variable is normal, it would be (1 - Phi(tau_1)) * (1 - Phi(tau_2)) * Phi(tau_3), which is consistent with what we have here https://github.com/bambinos/bambi/blob/169564fe640818525049d7d946636266e4a7b968/bambi/families/univariate.py#L390</p><p> </p><p>and what you wrote below </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/719/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

tomicapretto, line 1: <p>I'm not  completely sure about this since it is not cumulative probability, right? I think these are the probabilities of making it to K + 1 years given that you made it to K years. </p><p> </p><p>Also, it makes sense, since the thresholds are not sorted anymore (you can see that in the mean in the summary above)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/719/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='35'/>

tomicapretto:
Hi Gabriel! This is a fantastic example, it's not simple at all and you're making a great job. I left some comments with suggestions and thoughts. Also:

* Can you silence `FutureWarning`s?
* Is the warning about the usage of the C-API making the model(s) run slower?
* I'm starting to think the prior for the threshold in the StoppingRatio family is not the most adequate. The means are sorted, but there's no reason to do that. So I'm thinking it makes more sense to have something like `Normal([0, 0, 0..., 0])` instead of what we have now here https://github.com/bambinos/bambi/blob/169564fe640818525049d7d946636266e4a7b968/bambi/priors/scaler.py#L117
what do you think about this last point?

docs/notebooks/ordinal_regression.ipynb
GStechschulte: <p>Thanks for the details. I agree with all of this.</p><p> </p>

GStechschulte: <p>Thanks for the clarification on the equations. That makes sense!</p>

GStechschulte: <p>Correct, it is not a cumulative probability. Ahhhh. I understand it now. For some reason this just now clicked for me.</p>

GStechschulte: <p>Yeah, that could be interpreted wrong. Thanks!</p>

GStechschulte: <p>Good eye. That is wrong. It should be $K-1$ thresholds where $K$ is the number of categories. Yes, changes are incorporated. :)</p>

GStechschulte:
Thank you! It has been a fun notebook to implement  😄

> * Can you silence `FutureWarning`s?

Done.

> * Is the warning about the usage of the C-API making the model(s) run slower?

Yes, it "feels" like it. I haven't ran any timed experiments though.

> * I'm starting to think the prior for the threshold in the StoppingRatio family is not the most adequate. The means are sorted, but there's no reason to do that. So I'm thinking it makes more sense to have something like `Normal([0, 0, 0..., 0])` instead of what we have now here https://github.com/bambinos/bambi/blob/169564fe640818525049d7d946636266e4a7b968/bambi/priors/scaler.py#L117

Agreed, since the ordering of the thresholds doesn't matter, doing something like `mu = np.zeros(response_level_n - 1)` makes more sense. I will try it out. What does brms do?

tomicapretto:
@GStechschulte 

> For now, I am thinking I keep this and explain how to change the default prior for the thresholds. Then, we could open a separate PR? What do you think?

I know it's more "tidy" to do it in a separate PR, but I think it's OK to do it in this one too. Anyway, I leave it up to you. Whatever you want is OK!

tomicapretto:
@GStechschulte 

> For now, I am thinking I keep this and explain how to change the default prior for the thresholds. Then, we could open a separate PR? What do you think?

I know it's more "tidy" to do it in a separate PR, but I think it's OK to do it in this one too. Anyway, I leave it up to you. Whatever you want is OK!

GStechschulte:
I implemented the zero $\mu$ vector in this PR 😄

bambi/priors/scaler.py
```diff
@@ -114,7 +114,7 @@ def scale_threshold(self):
             threshold = self.model.components["threshold"]
             if isinstance(threshold, ConstantComponent) and threshold.prior.auto_scale:
                 response_level_n = len(np.unique(self.response_component.response_term.data))
-                mu = np.round(np.linspace(-2, 2, num=response_level_n - 1), 2)
+                mu = np.zeros(response_level_n - 1)
```
GStechschulte, line 117: If sequential models assume that for every response level there is a latent continuous variable $Z_k$, then wouldn't we need each response level, i.e., `mu = response_level_n` and **not** `response_level_n  - 1`?

tomicapretto: Ohh, I think you're right! 

I'm thinking why the current approach is working and not failing. Is it because it's not considering the probability of Y being larger than the largest observed category? I think it would make sense for the years example, but I'm not sure if it would make sense for cases where there is a pre-specified set of categories. 

I wrote that as I was looking at this visualization from the ordinal tutorial by Bürkner and Vuorre:

image

and I wonder: Do we always have a Pr(Y > K)?

docs/notebooks/ordinal_regression.ipynb
```diff
@@ -0,0 +1,1695 @@
...
```
tomicapretto, line 1: <p>nit: the linear predictor does not contain the error term (for reference, the paragraph between eqs 2 and 3 in the tutorial by Buerkner and Vuorre</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/719/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

tomicapretto, line 1: <p>here you could also use the ppc function that I shared</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/719/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='40'/>

tomicapretto:
@GStechschulte on top of the comment that you raised, there are two nits. After that, feel free to merge. Excellent work!!

bambi/priors/scaler.py
GStechschulte: If I remove the `-1`, I get `ValueError: Incompatible Elemwise input shapes [(35,), (36,)]`. This makes sense as I stated in the docs because the sequential model is a product of probabilities, i.e., the probability that $Y$ is equal to category $k$ is equal to the probability that it did not fall in one of the former categories $1: k-1$ multiplied by the probability that the sequential process stopped at $k$.

GStechschulte:
Git got me good on this one 😢

tomicapretto:
I'm not sure if I follow what happened. Do you need help?

GStechschulte:
> I'm not sure if I follow what happened. Do you need help?

No haha. My local branch somehow diverged from this remote branch and I attempted to fix the conflicts manually. At the end the easiest was to force push. By the way, thanks for the reviews!

PR #716, opened 2023-09-03
[715] Adding Mr P docs

NathanielF:
## Mister P

* This is a PR trying to explain and illustrate the technique and value of Multilevel regression and post-stratification techniques using bambi. It is related to issue 715.

The structure of the notebook is seperated in two parts: (a) a refresher on how to think about the properties of the regression modelling as a method for generating an appropriately weighted average across the conditional strata in your in X matrix. We demonstrate this in some detail because MrP can be viewed as a correction to the effects of regression weighting when regression is applied to non-representative data. 

We then replicate the kind of analysis achieved with R [here](https://bookdown.org/jl5522/MRP-case-studies/)

PR #721, opened 2023-09-15
`interpret` support group effects

GStechschulte:
This PR resolves the `KeyError` in PR #716. This error was the result of no default values being computed for model terms specified as group level effects. This is because group level effects do not have a components attribute, and thus never satisfied the logic in the `set_default_values()` function of `../interpret/utils.py`. 

Now, to compute default values, all **unique** common and group specific effects variable names are identified by accessing the `components` and `factor` attributes and are put into a list. From there, the data type is determined by identifying the data types used to fit the model which is then used to determine the function call (`mean` or `mode` for numeric and categorical dtypes, respectively).

Additionally, with group specific effects, when the user (or default values are computed) passes values, it is possible these values are new unseen data. Thus, I leverage PR #693 and add the arg `sample_new_groups` to the functions `predictions`, `comparisons`, and `slopes` (and the associated plotting functions) to allow plotting of new groups.

For a working example, view my last comment in PR #716.

To do:

- [x] Add tests for new a model with group specific effects


tomicapretto:
I would like to ask two things: 

* Don't include anything related to the PR #716. As far as it is, this PR would add the notebook. I think it's because the branch was created from the branch in that PR. It should be created from the main branch.
* Could you add a test for the new supported cases? One reason why we didn't catch it before is that it was untested :sweat_smile:

GStechschulte:
Yes, you are right. I checked out this branch from Nathaniel's branch. I have removed everything from that branch now. As for tests go, I have added a model with common and group effects. I test each plotting function in `interpret` for new unseen data and non-new observed data.

tomicapretto (approved):
Looks great! You can hit the green button! (squash and merge)

PR #724, opened 2023-09-19
New ordered name

tomicapretto:
Two things happening here

 * I'm changing the name of the `univariate_ordered` transform for `ordered`.
 * And I'm pinning numpy to < 1.26.0 to make it work with pytensor


GStechschulte (approved):
LGTM 👍🏼 Thanks a lot! Once PR [444](https://github.com/pymc-devs/pytensor/pull/444) is merged, we should be able to remove the numpy < 1.26.

PR #725, opened 2023-09-20
Zero inflated docs

GStechschulte:
This PR adds documentation for two classes of zero inflated data: (1) Zero inflated poisson (ZIP), and (2) Hurdle poisson. 

First, I describe why zero inflated data needs such a class of models followed by a description and implementation of the ZIP and Hurdle models. Then, I describe the differences between the two models. 

I am not finished, but it is far enough ahead that I am drafting this PR.


tomicapretto:
> Notice how there are two linear models and two link functions: one for each process in the ZIPoisson

I would say "parameter" instead of "process"

tomicapretto:
I would say the example is already looking great! Will review again when it's done but so far I don't have any other suggestions.

GStechschulte:
Thank you! 👍🏼 If anything, I would like to have the Hurdle poisson section reviewed. In my opinion, the difference between the two is subtle, and I want to make sure my writing makes sense and is intuitive for the users. Thanks!

docs/notebooks/zero_inflated_regression.ipynb
```diff
@@ -0,0 +1,1905 @@
+{
```
tomicapretto, line 1: <p>Poisson better than "poisson" because it's a surname</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/725/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

tomicapretto, line 1: <p>I would write the "Number of Fish Caught" on the horizontal axis label so it's clear it refers to that.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/725/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

tomicapretto, line 1: <p>"as a linear combination of" -&gt; "as a function of a linear combination of"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/725/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

tomicapretto, line 1: <p>This is a little bit ambiguous: "the probability of the Poisson process, i.e., the probability of non-zero data, increases"</p><p> </p><p>It is correct to say the probability of the Poisson process increases, but that doesn't mean we have no zeroes, since zeroes are also possible under the Poisson distribution.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/725/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

tomicapretto, line 1: <blockquote> A zero-inflated model assumes that zero counts result from a mixture of two distributions</blockquote><p>This is true for both models. The difference in the ZIPoisson is that the zeroes can come from any of the components, while in the Hurdle Poisson they come only from one of the processes. You get this second thing right in your text, but I think you can be a little bit more explicit for the first. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/725/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

tomicapretto, line 1: <blockquote>However, notice how the hurdle model uses the Poisson cumulative distribution function to evaluate the probability of a zero. This can be thought of as the "zero mass" component</blockquote><p>This is correct, but I'm not sure it's clear. I would add that this evaluation of of a Poisson CDF at zero, but that doesn't play into the P(Y=0) but into the P(Y = y) where y &gt; 0. You explain it correctly in the final sentence.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/725/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

tomicapretto, line 1: <p>The warning is telling that the LOO-CV approximation may not be reliable, so we shouldn't use it to perform model comparison. I think I would remove this part, even though one can feel tempted to perform it.</p><p> </p><p>Instead, I would have a section where we reflect on the different data generating processes behind the two distributions. I <em>believe</em> this dataset is more reasonable for a ZIPoisson since the zeroes can appear because the group doesn't fish, or because they fished but they caught zero fish. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/725/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='38'/>

GStechschulte: <blockquote>&nbsp;but that doesn't mean we have no zeroes, since zeroes are also possible under the Poisson distribution.</blockquote><p> </p><p>Yeah, that's why I added the "probability of non-zero data" since it implies that zeros are still plausible. I wanted to add an interpretation of the statement "the probability of the Poisson process" for the users. For some it may be obvious, but for others not. </p>

tomicapretto:
Only one minor last thing (this time it's true!), because all the rest is perfect.

Change `np.linspace(..., num=20)` for `np.arange(39)`. This way every bar in the posterior predictive check visualization corresponds to one different count value and it's easier to conclude. Otherwise, the first bar represents more than the zeroes.

GStechschulte:
Thanks for the code reviews on this docs PR. Much appreciated! 😄

PR #726, opened 2023-09-22
Added missing doc string part for center_predictors

jt-lab:
Added missing doc string part for center_predictors

Perhaps the rationale behind centering could be added to help user decide whether to turn it on or not. But I'm not entirely sure about this.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,7 +50,7 @@ jobs:
         shell: bash -l {0}
         run: |
           echo "Running black..."
-          black bambi --check
+          black bambi --check --diff --color
 
           echo "Checking code style with pylint..."
           pylint bambi
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -89,7 +89,7 @@ class Model:
     center_predictors : bool
         If ``True`` (default), and if there is an intercept in the common terms, the data is
         centered by subtracting the mean. The centering is undone after sampling to provide
-        the actual intercept in all distributional components that have an intercept.  Note 
+        the actual intercept in all distributional components that have an intercept. Note
         that this changes the interpretation of the prior on the intercept because it refers
         to the intercept of the centered data.
     extra_namespace : dict, optional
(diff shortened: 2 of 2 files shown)


tomicapretto:
Thanks a lot for the contribution! The only thing I would add is that this changes the interpretation of the prior on the intercept, as it's on the intercept after the data was centered.

jt-lab:
> Thanks a lot for the contribution! The only thing I would add is that this changes the interpretation of the prior on the intercept, as it's on the intercept after the data was centered.

Added! This was a good point. The zero-centered priors from some examples were  actually how we stumbled on the data being automatically centered in the first place.

tomicapretto:
@jt-lab CI is failing because of `black`. Can you fix it? If you can't, I can do it for you :) 

Reference: https://github.com/bambinos/bambi/actions/runs/6273481237/job/17037212920?pr=726

jt-lab:
> @jt-lab CI is failing because of `black`. Can you fix it? If you can't, I can do it for you :)
> 
> Reference: https://github.com/bambinos/bambi/actions/runs/6273481237/job/17037212920?pr=726

I'll fix it. At least I'll try! I thought I couldn't break anything just changing a docstring... but I was wrong! :-)

jt-lab:
> > @jt-lab CI is failing because of `black`. Can you fix it? If you can't, I can do it for you :)
> > Reference: https://github.com/bambinos/bambi/actions/runs/6273481237/job/17037212920?pr=726
> 
> I'll fix it. At least I'll try! I thought I couldn't break anything just changing a docstring... but I was wrong! :-)

BTW: When running the tests I got these seemingly unrelated errors:
FAILED tests/test_hsgp.py::test_minimal_1d_fits - ValueError: Incompatible Elemwise input shapes [(100,), (10,)]
FAILED tests/test_hsgp.py::test_minimal_1d_predicts - ValueError: Incompatible Elemwise input shapes [(100,), (10,)]

tomicapretto:
@jt-lab I just saw those errors too. I think it's due to some change upstream. I'll merge this because this is an unrelated change. We'll have to figure the problem out in a separate PR (or maybe it's fixed upstream). Thanks for the contribution! :smile:

PR #728, opened 2023-09-24
Add kidney.csv to test data

donaldbraman:
Adding the kidney.csv (converted from kidney.rda from brms) to the test data folder.


tomicapretto:
Hi @donaldbraman! Thanks for the contribution.

However, I don't understand why we need to have the `kidney.csv` file in our repository. If you do `bmb.load_data("kidney")` it just works.

donaldbraman:
Oh, that's strange. It didn't work for me, which is why I created the pull request. But if it already exists in the package somewhere, apologies for the fake out!

tomicapretto:
> Oh, that's strange. It didn't work for me, which is why I created the pull request. But if it already exists in the package somewhere, apologies for the fake out!

Before closing, does it work on your end?

```python
import bambi as bmb
bmb.clear_data_home()
bmb.load_data("kidney")
```

donaldbraman:
That doesn't work for me.  [Here's my notebook that tests it.](https://colab.research.google.com/drive/1MKR0DaazJRG2jaVcxZq2JUdk64S0PCRS)

donaldbraman:
Looking a little more, I see that it's not included in the DATASETS listed in the datasets.py used in the bambi I pip installed, which is obviously not the latest. Closing out now.

PR #731, opened 2023-09-28
Bug fix for plotting order in plot_comparison (fixes mismatch between labels and plot)

jt-lab:
A call like ...
```
bmb.interpret.plot_comparisons(
        model=model, 
        idata=trace, 
        contrast={"factor1": ["X", "Y"]},
        conditional={"factor2": ["C", "A", "B"], "factor3": [1, 2]},
)
```
currently leads to a mismatch of the order with which the levels of factor2 are plotted on the x axis and the x axis labels!

I have traced this down to .values.dtype.categories returning the names in sorted order which is eventually used for the labels. The actual plot is based on the user provided order. 

This commit is only a quick fix which sorts the user-provided levels to match the order to avoid any false results. It would be better to give the user control about the order by using the order of the user provided list for the labels but a change like that seems to interact with further code which I don't fully understand. 

Moreover, I'm not entirely sure what the benefit of the call above is compared to:
```
bmb.interpret.plot_comparisons(
        model=model, 
        idata=trace, 
        contrast={"factor1": ["X", "Y"]},
        conditional=["factor2", "factor3"],
)
```
I thought it was to give the user control over which levels are plotted (and in which order). But removing levels from the list does lead to errors (and the order is not applied). I might miss something, but if there is no additional functionality comapred to the list version, perhaps the dict version should be removed (for now).

I'm sorry for the many incremental commits just for fixing spaces and typos. I could not work with a local version but make change in githubs web interface (and test the code on a server).


tomicapretto:
@jt-lab thanks for the contribution! Can I ask for an example of how this looked before and how it looks after the change? Thanks!

Bonus: If it comes with data so I can test on my end, even better!

jt-lab:
> @jt-lab thanks for the contribution! Can I ask for an example of how this looked before and how it looks after the change? Thanks!
> 
> Bonus: If it comes with data so I can test on my end, even better!

Here you go: 
Code (note the two different orders of the factors in the conditional argument of plot_comparisons):
```
import pandas as pd
import bambi as bmb
from matplotlib.pylab import plt

df = pd.read_csv('simulated_data_order_prob.csv')

model = bmb.Model('p(correct, count) ~ factor1*factor2*factor3 + (factor3+factor2|individual)',
                  df, family='binomial')
trace = model.fit(tune=500, draws=1000, target_accept=0.7, init='adapt_diag')

f, ax = plt.subplots(2)
bmb.interpret.plot_comparisons(
    model=model,
    idata=trace,
    contrast={"factor2": ["X", "Y"]},
    conditional={"factor1": ["A", "B", "C"], "factor3": [1, 2]},
    ax=ax[0]
) 
bmb.interpret.plot_comparisons(
    model=model,
    idata=trace,
    contrast={"factor2": ["X", "Y"]},
    conditional={"factor1": ["C", "B", "A"], "factor3": [1, 2]},
    ax=ax[1]
)
```
Plot before the fix:
(e.g.: git+https://github.com/jt-lab/bambi.git@a072c84809fe62e6efe1d7818f3d5d28d81e00ed)
before

After the [fix:](url)
(e.g.: git+https://github.com/jt-lab/bambi.git)
after

Data set:
[simulated_data_order_prob.csv](https://github.com/bambinos/bambi/files/12752326/simulated_data_order_prob.csv)
This is the same simulated data I also posted in question on the pymc discourse. I just added a further level for factor1 and matched the factor and level names with my original description of the problem.

GStechschulte:
@jt-lab thanks for finding the subtle bug and opening a PR (good attention to detail)👍🏼. The difference between 

```python3
# user provided values only (no default values are computed for "factor2" and "factor3")
bmb.interpret.plot_comparisons(
        model=model, 
        idata=trace, 
        contrast={"factor1": ["X", "Y"]},
        conditional={"factor2": ["C", "A", "B"], "factor3": [1, 2]},
)
```
and
```python3
# computes default values for "factor2" and "factor3"
bmb.interpret.plot_comparisons(
        model=model, 
        idata=trace, 
        contrast={"factor1": ["X", "Y"]},
        conditional=["factor2", "factor3"],
)
```
is that in the latter code snippet, since no values are passed for `factor2` and `factor3`, default values are computed for these covariates. In the _Default contrast and conditional values_ section of the [plot_comparisons](https://bambinos.github.io/bambi/notebooks/plot_comparisons.html#default-contrast-and-conditional-values) docs, these details are outlined. 

Thanks!

jt-lab:
@GStechschulte, thanks for the explantions! I'm not sure I understand the purpose of the dictionary conditional  argument. When I list all levels of the factors, I get the same as with the list version defaults. When I list a subset of the levels, I get errors. 

Unrelated: When reading through the plot_comparisons doc, I noticed this sentence "Bambi uses the ordering (keys if dict and elements if list)...". Im not sure I understand correctly, but dictionaries don't gurantee any order of the items or keys. So this might also be a source of unexpected behavior. So the user might need to pass an OrderedDict to be safe. But maybe i'm missing something!

Many thanks again!

GStechschulte:
> I'm not sure I understand the purpose of the dictionary conditional argument. When I list all levels of the factors, I get the same as with the list version defaults

This is due to the data types of your data being used to fit the model. 

When you pass the list `["factor1", "factor3"]` for the conditional arg., `factor1` is assigned as the "main" covariate and `factor3` is assigned the "group" covariate.  Then, since you don't specify any values to be associated with the factor covariates, a default value is computed for them. Default values for "main", "group", and "panel" are the following pseudocode:

```python3
if v == "main":
    
    if v == numeric:
        return np.linspace(v.min(), v.max(), 50)
    elif v == categorical:
        return np.unique(v)

elif v == "group":
    
    if v == numeric:
        return np.quantile(v, np.linspace(0, 1, 5))
    elif v == categorical:
        return np.unique(v)

elif v == "panel":
    
    if v == numeric:
        return np.quantile(v, np.linspace(0, 1, 5))
    elif v == categorical:
        return np.unique(v)
```

In your case, "main" and "group" are both categorical data types. Thus, the default values are the unique values. 

When you pass a dict to the conditional arg., the keys of the dict are taken as the "main", "group", and "panel" covariates. Thus, `{"factor2": ["C", "A", "B"], "factor3": [1, 2]}` corresponds to `{"main": "factor2", "group": "factor3"}`. Since you passed values for the covariates, no default values are computed. 

Then, since you ended up passing all the unique values for `factor2`, and `factor3` as the dict values, the result of passing a list to conditional and the passing of the dict (stated above) results in the same data being generated. If you were to pass `{"factor2": ["A", "B"], "factor3": [1]}`, you would get a different result between the two.

I hope this helps 👍🏼

GStechschulte:
Should I make a separate PR incorporating @jt-lab commits and my local changes to fix the bugs? Or should @jt-lab keep this PR and then commit my changes to fix the bugs? @tomicapretto

tomicapretto:
@GStechschulte if your fixes are related to what @jt-lab is doing here, I think @jt-lab could allow this PR to be edited by maintainers and then you could contribute here so everything is in place and we respect the original authorship. But if they are separate things, you could open a different PR.

But if you think that's too much, feel free to do something else.

jt-lab:
@tomicapretto, how can I make it editable?

tomicapretto:
I think it's already enabled

GStechschulte:
I have pushed 3 commits to your branch. Now, the plots work with no errors. I also decided that your original commit to sort the dict values in ascending order is the most appropriate. In particular, the x-axis of graphs should be in ascending order $1, 2, 3,...,n$ or $A, B, C,...,n$.

Thanks a lot for raising the issue and for opening a PR 😄 It is much appreciated! 

Below are a few of your examples:

```python3
f, ax = plt.subplots(2)
bmb.interpret.plot_comparisons(
    model=model,
    idata=trace,
    contrast={"factor2": ["X", "Y"]},
    conditional={"factor1": ["A", "B", "C"], "factor3": [1, 2]},
    ax=ax[0]
) 
bmb.interpret.plot_comparisons(
    model=model,
    idata=trace,
    contrast={"factor2": ["X", "Y"]},
    conditional={"factor1": ["C", "B", "A"], "factor3": [1, 2]},
    ax=ax[1]
)
plt.tight_layout();
```
image

```python3
bmb.interpret.plot_comparisons(
    model=model,
    idata=trace,
    contrast={"factor2": ["X", "Y"]},
    conditional={"factor1": ["A", "C"], "factor3": [1, 2]},
    fig_kwargs=dict(figsize=(7, 3))
);
```
image

```python3
bmb.interpret.plot_comparisons(
    model=model,
    idata=trace,
    contrast={"factor2": ["X", "Y"]},
    conditional=["factor1", "factor3"],
    fig_kwargs=dict(figsize=(7, 3))
);
```
image

tomicapretto:
@GStechschulte @jt-lab ok I struggled a little bit with git and the fact that we're working on @jt-lab's main branch (I used a different name locally). Nevertheless, it seems everything is working. I just made a small modification to the if-else structure, I think it's clearer now, and I also wrote a fix for the HSGP and the failing tests. We can merge once CI passes.

tomicapretto (approved):

GStechschulte:
@tomicapretto Ahh great! And thanks for the HSGP fix 👍🏼

GStechschulte:
Mmmm. The `test_plots.py` passed locally before I pushed those three commits (Tomas's commits shouldn't cause them to fail). I will look into this evening.

tomicapretto:
> Mmmm. The `test_plots.py` passed locally before I pushed those three commits (Tomas's commits shouldn't cause them to fail). I will look into this evening.

The problem happens when we test this

```python
plot_slopes(model, idata, wrt={"Days": 2}, conditional={"Subject": 308}
```

and it's because it's calling `sorted(308)`, which cannot be sorted because it's not iterable. I'm fixing it now :)

PR #732, opened 2023-09-28
`interpret` support for model predictions with response levels

### LINKED ISSUE #723: `interpret` value errors for ordinal and categorical models
`interpret` does not work with models (ordinal and categorical regression) whose predictions are a vector of some quantity, e.g., probabilities. This is because each value in the vector of probabilities is associated with a category / class. When the model's predictions are assigned to the summary dataframe, a value error is raised as the summary dataframe is expecting one prediction for each row. 

For example:

```python3
length = [
    2.72, 2.79, 2.84
]
choice = [
]

    ordered=True
)

model = bmb.Model("choice ~ length + sex", data, family="categorical")
idata = model.fit(
    draws=1000, target_accept=0.95, random_seed=1234, chains=4
)

bmb.interpret.predictions(
    model=model,
    idata=idata,
    covariates="length"
)
```
```
ValueError: Expected a 1D array, got an array with shape (50, 3)
```

The `ValueError` is raised due to the 3 classes. For each sample in the new dataframe fed to the model, the model predicts a probability for each class. We could convert the xarray into a dataframe and perform a left join on this to ensure the data used to perform predictions is joined correctly with the predictions and uncertainty intervals. 

In the example of `bmb.interpret.predictions` I envision the summary dataframe looking something like this

length | sex | choice_dim | estimate
-- | -- | -- | --
1.240000 | Male | Other | 0.097136
1.240000 | Male | Invertebrates | 0.638881
1.240000 | Male | Fish | 0.263982
... | ... | ... | ...
3.890000 | Male | Other | 0.316271
3.890000 | Male | Invertebrates | 0.002594
3.890000 | Male | Fish | 0.681135

where each row (sample) has been duplicated two times due to the left join. The user can now analyse the predicted probabilities for each class when `length = 1.24` and `sex = Male`. This summary df would then be passed to the `plot_predictions`.

GStechschulte:
This PR resolves #723 and will allow the sub-package `interpret` to work with models, such as ordinal and categorical regression, whose predictions are a vector of some quantity, e.g., probabilities. 

If a model's prediction is a vector of some quantity, then a series of joins are performed on the mean predictions `y_hat_mean`, uncertainty interval `bounds`, and data used to compute predictions (e.g., `cap_data`). Ultimately, the left join is used to ensure the data used to compute predictions is duplicated correctly with the predictions and uncertainty intervals.

Here is a [link](https://gist.github.com/GStechschulte/82b77453312b34b355fb86773faac006) to a Gist demonstrating the bug fix on `bmb.interpret.plot_predictions` for categorical and ordinal regression.

To do:
- [x] support for `plot_predictions`
- [x] support for predictive differences (`plot_comparisons` and `plot_slopes`)
- [x] add tests
- [x] run Pylint
- [x] run black


GStechschulte:
The [gist](https://gist.github.com/GStechschulte/82b77453312b34b355fb86773faac006) has been updated with demos for `comparisons`, `slopes`, and `predictions`.

tomicapretto:
I'm trying to replicate the plot [here](https://bambinos.github.io/bambi/notebooks/categorical_regression.html#using-numerical-and-categorical-predictors)

image


The first thing I do is the following:

```python
plot_predictions(model, idata, ["length"]);
```

which results in:

image

The problem is when I try to map `"sex"` to the panel variable.

```python
plot_predictions(model, idata, {"horizontal": "length", "panel": "sex"});
```

which results in the following error:

```
TypeError: covariates must be a string or a list of strings.
```

And if I do the following: 

```python
plot_predictions(model, idata, ["length", "sex"]);
# or
plot_predictions(model, idata, ["length", "sex", "sex"]);
```

it returns a weird result because it's mapping "sex" to color as well.

image
image

And finally I tried

```python
plot_predictions(model, idata, ["length", "choice", "sex"]);
```

but the result doesn't look good either

image

So, **my question is:** why did we remove the ability to pass a dictionary to `covariates`? And on top of that, do you think it should be possible to map another attribute to the color apart from the response level? My initial approach would be not to allow it, but if you think it's not hard, go ahead.

GStechschulte:
We removed the ability in the plot comparisons PR #684 and I described why we should remove this ability in [this comment](https://github.com/bambinos/bambi/pull/684#issuecomment-1592407602).

> do you think it should be possible to map another attribute to the color apart from the response level

It is possible without needing to code anything. It is subtle, but since the plot functions use the columns from the summary dataframe, and if the user knows what those columns are a priori, then they can use whatever column they feel necessary for main, group, and panel.

tomicapretto:
@GStechschulte ha! excellent! 

Do you want to modify the example to use the new functionality? After that, I think it can be merged.

Thanks a lot!

GStechschulte:
For sure, I will add it. There's also a couple inline comments I want to add, and then it can be merged. Thanks! 👍🏼

PR #733, opened 2023-09-29
Multiple doc notebooks

GStechschulte:
This PR removes duplicate notebooks in the `gallery.yml` file which causes the [docs](https://bambinos.github.io/bambi/notebooks/) to render duplicate notebooks. Additionally, I resolved some grammar mistakes in the ordinal and zero inflated regression notebooks.


tomicapretto:
Thanks!

PR #742, opened 2023-10-21
Survival docs

GStechschulte:
This PR adds a notebook explaining how to perform survival analysis in Bambi using the cat adoption dataset from chapter 11.4 of Richard McElreath's Statistical Rethinking. Sections include:

- explanation of survival and censoring data (including what right and left censoring is)
- survival functions and popular estimators (emphasis on Kaplan-Meirer estimator)
- regression models with survival responses and how the hazard rate and cox proportional hazards facilitate such an analysis
- explanation of the `pm.Censored` distribution with and without predictors
- two Bambi models (with and without predictors)

To do:
- [x] finish the _Regression models with survival responses_ section
- [x] add summary section 

Although there are a couple "to do" items, this docs example is far enough along I am opening this draft PR for feedback. Thanks! 🐱


docs/notebooks/survival_model.ipynb
```diff
@@ -0,0 +1,1412 @@
+{
```
GStechschulte, line 1: <p>Before <code>cat_model.predict(idata, data=new_data, kind="mean")</code> is performed, the only data variable in the inference data posterior is <code>color_id</code>. However, after performing predictions on new data (or on the observed data), another data variable <code>censored(days_to_event / 31, adopt)</code> is populated. </p><p> </p><p><code>censored(days_to_event / 31, adopt)</code> is on the scale of $Y$ (in this example months). <code>color_id</code> is on the rate scale $lambda$. However, $lambda$ can also be calculated from <code>censored(days_to_event / 31, adopt)</code> by doing <code>1 / censored(days_to_event / 31, adopt) </code>. </p><p> </p><p>I am pointing this out because there is a "large" difference in the survival curves.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/742/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

tomicapretto: <p>First I didn't understand what you were trying to do. Then I got worried. Finally, I realized we were missing the link function!  Do what's below, and see it matches the posterior draws of <code>color_id</code>  <span class="ql-cursor">﻿</span></p><p> </p><pre class="ql-syntax" spellcheck="false">np.log(idata["posterior"]["censored(days_to_event / 31, adopt)_mean"])
</pre>

docs/notebooks/survival_model.ipynb
```diff
@@ -0,0 +1,1196 @@
...
```
tomicapretto, line 1: <p>Typo in "Censored individuas".</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/742/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

tomicapretto, line 1: <p>Typo: sensoring -&gt; censoring</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/742/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

tomicapretto, line 1: <p>Is the horizontal axis in days? I find the results here not consistent with the results below. Somewhere I saw you said you didn't know if you were computing the K-M estimator correctly. Maybe you can use <code>KaplanMeierFitter</code> from the <code>lifelines</code> package?  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/742/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

tomicapretto, line 1: <p>The text is excellent so far. The only thing I want to point is that you have three different expressions where you write the same on the left, but something different on the right. I think that should be changed.</p><p> </p><p>**Edit** a minor comment is that I don't like talking about "Censored" as a distribution because it's not. It's an utility that helps you to create censored distributions. But since PyMC calls it a distribution, let's leave it as it is.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/742/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

tomicapretto, line 1: <p>I would add one or two sentences about the <code>censored()</code> function, which you can find here <a href="https://github.com/bambinos/bambi/blob/65bbae4faa70461a2b3983a72965b4fb337de3d9/bambi/transformations.py#L12" target="_blank">https://github.com/bambinos/bambi/blob/65bbae4faa70461a2b3983a72965b4fb337de3d9/bambi/transformations.py#L12</a>. Basically, I would say it takes two arguments. The first is the observed value, and the second is "left", "none", or "right" indicating left, none, or right censoring. I wouldn't get into interval censoring as it doesn't work very well out of the box in PyMC. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/742/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

tomicapretto, line 1: <p>I would add that if we wanted we could derive the survival function, but we'll use scipy because it's already there. I'm saying this because I found it surprising when I saw <code>scipy.stats.sf</code> in the text. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/742/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

tomicapretto, line 1: <p>I think showing a normal likelihood here and using an exponential likelihood below can be confusing. Maybe I'm missing something, let me know</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/742/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

tomicapretto, line 1: <p>Nit: add some transparency</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bambinos/bambi/pull/742/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

tomicapretto:
As usual this is an amazing tutorial. Feel free to merge after you address the comments (or ping me if you need help!)

docs/notebooks/survival_model.ipynb
GStechschulte: <p>Yeah that's a good point. I updated so PDF uses $f(D_i | \lambda_i)$, CDF uses $F(D_i | \lambda_I)$, and the CCDF uses $S(D_i|\lambda) = \text{exp}(-\lambda_i D_I)$. CCDF uses $S$ because the survival function can be written as the CCDF.</p>

GStechschulte: <p>Definitely agree. Thanks!</p><p> </p>

GStechschulte: <p>After second thought, I agree with that. In the first pass, I wanted to implicitly show you didn't have to pass <code>exponential</code> as the likelihood. But it is better to stay consistent with the cat model. </p><p> </p>

GStechschulte: 🙈

GStechschulte: If I wanted to add a noise term at the observation level to account for variability not explained by $\mu$ (the deterministic part), would that be added to $\lambda$? For example,

$$\alpha \sim \mathcal{N}(0, 1)$$
$$\beta \sim \mathcal{N}(0, 1)$$
$$\epsilon \sim \mathcal{N}(0, 1)$$
$$\mu = \alpha + \beta X$$
$$\lambda \sim \text{Exponential}(\mu)$$
$$Y \sim \lambda + \epsilon$$

docs/notebooks/survival_model.ipynb
GStechschulte: Indeed, I was missing that!

GStechschulte:
> As usual this is an amazing tutorial. Feel free to merge after you address the comments (or ping me if you need help!)

Thanks a lot!! Especially for the review. My only question is regarding the noise term (commented above).

docs/notebooks/survival_model.ipynb
GStechschulte: Yeah I will give the lifelines package a try. I initially wanted to avoid the external dependency.

tomicapretto: Good question! Both in Normal and Exponential likelihoods the noise term in the linear predictor is absordeb into the likelihood. 

In the normal case, you say epsilon ~ Normal(0, sigma), consequently, y = mu + epsilon ~ Normal(mu, sigma). Something similar occurs in the Exponential case. So you just need to write

alpha ~ N
beta ~ N
mu = exp(alpha + beta * X)
lambda = 1 / mu
Y ~ Exponential(lambda)

tomicapretto: I like avoiding external dependencies, but it's not too bad for an example

GStechschulte: Ahh okay. Thanks!!

GStechschulte: My KM estimator was wrong :( I will use the `lifelines` estimator in the docs.

Issue #740, opened 2023-10-20
Make `interpret` module verbose about default computations


tomicapretto:
Some combination of inputs results in the `interpret` module having to choose the value of some predictor, the aggregation to apply, etc. This is documented but not all users read the documentation. As suggested in https://github.com/bambinos/bambi/issues/735#issuecomment-1759800403, it would be good to be more verbose.

Of course, it has to be possible to turn it off. For that, see https://github.com/bambinos/bambi/issues/735#issuecomment-1767459932

PR #745, opened 2023-10-26
`interpret` verbosity using logging

GStechschulte:
Resolves PR #740. I am opening as a draft PR to collect feedback on my initial idea and design. 

In order to make `interpret` verbose about some of the "hidden" computations such as default values, I am creating a config instance (similar to [formulae](https://github.com/bambinos/formulae/blob/master/formulae/config.py)) that allows the user to optionally turn on or off the logging of various information. To log information to the console, I want to use the built in `logging`  module to create a logger object specific to the `interpret` module.

Then, I create a `log_interpret_functions` decorator so I can simply decorate the desired functions in `utils.py` that compute default values. This allows me to extend the functionality of the existing util functions without changing the underlying code, and thereby separating the logging code into a different file `logs.py`. 

Below is a demo:

```python3
import numpy as np
import pandas as pd

import bambi as bmb

# logger is the config instance and we set messages to True (default is False)
bmb.interpret.logger.messages = True

data = bmb.load_data("mtcars")
data["am"] = pd.Categorical(data["am"], categories=[0, 1], ordered=True)
model = bmb.Model("mpg ~ hp * drat * am", data)
idata = model.fit(tune=250, draws=250, random_seed=1234)
``` 

Calling `predictions`:

```python3
# model includes hp, am, drat so defaults computed for am and drat
df = bmb.interpret.predictions(
    model,
    idata,
    conditional={
        "hp": [100, 150]
    }
)
```
```
INFO:root:Default values computed for: {'am', 'drat'}
```

Calling `plot_predictions`

```python3
# model includes hp, am, drat so defaults computed for am
bmb.interpret.plot_predictions(
    model,
    idata,
    conditional={
        "hp": [100, 150],
        "drat": [2.5, 3.5]
    }
);
```
```
INFO:root:Default values computed for: {'am'}
```

Calling `predictions` where no default values are computed means no messages.

```python3
df = bmb.interpret.predictions(
    model,
    idata,
    conditional={
        "hp": [100, 150],
        "drat": [2.5, 3.5],
        "am": [0, 1],
    }
)
```

I need to make sure the logger object I am creating and referencing is not messing with the `"bambi"` logger in `bambi/__init__.py` (which I think it is because of the `INFO:root:` tag in the console logs above). Using logging in multiple modules should be possible according to the [logging cookbook](https://docs.python.org/3/howto/logging-cookbook.html#logging-cookbook). Thus, Bambi would have two loggers: one for the main module `bambi`, and then one for `interpret`.

Before I move forward with this implementation, I want to get feedback from the community and or anyone with more experience in logging. Thanks!


bambi/interpret/logs.py
```diff
@@ -0,0 +1,32 @@
+import logging
+
+from bambi.interpret import logger
+
+
+def log_interpret_defaults(func):
+    interpret_logger = logger.get_logger()
```
tomicapretto, line 7: If I recall correctly, you need to pass `getLogger` a name to get the same logger. As it's done here https://github.com/bambinos/bambi/blob/d001e069620a494bde730d69b47fb4c5ad3ba3e6/bambi/__init__.py#L30.

So you could do `logger.get_logger("bambi_interpret")`

bambi/interpret/logs.py
```diff
@@ -0,0 +1,32 @@
...
+    interpret_logger.setLevel(logging.INFO)
```
tomicapretto, line 8: I also think you need this

https://github.com/bambinos/bambi/blob/d001e069620a494bde730d69b47fb4c5ad3ba3e6/bambi/__init__.py#L32-L36

I don't remember the details of when it was added. But I think it makes sure the messages are sent once to the place where you want to send them.

bambi/interpret/logs.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def wrapper(*args, **kwargs):
+        if not logger.messages:
+            return func(*args, **kwargs)
+
+        if func.__name__ == "set_default_values":
```
tomicapretto, line 14: I'm not sure if we should handle specific cases here. I think it would be better to allow the decorator to receive arguments and then determine what to do from those. 

Also, in this case, the function receives a dictionary and returns a dictionary. So you compare keys. Are there any other cases we may be interested?

tomicapretto:
I didn't have the decorator idea in my mind when we discussed logging. But I do think it's a very neat idea. I'm not 100% sure if it generalizes well. Maybe, it's enough for it to generalize to the cases we care about. Do we know if we're going to decorate functions with different a wider variety of input/output types?

bambi/interpret/logs.py
GStechschulte: That's correct. When I did this, it wasn't returning the info logs to the console. Once I removed the name, it returned the info logs.

bambi/interpret/logs.py
GStechschulte: > I'm not sure if we should handle specific cases here. I think it would be better to allow the decorator to receive arguments and then determine what to do from those.

I agree. That would generalize better.

> Also, in this case, the function receives a dictionary and returns a dictionary. So you compare keys. Are there any other cases we may be interested?

There are two functions that compute default values. Both in `utils.py`.

1. [set_default_values](https://github.com/bambinos/bambi/blob/d001e069620a494bde730d69b47fb4c5ad3ba3e6/bambi/interpret/utils.py#L336) that computes default values for covariates specified in the model, but not passed to `conditional`.
2. [set_default_variable_values](https://github.com/bambinos/bambi/blob/d001e069620a494bde730d69b47fb4c5ad3ba3e6/bambi/interpret/utils.py#L89) that computes default `contrast` or `wrt` values for `comparisons` and `slopes`, respectively if the user does not pass a value (but they obviously need to pass the variable name of interest). 

These are the functions that I initially wanted to decorate. However, in other scenarios, e.g. a user passes a list of covariate names to `conditional`, but no values, then the following functions are called to compute default values:
1. [make_main_values](https://github.com/bambinos/bambi/blob/d001e069620a494bde730d69b47fb4c5ad3ba3e6/bambi/interpret/utils.py#L366) that computes main values based on original data using a grid of evenly spaced values for numeric predictors and unique levels for categoric predictors.
2. [make_group_values](https://github.com/bambinos/bambi/blob/d001e069620a494bde730d69b47fb4c5ad3ba3e6/bambi/interpret/utils.py#L378) that computes group values based on original data using unique levels for categoric predictors and quantiles for numeric predictors.

Knowing that a `np.linspace` or `np.quantile` would also be informative for the user. I will decorate these functions also since they compute default values as well.

GStechschulte:
> I didn't have the decorator idea in my mind when we discussed logging. But I do think it's a very neat idea. I'm not 100% sure if it generalizes well. Maybe, it's enough for it to generalize to the cases we care about. Do we know if we're going to decorate functions with different a wider variety of input/output types?

That is also what I am questioning myself on after replying to your comment above with the list of functions that compute default values. Most function parameters accept an array and return an array.

`set_default_values(dict) -> dict`
`set_default_variable_values(self) -> array` (this is a method in `class VariableInfo`)
`make_main_values(array) -> array`
`make_group_values(array) -> array`

Unless I added parameters to the decorator or had different decorators for different functions, then I don't see how it will generalize. I am open to other ideas. I don't have too much time sunk into this PR. hah

bambi/interpret/logs.py
tomicapretto: Sounds great

GStechschulte:
Based on the discussion above, the following functions that compute default values

- `set_default_values`
- `set_default_variable_values`
- `make_main_values`
- `make_group_panel_values`: this is decorated instead of `make_group_values` as this function is called within `make_group_panel_values `

now have a decorator `@log_interpret_defaults` that log a message(s) to console if the user sets `bmb.interpret.logger.messages = True`. 

I decided to keep the handling of specific cases, i.e., `if func.__name__ == <some func that computes default values>` as this allows more detailed messages to be logged as the logger is "aware" if a default is computed for the contrast or wrt variable, or conditional covariates. For example (using the model from above):

```python3
bmb.interpret.comparisons(
    model,
    idata,
    contrast="hp",
    conditional="drat"
)
```
```
Default computed for contrast variable: hp
Default computed for main variable: drat
Default computed for unspecified variable: am
```

I am open to changing the specific cases if anyone has a better design idea. I have also added tests `test_interpret_messages.py` that tests `comparisons`, `predictions`, and `slopes` for different scenarios of default computations and the expected log messages.

tomicapretto:
You successfully implemented a solution and I think that's great. As far as I understand, the decorator doesn't have arguments that determine how to behave with that function. Instead, the decorator "knows" about some functions, and the functionality for those is hardcoded. I think it's fine and I wouldn't invest a lot of effort in modifying it.

The only thing I want to raise is that I'm not sure why we need an `InterpretLogger` class. My points are the following

* Its only method, `get_logger()`, is the same as doing `logging.getLogger("interpret")`. I would replace it with `logging.getLogger("bambi-interpret")` which identifies the logger in a more unique manner. All the initialization stuff related to the handler could be handled in `bambi/interpret/__init__.py` in a similar way to the logger we have in `bambi/__init__.py`.
* It also provides configuration facilities, the `messages` attributes. But not much is guaranteed. I could set it to whatever I wanted and Bambi will not complain. This is Python and it's very flexible. But as this is a configuration variable, I think it would be nice to handle this better. What I have in mind for this configuration variable is something like the configuration in formulae, which is the single place for configuration variables in formulae and it also checks users pass values that are allowed. https://github.com/bambinos/formulae/blob/master/formulae/config.py


Regarding these last two points, I'm more than happy to add an implementation to this existing PR as you have already done a lot of effort @GStechschulte, just let me know what you think

GStechschulte:
Thank you @tomicapretto, and your interpretation is correct. The two points you bring up are great, and I think they would both enhance this PR. Since you have already implemented this in formulae, and if you don't think it will take you long, that would be greatly appreciated 😄

tomicapretto (approved):
Feels weird to be approving my own changes haha. @GStechschulte I think this is good to go. Let me know if you have questions or suggestions about the changes.

GStechschulte:
> Feels weird to be approving my own changes haha. @GStechschulte I think this is good to go. Let me know if you have questions or suggestions about the changes.

LGTM 👍🏼  Thanks a lot! I was testing it with some of our docs notebooks and it works great. However, the `INTERPRET_VERBOSITY` field is set to `True` by default. Do we want to keep this as the default, or set to False?

I am updating docs related to `interpret` to reflect this feature request. Then, I will merge.

tomicapretto:
@GStechschulte i would set it to True by default. Otherwise, I think people not familiar with how it works will not realize this is happening. This way, the message will be a signal telling something is happening.

GStechschulte:
@tomicapretto I updated the three notebooks related to `interpret`  to inform the user the purpose of this logging and how to turn it off. I will merge once the checks are ✅

PR #752, opened 2023-11-10
Implement truncated responses

tomicapretto:
This PR adds a function `truncated()` that can be used in the formula syntax to specify truncated variables.

Example (taken from [here](https://www.pymc.io/projects/examples/en/latest/generalized_linear_models/GLM-truncated-censored-regression.html))

```python
import bambi as bmb
import numpy as np
import pandas as pd

rng = np.random.default_rng(12345)
slope, intercept, σ, N = 1, 0, 2, 200
x = rng.uniform(-10, 10, N)
y = rng.normal(loc=slope * x + intercept, scale=σ)

def truncate_y(x, y, bounds):
    keep = (y >= bounds[0]) & (y <= bounds[1])
    return (x[keep], y[keep])

bounds = [-5, 5]
xt, yt = truncate_y(x, y, bounds)

df = pd.DataFrame({"x": xt, "y": yt})
priors = {
    "Intercept": bmb.Prior("Normal", mu=0, sigma=1), 
    "x": bmb.Prior("Normal", mu=0, sigma=1),
    "sigma": bmb.Prior("HalfNormal", sigma=1)
}
model = bmb.Model("truncated(y, -5, 5) ~ x", df, priors=priors)
```

**NOTE** I still need to implement tests


GStechschulte (approved):
Looks great! 👍🏼

Issue #753, opened 2023-11-10
Improve tests to reduce durations


tomicapretto:
The current tests take a lot to run, between 30 and 35 minutes in the CI.

One of the reasons is that we build **many** PyMC models and in most of the cases we also test we can sample from the posterior. What makes it worse is that we have many duplicated tests. Or, put it in another way, there are many tests that could be merged.

For example, the module `tests/test_built_models.py` contains many tests where we verify we can build and sample with different model families. Then, `tests/test_predict.py` rebuilds many of these models to test whether it's possible to compute predictions in various scenarios. 

I think one low-hanging fruit is to "merge" tests where we use the same (or very similar) models. This way, we only have to compile the logp and sample from the posterior once.

Maybe this idea also opens the door to the construction of a more structured class where we test each family following a pre-specified set of steps. I'll try to find time to work on it.

PR #757, opened 2023-11-14
Improve tests

tomicapretto:
As I wrote in #753, our tests were taking too much time to run. One of the reasons is that we had a `test_predict.py` file where we wanted to test whether we could compute predictions for a variety of models, but we ended up rebuilding and refitting many models already tested in `test_built_models.py`.

In this PR I'm moving and merging many tests from `test_predict.py` and `test_built_model.py` into `test_models.py`. There's a class for every model family. There we implement as many methods as needed. The goal is to avoid building and fitting the same model more than once.

I'm sure there's still room for improvement and reorganization. But I think this is now in a better state. And on top of that, I already invested a couple of days on this and I think it's more than enough :sweat_smile:. 

As soon as the run finishes on my end I will add the time it takes now. Before these changes, running the tests took 1709.07 seconds on my end (28:29 minutes)

**Edit** After the changes it took 949.76 seconds (15:49)


GStechschulte (approved):
This is great! Also good to have added tests for: alternative samplers, dropping NaNs, amongst others. 

I think this is good for now. But I know pytest also allows for a `conftest.py ` file that allows you to reuse fixtures in multiple test files. Perhaps in the future we could explore putting simulated data and built models in this file to share amongst the other test files. For example, any test functions for `interpret` requires loading / simulating of data and building models also.

GStechschulte:
@tomicapretto I am pretty sure I prematurely merged this into main. Even though all the checks passed, this branch is behind main by 2 commits (per viewing this PR's branch). I think we need to revert and rebase first.

GStechschulte:
> @tomicapretto I am pretty sure I prematurely merged this into main. Even though all the checks passed, this branch is behind main by 2 commits (per viewing this PR's branch). I think we need to revert and rebase first.

Did we ever take care of this 🙈

PR #761, opened 2023-11-24
Add support for weighted responses

### LINKED ISSUE #561: Add survey weights
I'm not sure it's going to be easy, but it would be a nice to have feature.

### LINKED ISSUE #639: Weights option
In BRMS we can give differents weights to each sample as shown below:

`brm(
mvbind(stflife, happy)|weights(anweight) ~ 1 + (1|(cntry)),
        data = X, chains = 2, cores = 2, control = list(max_treedepth = 15))`

Is it possible to do the same in Bambi?

tomicapretto:
With this PR Bambi supports weighted responses via the syntax `weighted(values, weights) ~ 1 + x + y + ...`.

Weights in statistics can mean many things. In this case, the weights multiply the contribution of the observation to the log-likelihood. Thus, if an observation has a weight of 3 it is as if there were three records like that one.

Closes #639. Closes #561.


tomicapretto:
Found room for improvement so I'm converting to draft

PR #765, opened 2023-11-30
Fixed typos, fixed one incomplete doc strings, etc.

jt-lab:
Fixed typos,  one incomplete doc strings (see formula.py l19; is this correct?), made docstring punctuation and emphasis (more) consistent:
Single line docstrings end with '.'
Multi-line part docstrings' first lines are "titles", without period at their ends
Further paragraphs end with '.'
All lines under 'parameters', 'returns' etc. end with '.'
This scheme was derived from the most common cases in the current version, and I have tried to consistently apply it where there were differences.


tomicapretto:
Hi @jt-lab thanks a lot for taking care of this! 

I _think_ double backticks (``) are not needed anymore since we're using Quarto and not sphinx for the docs. Could you adapt those cases to use single backticks? If you want, you can also try to build the documentation locally and see if it works. But if you can't or simply don't want, just let me know and I'll check it on my end. 

Thanks again!

tomicapretto:
> one incomplete doc strings (see formula.py l19; is this correct?)

yes that is correct :)

tomicapretto:
Hi @jt-lab, do you think you have time to incorporate my comments? I could do it for you if you want, so we get this merged :)

jt-lab:
Hi Tomás,

yes I can, I have done some of it already locally. Sorry for the delay. I'll be on a conference next week but after it I can finish it. How will it merge most smoothly? Can I pull the most recent version despite there already being a PR based on some old version?

PR #770, opened 2024-01-07
Fix legend in EDA for Hierarchical baseball demo

kddubey:
It looks like the legend labels in the EDA plot for the [Hierarchical baseball demo](https://bambinos.github.io/bambi/notebooks/hierarchical_binomial_bambi.html#hierarchical-model) are flipped.

Also, the current notebook explanation says that there were divergences when fitting the hierarchical model when there weren't any. This PR changes that explanation.


docs/notebooks/hierarchical_binomial_bambi.ipynb
```diff
@@ -354,11 +362,11 @@
     "# Create handles for the legend (just dots and labels)\n",
     "handles = [\n",
     "    Line2D(\n",
-    "        [0], [0], label=\"At Bat\", marker=\"o\", color=\"None\", markeredgewidth=0,\n",
+    "        [0], [0], label=\"Hits\", marker=\"o\", color=\"None\", markeredgewidth=0,\n",
     "        markerfacecolor=RED, markersize=12\n",
     "    ),\n",
     "    Line2D(\n",
-    "        [0], [0], label=\"Hits\", marker=\"o\", color=\"None\", markeredgewidth=0, \n",
+    "        [0], [0], label=\"At Bat\", marker=\"o\", color=\"None\", markeredgewidth=0, \n",
```
GStechschulte, line 369: If we change the label name, then we also need to change the ordering of the scatter plots above. I think this bug can be solved by changing the `color` kwarg in `Line2D` to appropriately match the `color` kwarg of the scatter plot.

GStechschulte (changes requested):
Thanks for the attention to detail on this (and opening the issue). I requested one change. Thanks 😄

docs/notebooks/hierarchical_binomial_bambi.ipynb
kddubey: The latest commit standardizes the order to be "Hits" and then "At Bat" so that the legend labels appear in the same order as the points

GStechschulte (approved):
Great, it looks good now. Thank you!

PR #773, opened 2024-01-23
interpret get name of LazyVariable 

### LINKED ISSUE #772: bambi.interpret.prediction fails for polynomial regression models
When doing polynomial regression, the `interpret.prediction` method can't seem to handle polynomial terms within the formula. A MWE:

```
np.random.seed(0)
x1 = np.random.normal(size=100)
x2 = np.random.normal(size=100)
y = 2 + 3*x1 + 1.5*x1**2 + 2*x2 + np.random.normal(scale=1, size=100)
data = pd.DataFrame({'x1': x1, "x2": x2, 'y': y})
model = bmb.Model('y ~ poly(x1, 2) + x2', data)
results = model.fit()
bmb.interpret.predictions(model, results, "x2")
```

Gives the following error: `AttributeError: 'LazyOperator' object has no attribute 'name'`

The same thing also happens if you do `y ~ I{x**2} + x` as the formula.

GStechschulte:
This PR resolves #772 to parse model components that are a formulae `LazyVariable` instance. These instances happen when a user calls a formulae stateful transformation function. In Bambi, there are really "two" function you can call on model terms:

1. A formulae stateful transformation, i.e. basis splines, polynomial
2. Function call from the environment, i.e. log transformation using numpy

I have added tests to the interpret module to reflect the two different types of function calls. I only added the tests for `bmb.interpert.predictions` since comparisons and slopes call the same function `get_model_covariates` where the bug was happening. So testing predictions implicitly tests the other two functions.

### DIFF
--- a/bambi/interpret/utils.py
+++ b/bambi/interpret/utils.py
@@ -4,9 +4,11 @@ from dataclasses import dataclass, field
 from typing import Union
 
 import numpy as np
-from formulae.terms.call import Call
 import pandas as pd
 import xarray as xr
+
+from formulae.terms.call import Call
+from formulae.terms.call_resolver import LazyVariable
 
 from bambi import Model
 from bambi.utils import listify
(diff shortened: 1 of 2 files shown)


tomicapretto:
Thanks a lot!

tjburch:
Thanks @GStechschulte!

PR #775, opened 2024-02-04
Use `bayeux` to access a wide range of samplers

GStechschulte:
I have been following @ColCarroll [bayeux](https://github.com/jax-ml/bayeux) library and thought it would be interesting to see how Bambi could incorporate it to offer the users a wide range of samplers (more than `nuts_blackjax` and `nuts_numpyro`).

_Edit_: Now I access the samplers programmatically using the `inference_method` arg. This removes previously needed code for `nuts_blackjax` and `nuts_numpyro`. If a user passes an MCMC inference method other than the PyMC MCMC sampler `mcmc`, Bambi will use bayeux to call that sampler. 

```python3
data = bmb.load_data("ANES")
clinton_data = data.loc[data["vote"].isin(["clinton", "trump"]), :]

model = bmb.Model("vote['clinton'] ~ party_id + party_id:age", clinton_data, family="bernoulli")
model.build()

idata = model.fit(inference_method="blackjax_hmc")
```

However, when cleaning the InferenceData, I am getting an xarray error 

```shell
"name": "ValueError",
	"message": "('chain', 'draw') must be a permuted list of FrozenMappingWarningOnValuesAccess({'chain': 8, 'draw': 500, 'party_id_dim_0': 2, 'party_id:age_dim_0': 3}), unless `...` is included"
```

It seems xarray is not liking something that bayeux is doing with the InferenceData.

Another thought is that using bayeux with Bambi is so easy 

```python
model = bmb.Model()
model.build()

bx_model = bx.Model.from_pymc(model.backend.model)
bx_model.<some sampling func>
```

that maybe we just add documentation explaining how to use bayeux with Bambi to avoid overhead from Bambi's side?

To Do:

- [x] add additional tests in `test_alternative_samplers.py`
- [x] update docstrings referencing JAX based samplers
- [x] add docs on using alternative backends


ColCarroll:
Really cool! A few suggestions -- 

- `bayeux` could be invisible here, and you could access all the methods programatically. That's done [here](https://github.com/jax-ml/bayeux/blob/main/bayeux/_src/bayeux.py#L101), but i can factor that out into a function that gives methods instead of strings -- currently it only adds a method if the underlying library, e.g. `optax`, is installed. i'm not sure how to avoid using a string at _some_ point. You could have an api like `model.fit.bx` that initializes the `bayeux.Model`?
- I'm happy to add a `from_bambi` constructor on the bayeux side to make your second option even easier.

ahartikainen:
What kind of API bayex has? Could we enable support for external samplers if we define specific API we support (need)? (Users could create class for external samplers if needed?)

Of course that does not mean we could not have a text based support on certain libraries?

ColCarroll:
`bayeux` is inspired by `arviz`, in that it just provides a representation of a model that is general enough for most samplers, but it does make the decision that it is specialized to JAX-based models (most of the algorithms use autodiff, vectorization is baked in, and automatic function inverses are also used). If you've got a sampler that accepts a JAX-based log density, you could use bayeux with it (or contribute it to bayeux!)

GStechschulte:
> Really cool! A few suggestions --
> 
> * `bayeux` could be invisible here, and you could access all the methods programatically. That's done [here](https://github.com/jax-ml/bayeux/blob/main/bayeux/_src/bayeux.py#L101), but i can factor that out into a function that gives methods instead of strings -- currently it only adds a method if the underlying library, e.g. `optax`, is installed. i'm not sure how to avoid using a string at _some_ point. You could have an api like `model.fit.bx` that initializes the `bayeux.Model`?
> * I'm happy to add a `from_bambi` constructor on the bayeux side to make your second option even easier.

Thanks for the suggestions! That makes sense. I am liking the second option, but I will run some ideas past the others first before asking for the feature. Thanks!

GStechschulte:
In this example, the error is because bayeux is appending `_0` to `party_id_dim` of the posterior dims. This results in Bambi discarding all posterior dims because the dims in the PyMC model are inconsistent with the dims of the InferenceData returned by bayeux. 

For example:

```python3
bayeux_idata.posterior.dims, pymc_idata.posterior.dims
```
```
(FrozenMappingWarningOnValuesAccess({'chain': 8, 'draw': 500, 'party_id_dim_0': 2, 'party_id:age_dim_0': 3}),
 FrozenMappingWarningOnValuesAccess({'chain': 4, 'draw': 1000, 'party_id_dim': 2, 'party_id:age_dim': 3}))
```

GStechschulte:
> In this example, the error is because bayeux is appending `_0` to `party_id_dim` of the posterior dims. This results in Bambi [discarding](https://github.com/GStechschulte/bambi/blob/9f1d9d179071abbb4cc6255242132829aae80faf/bambi/backend/pymc.py#L261C4-L261C86) all posterior dims because the dims in the PyMC model are inconsistent with the dims of the InferenceData returned by bayeux.
> 
> For example:
> 
> ```python
> print(bayeux_idata.posterior.dims, pymc_idata.posterior.dims)
> ```
> 
>  ```
> (FrozenMappingWarningOnValuesAccess({'chain': 8, 'draw': 500, 'party_id_dim_0': 2, 'party_id:age_dim_0': 3}),
>  FrozenMappingWarningOnValuesAccess({'chain': 4, 'draw': 1000, 'party_id_dim': 2, 'party_id:age_dim': 3}))
> ```

Update: I have added logic in the cleaning of idata to: (1) identify bayeux idata and to remove the trailing numeric suffix from the _dims, and (2) to rename the posterior dims to be consistent with the PyMC model coords.

zwelitunyiswa:
Would it be possible to allow access to the optimization methods from
Bayeux as well via Bambi?

ColCarroll:
I think you could -- @GStechschulte has a good outline here. If @tomicapretto thinks this is a reasonable idea in principle, I'd be happy to either collaborate on this to (programatically) get the `bayeux` inference methods in, or send a follow-up that generalizes it a bit, and allows optimization and VI.

tomicapretto:
I think this is really cool, thanks @GStechschulte and thanks @ColCarroll for `bayeux`. I'm not sure I am aware of all the details, but what is the reason why `bayeux` is appending `_dim_0` to dimension names? As far as I remember that was an `xarray` thing. Or is it that `bayeux` is not receiving dimension names from the PyMC model and thus it appends `_dim_0`?

Another thing, I see we're replacing `blackjax`, `jax`, `jaxlib`, and `numpyro` with `bayeux`. However, as far as I know `bayeux` does not install these dependencies, so doing `pip install bambi[jax]` won't give users access to JAX based samplers, right? (I'm not very familiar with `bayeux` so I may be wrong).

ColCarroll:
bayeux [will pull those in](https://github.com/jax-ml/bayeux/blob/main/pyproject.toml#L16), but i agree it is better to be explicit and require dependencies (in case bayeux makes weird decisions). 

I'll double check on the naming conventions!

ColCarroll:
Oh right, yes: `bayeux` has no concept of the dimensions from `pymc`. That would have to be implemented as a post-processing step to rename the `arviz` dimensions.

tomicapretto:
> Oh right, yes: `bayeux` has no concept of the dimensions from `pymc`. That would have to be implemented as a post-processing step to rename the `arviz` dimensions.

Thanks for the answer, it makes much more sense now!

tomicapretto:
> @ColCarroll thanks for the information.
> 
> @tomicapretto I can apply this post processing step on Bambi's side.

Sounds great, just let me know if you need help or a second opinion :)

GStechschulte:
Two updates:

1. I added a processing step for when Bambi cleans the idata, it renames the idata dims and coordinates to match those of the underlying PyMC model.
2. I explicitly added JAX based sampler dependencies.

Regarding

> I'd be happy to either collaborate on this to (programatically) get the bayeux inference methods in, or send a follow-up that generalizes it a bit, and allows optimization and VI.

@ColCarroll I'd be happy to collaborate and see how you would do this 👍🏼

bambi/backend/pymc.py
```diff
@@ -203,43 +221,27 @@ def _run_mcmc(
...
-                    f"'mcmc', 'nuts_numpyro' or 'nuts_blackjax'"
-                )
-        idata = self._clean_results(idata, omit_offsets, include_mean)
+        elif sampler_backend in BAYEUX_SAMPLERS:
+            import bayeux as bx
+            import jax
+
+            bx_model = bx.Model.from_pymc(self.model)
+            bx_sampler = getattr(bx_model.mcmc, sampler_backend)
+            idata = bx_sampler(seed=jax.random.key(0), **kwargs)
```
ColCarroll, line 233: there's a `random_seed` that gets passed in -- I presume from `numpy`? We could use `jax.random.key(np.random.randint(2**32 - 1, seed=random_key))`, which I think maintains the expected entropy?

bambi/backend/pymc.py
```diff
@@ -23,6 +24,23 @@
 __version__ = version("bambi")
 
 
+PYMC_SAMPLERS = ["mcmc"]
+BAYEUX_SAMPLERS = [
```
ColCarroll, line 28: We could do something like:

```
# TODO(colcarroll): Make this public
import importlib

def get_bayeux_samplers():
  bx_samplers = []
  if importlib.util.find_spec('bayeux') is None:
    return bx_samplers
  import bayeux as bx
  for module in bx._src.bayeux._MODULES:
    mname = module.__name__.split('.')[-1]
    for k in module.__all__:
      bx_samplers.append(mname + '.' + getattr(module, k).name)
  return bx_samplers
```

or, once a version of this is in `bayeux`, 

```
def get_bayeux_samplers():
  if importlib.util.find_spec('bayeux') is None:
    return bx_samplers
  import bayeux as bx
  return bx.algorithms 
```

this currently looks like 

```
['mcmc.tfp_hmc',
 'mcmc.tfp_nuts',
 'mcmc.tfp_snaper_hmc',
 'mcmc.blackjax_hmc',
 'mcmc.blackjax_chees_hmc',
 'mcmc.blackjax_meads_hmc',
 'mcmc.blackjax_nuts',
 'mcmc.blackjax_hmc_pathfinder',
 'mcmc.blackjax_nuts_pathfinder',
 'mcmc.flowmc_rqspline_hmc',
 'mcmc.flowmc_rqspline_mala',
 'mcmc.flowmc_realnvp_hmc',
 'mcmc.flowmc_realnvp_mala',
 'mcmc.numpyro_hmc',
 'mcmc.numpyro_nuts',
 'optimize.jaxopt_bfgs',
 'optimize.jaxopt_gradient_descent',
 'optimize.jaxopt_lbfgs',
 'optimize.jaxopt_nonlinear_cg',
 'optimize.optimistix_bfgs',
 'optimize.optimistix_chord',
 'optimize.optimistix_dogleg',
 'optimize.optimistix_gauss_newton',
 'optimize.optimistix_indirect_levenberg_marquardt',
 'optimize.optimistix_levenberg_marquardt',
 'optimize.optimistix_nelder_mead',
 'optimize.optimistix_newton',
 'optimize.optimistix_nonlinear_cg',
 'optimize.optax_adabelief',
 'optimize.optax_adafactor',
 'optimize.optax_adagrad',
 'optimize.optax_adam',
 'optimize.optax_adamw',
 'optimize.optax_adamax',
 'optimize.optax_amsgrad',
 'optimize.optax_fromage',
 'optimize.optax_lamb',
 'optimize.optax_lion',
 'optimize.optax_noisy_sgd',
 'optimize.optax_novograd',
 'optimize.optax_radam',
 'optimize.optax_rmsprop',
 'optimize.optax_sgd',
 'optimize.optax_sm3',
 'optimize.optax_yogi',
 'vi.tfp_factored_surrogate_posterior']
```

We can then use [`operator.attrgetter`](https://docs.python.org/3/library/operator.html#operator.attrgetter) where you use `getattr` on line 232. 

Note that the benefits of having this global constant here is just for debugging: a user can see valid inputs. You could also just use `pymc` if the input is "mcmc", or _try_ to use `bayeux` and fail in a reasonable way. 

Note also that `bayeux` includes optional dependencies using the same `importlib.util.find_spec` trick, so `flowMC` is not actually required, and the list of available algorithms depends on the user's environment.

pyproject.toml
```diff
@@ -36,10 +36,12 @@ dev = [
     "seaborn>=0.9.0",
 ]
 jax = [
+    "bayeux>=0.1.6",
     "blackjax>=1.0.0",
     "jax>=0.3.1",
     "jaxlib>=0.3.1",
     "numpyro>=0.9.0",
+    "flowMC>=0.2.4",
```
ColCarroll, line 44: can remove unless we think this is particularly handy for user problems (I don't!)

GStechschulte:
@ColCarroll thanks a lot for the review! I will incorporate these in the coming days.

bambi/backend/pymc.py
```diff
@@ -203,43 +221,27 @@ def _run_mcmc(
...
-                raise ValueError(
-                    f"sampler_backend value {sampler_backend} is not valid. Please choose one of"
-                    f"'mcmc', 'nuts_numpyro' or 'nuts_blackjax'"
-                )
-        idata = self._clean_results(idata, omit_offsets, include_mean)
+        elif sampler_backend in BAYEUX_SAMPLERS:
+            import bayeux as bx
+            import jax
+
+            bx_model = bx.Model.from_pymc(self.model)
```
ColCarroll, line 231: ok, as of bayeux 0.1.9, you could use `bx_model.methods`, which returns a dictionary. I would do something like

```
def maybe_get_bayeux_method(sampler_backend):
  if sampler_backend.count('.') == 1:
    return sampler_backend.split('.')
  return None

bayeux_method = maybe_get_bayeux_method(sampler_backend)

...later

if bayeux_method is not None:
  inference_kind, inference_type = bayeux_method
  bx_model = bx.Model.from_pymc(self.model)
  if inference_kind in bx_model.methods and inference_type in bx_model.methods[inference_kind]:
    sampler = getattr(getattr(bx_model, inference_kind), inference_type)
    res = sampler(seed=seed, **kwargs)
    if inference_kind == "mcmc":
      res = self._clean_results(res, ...)
    return res
formatted_methods = [f'{k}.{v} for k, vals in bx_model.methods.items() for v in vals]
raise ValueError(f"Use one of {formatted_methods}")
```

GStechschulte: Is this suggesting that users pass, for example, `mcmc.blackjax_nuts`, `mcmc.numpyro_nuts` instead of `blackjack_nuts`, `numpyro_nuts` to the parameter `inference_method`?

ColCarroll: yeah, exactly. and it allows optimization or vi to be run (but then the return type is kind of a wildcard)

GStechschulte:
Thanks for the reviews and suggestions @ColCarroll 😄 I have taken your suggestions and implemented them in slightly different ways. 

- Bambi has separate inference method calls for VI, MCMC, and Laplace approx. within the [PyMCModel](https://github.com/bambinos/bambi/blob/ff685b75a0ef17a11c8399de6de0a5f9d388d231/bambi/backend/pymc.py#L26) class. I think we should maintain this encapsulation. This does result in duplicate code though, e.g. 

```python3
import bayeux as bx
import jax

bx_model = bx.Model.from_pymc(self.model)
bx_method = operator.attrgetter(inference_method)(bx_model.<method>)
```

is written for MCMC, VI, Laplace approx., and optimize.

- When passing bayeux methods to `inference_method`, I have kept the naming convention to be `tfp_hmc`, `blackjax_nuts ` since this is also how the dict values are returned in `bx_model.methods`.

GStechschulte:
@tomicapretto should we allow optimization methods? I think it is really cool but `model.predict` will not work out of the box as optimization methods return `OptimizerResults` and not an idata object.

bambi/backend/pymc.py
GStechschulte: The `random_seed` parameter of `model.fit` is an int or list of ints (if cores > 1). `np.random.randint` doesn't have a `seed` param so it will generate a random in each time. Did you perhaps mean to use `jax.random.randint`?

In the case of JAX based methods, we could do something like

```python3
            import bayeux as bx
            import jax
            
            # Seed is required for bayeux
            if random_seed is None:
                random_seed = 0

            bx_model = bx.Model.from_pymc(self.model)
            bx_sampler = operator.attrgetter(sampler_backend)(bx_model.mcmc)
            idata = bx_sampler(seed=jax.random.key(random_seed), **kwargs)
```

although the name `random_seed` is ambiguous because it is not random. 

Since JAX random functions explicitly consume the state (the key), feeding the same key to a random function will always result in the same sample being generated. If the user does not pass a key, we explicitly set it to `0`, else the user passed key is passed to `jax.random_key`. This way, we get reproducible results. 

Apologies for the newb question, I am relatively new to JAX PRNG.

zwelitunyiswa:
It would be cool if it could. It would allow non-Bayesians an entry point into Bayesian software/methods, with a simple change of sampler.

GStechschulte:
> It would be cool if it could. It would allow non-Bayesians an entry point into Bayesian software/methods, with a simple change of sampler.
> […](#)
> On Mar 1, 2024 at 02:24 -0500, Gabriel Stechschulte ***@***.***>, wrote: @tomicapretto should we allow optimization methods? I think it is really cool but model.predict will not work out of the box as optimization methods return OptimizerResults and not an idata object. — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you commented.Message ID: ***@***.***>

That's true. Maybe in this PR I will leave optimization methods out. Then, in a separate PR, add optimization algorithms and get the `OptimizerResults` object to work with `model.predict`. What do you think?

bambi/backend/pymc.py
```diff
@@ -406,3 +406,39 @@ def _posterior_samples_to_idata(samples, model):
...
+        A dict where the keys are the module names and the values are the methods
+        available in that module.
+    """
+    bx_methods = {}
+    if importlib.util.find_spec("bayeux") is None:
+        return bx_methods
+
+    import bayeux as bx  # pylint: disable=import-outside-toplevel
+
+    bx_methods = {}
```
ColCarroll, line 427: maybe replace the rest of the function with 
```
# use dummy log density to get access to all methods.
return bx.Model(lambda x: -x**2, 0.).methods
```

bambi/model_components.py
```diff
@@ -239,11 +239,12 @@ def predict_common(
             X = np.delete(X, term_slice, axis=1)
 
         # Add HSGP components contribution to the linear predictor
+        hsgp_slices = []
```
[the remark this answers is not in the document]
GStechschulte: They are unrelated. I rebased with main.

bambi/backend/pymc.py
GStechschulte: Hah. Thanks! I got it mixed up.

bambi/models.py
```diff
@@ -306,7 +306,7 @@ def fit(
         Returns
         -------
         An ArviZ ``InferenceData`` instance if inference_method is  ``"mcmc"`` (default),
-        "nuts_numpyro", "nuts_blackjax" or "laplace".
+        "numpyro_nuts", "blackjax_nuts" or "laplace".
```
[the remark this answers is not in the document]
GStechschulte: I am going to implement a property attribute so the user can do `model.inference_methods` and get a dictionary similar to `bx_model.methods`. Then in docstrings, we can reference this.

tests/test_alternative_samplers.py
```diff
@@ -52,27 +56,13 @@ def test_vi():
...
-@pytest.mark.parametrize(
-    "args",
-    [
-        ("mcmc", {}),
-        ("nuts_numpyro", {"chain_method": "vectorized"}),
-        ("nuts_blackjax", {"chain_method": "vectorized"}),
-    ],
-)
-def test_logistic_regression_categoric_alternative_samplers(data_n100, args):
+@pytest.mark.parametrize("sampler", MCMC_METHODS)
```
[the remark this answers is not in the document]
GStechschulte: Indeed, they do. My original intention for these tests was to ensure that all of the samplers can be called and "do something". Not necessarily to test how well they approximate the simulated data. 

For this, I review the corresponding libraries to ensure they have tests in place for their algorithms.

tomicapretto:
> @tomicapretto should we allow optimization methods? I think it is really cool but `model.predict` will not work out of the box as optimization methods return `OptimizerResults` and not an idata object. I added it, but we can also delete it.
> 
> _Update_: In this PR I will leave optimization methods out. Then, in a separate PR, add optimization algorithms and get the `OptimizerResults` object to work with `model.predict`.

I agree it's good to have them but also good to leave them out for now. 

I don't know all the details of the optimization methods. I guess we would get a single number for every parameter in the posterior, right? If that is the case, how do they differ from MAP? Is it possible to use those optimization results to get distributions for the parameters? I mention this because all the functions in the `interpret` submodule show both point estimates (i.e. a point or a line) plus uncertainty (i.e. a band or a range). If we can't get any measure of uncertainty, then the plots will look different and we'll need to account for that in the implementation

bambi/models.py
```diff
@@ -266,7 +266,7 @@ def fit(
             using the ``fit`` function.
             Finally, ``"laplace"``, in which case a Laplace approximation is used and is not
             recommended other than for pedagogical use.
-            To use the PyMC numpyro and blackjax samplers, use ``nuts_numpyro`` or ``nuts_blackjax``
+            To use the PyMC numpyro and blackjax samplers, use ``numpyro_nuts`` or ``blackjax_nuts``
```
tomicapretto, line 269: Quick question. What is the reason to change these? Is it because the documentation was wrong?

If that is not the case, and we're changing the implementation, we should add some message as it may break users' existing code.

GStechschulte: The reason for the change is to match the naming convention in bayeux.

GStechschulte:
> I don't know all the details of the optimization methods. I guess we would get a single number for every parameter in the posterior, right? If that is the case, how do they differ from MAP? Is it possible to use those optimization results to get distributions for the parameters? I mention this because all the functions in the `interpret` submodule show both point estimates (i.e. a point or a line) plus uncertainty (i.e. a band or a range). If we can't get any measure of uncertainty, then the plots will look different and we'll need to account for that in the implementation.

Yeah, it really depends on the parameters of the optimization method called. If you use a lot of particles, then you would get more than one parameter estimate. So it "looks" like a posterior. Nonetheless, this is definitely one aspect we need to consider.

bambi/backend/pymc.py
```diff
@@ -424,21 +432,5 @@ def _get_bayeux_methods():
```
ColCarroll, line 432: Probably want to return `{"mcmc": []}` on line 431 to avoid `KeyError` elsewhere.

ColCarroll (approved):
LGTM in terms of working with `bayeux`. It looks like CI needs to be fixed and @tomicapretto give final approval for it being in the `bambi` style. 

Thanks for taking this on!

bambi/models.py
tomicapretto: I think it's the right thing to do, but right now I would like to have something that doesn't break user existing code. Something along the following lines

```python
if inference_method == "nuts_numpyro":
    inference_method = "numpyro_nuts"
    warnings.FutureWarning("'numpyro_nuts has been replaced by 'nuts_numpyro' and will be removed in the future")
elif inference_method == "nuts_blackjax":
    inference_method = "blackjax_nuts"
    warnings.FutureWarning("'blackjax_nuts' has been replaced by 'nuts_blackjax' and will be removed in the future")
```

what do you think about it?

tomicapretto:
@GStechschulte I think this is a great addition and the PR is in great shape. Just want to know your opinion on my suggestion about handing, for now, the old argument values as well

zwelitunyiswa:
@tomicapretto @GStechschulte will the samplers work with PYMC's mixture modes like the zero-inflated and hurdle models?

Z.

GStechschulte:
> LGTM in terms of working with `bayeux`. It looks like CI needs to be fixed and @tomicapretto give final approval for it being in the `bambi` style.
> 
> Thanks for taking this on!

Thanks for the reviews, patience, and bayeux 😄

bambi/models.py
GStechschulte: I think this is a good idea. It also gives me time to go through the Bambi docs and any `inference_method` passing `numpyro_nuts`.

GStechschulte:
> @GStechschulte I think this is a great addition and the PR is in great shape. Just want to know your opinion on my suggestion about handing, for now, the old argument values as well

Thanks a lot! I replied above and made the change 👍🏼

GStechschulte:
> @tomicapretto @GStechschulte will the samplers work with Bambi's mixture modes like the zero-inflated and hurdle models?
> 
> Z.
 
Yup! Though, some samplers are better suited for different problems, etc. 😄

GStechschulte:
Ugh. pylint is making the CI fail. It says it cannot import bayeux. However, when I check the logs of the step "Install Bambi and all its dependencies", I can see that bayeux was installed.

ColCarroll:
the package is named `bayeux-ml` -- i'm checking now that you got that one.

pyproject.toml
```diff
@@ -36,6 +36,7 @@ dev = [
     "seaborn>=0.9.0",
 ]
 jax = [
+    "bayeux>=0.1.9",
```
ColCarroll, line 39: yeah, `bayeux-ml` here.

GStechschulte: 🤦🏼

ColCarroll:
Oh. Last failures are because `bayeux` uses Python 3.10-only features (union types with a pipe, like `str | None`). I can get rid of those, but also [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table) suggests that Python 3.9 get dropped in 30 days. I see at least three choices here:

1. I release a Python 3.9 compatible bayeux version, then we can merge this.
2. We disable testing on Python 3.9 in this PR and merge. Things are confusing for users on Python 3.9 who `pip install bambi[jax]`, but I can't find a way to restrict python versions in optional dependencies.
3. We just wait 30 days, do a PR removing Python 3.9 support, then merge this.

Any preferences?

tomicapretto:
I'm in favor of following NumPy's deprecation cycle but at the same time I feel it moves faster than many users. For that reason, even when the pipe operator is a feature I really like, I have not been using it because of this (i.e.  not working for Python <= 3.9).

According to PyPI stats, users are downloading Bambi using Python 3.7, 3.8, and 3.9. I know PyPI stats may not be a crystalline reflection of reality, but it's still useful. I don't want to break their installations.

So, we could drop official support for 3.9 as soon as NumPy does (this includes testing with Python >= 3.10 only), but I would not start using Python >= 3.10 specific features in the default installation of Bambi right now. I guess there will be a point where it'll become unavoidable (for example, when PyMC or PyTensor start to do so) and that's fine.

What do others think?

GStechschulte:
Thanks @ColCarroll for the information, and I agree with @tomicapretto thoughts. Looking forward to eventually merge this!

tomicapretto:
> According to PyPI stats, users are downloading Bambi using Python 3.7, 3.8, and 3.9. I know PyPI stats may not be a crystalline reflection of reality, but it's still useful. I don't want to break their installations.

Just want to add that after some comms in our Slack it looks like we _won't_ be breaking users' installations. They'll just get an older version of Bambi if they use Python < 3.10

GStechschulte:
@tomicapretto so do we want to wait the $x$ days to drop support for Python 3.9, or to drop it now?

tomicapretto:
> @tomicapretto so do we want to wait the x days to drop support for Python 3.9, or to drop it now?

sorry for the slow response, we can do it now if you agree on that

GStechschulte:
CI failing on 3.12 seems related to this PyTensor [issue](https://github.com/pymc-devs/pytensor/pull/488).

tomicapretto:
> CI failing on 3.12 seems related to this PyTensor [issue](https://github.com/pymc-devs/pytensor/issues/589) and corresponding [PR](https://github.com/pymc-devs/pytensor/pull/488).

Ha! Should we avoid supporting 3.12 in the immediate future?

tomicapretto:
I was giving this a shot and just found that the development version of PyMC now supports Python 3.12

https://github.com/pymc-devs/pymc/blob/61ce412aa599939eaf299a6328059c184e6c25db/setup.py#L33-L36

This is not the case in the latest release 5.11. I think we can just wait a bit until PyMC releases a new version to upgrade the PyMC version requirement in our `pyproject.toml` and then we're done.

Another option would be to pin PyMC to the current development version but that will have to be updated as soon as PyMC releases the version suporting Python 5.12 so I think it's just better to wait a bit.

https://github.com/pymc-devs/pymc/blob/a06081e1e9649bd56e3528cb96380efdf6bb2dc0/setup.py#L33-L35

tomicapretto:
There's a new version of PyMC supporting Python 3.12... :crossed_fingers:

tomicapretto:
@GStechschulte I let you have the honor to click the green button :D

GStechschulte:
> @GStechschulte I let you have the honor to click the green button :D

Woooooo! Gracias! 🎉

ColCarroll:
Thanks for all the persistence @GStechschulte!

PR #782, opened 2024-02-21
Add polynomial regression example

tjburch:
This PR creates a new example notebook of polynomial regression using Bambi that I've been working on for the last few weeks. There's still some tweaks on the last example that need to be done, I'm hoping to get some feedback here.

Immediate questions:

1. Is there any appetite for including this example in the first place? It's a bit atypical compared to the others in that it's not featuring a unique functionality of Bambi itself, more in the formula component. However, I think there's some nuances that are nice to explain. This example also displays the impact of setting priors and using it with an experiment in mind, so I think those are interesting too. However, if it's better suited for a personal blog, that's fine, happy to do that.
2. If yes, there is a _long_ aside in the middle, where I derive how orthogonal polynomials are created. Is that of interest, or beyond the scope? We could definitely find ways to trim it down, or cut it.
3. I am pretty certain the last model suffers from some type of mis-specification, however I've struggled pinning it down. I was hoping to be able to include a hierarchical effect on the linear term, but it seem to be fitting quite right. If anyone has a second to take a look at the DGP and the associated model and suggest other approaches, that'd be super helpful.

Thanks in advance for feedback, hopefully it's an interesting notebook regardless if it's suited for the example gallery or not.


ColCarroll:
A bit of a side note that I thought this was really nice. 

I found the bit on orthogonal polynomials really interesting, perhaps as a separate notebook? In particular, I was disappointed when you tee'd up "these help with inference", then followed with "...but we won't use them". Show us the `poly`!

tjburch:
I wanted to add another example with poly but I was worried it was getting long as-is!

A separate notebook seems like a good solution, one more focused on the orthogonal polynomials with more of a statistical model lean and then the current notebook with the I(x**2) approach.

ColCarroll:
Also, my only productive feedback is that

```
df_falling.assign(tsquared=t**2)
```
is a neat trick to create a copy of `df_falling` with an extra column. I had trouble spotting it, but will try to remember to use that pattern more! I might be alone in failing to see that, but if there was a way to point out that you're doing that, it may be helpful.

tomicapretto:
@tjburch thanks a lot! I quickly skimmed through it and I like it! I will have a more in-depth read and come back with feedback :)

zwelitunyiswa:
@tjburch This is great! I love how you move between the science, the data, and the model, and a reader can see how the science is captured in the model, and how the model's posterior artifacts link back to the data and science. We also see the choices one has to make. I find the most valuable insights I have gleaned in Bayesianism is seeing people like Solomon Kurz and Andrew Heiss work through real examples and the internal dialogue they have.

tomicapretto:
@tjburch regarding what you shared as "immediate questions"

1. Yes, this is an excellent example.
2. I really like having it, but it's also your content, so feel free to use it as your own blog post too.
3. I think I already helped in the previous comment, but happy to answer if you have questions

tjburch:
Thanks @tomicapretto. I'll work on updating with the correct poly reference and give this a test! Much appreciated you taking a look at it.

Would a separate example with `poly` in a model be desired, and if so, in this notebook or do you think splitting it out into 2 (e.g. one "polynomial regression" and one "polynomial regression using orthogonal polynomials" or something along those lines?)

tomicapretto:
> Thanks @tomicapretto. I'll work on updating with the correct poly reference and give this a test! Much appreciated you taking a look at it.
> 
> Would a separate example with `poly` in a model be desired, and if so, in this notebook or do you think splitting it out into 2 (e.g. one "polynomial regression" and one "polynomial regression using orthogonal polynomials" or something along those lines?)

If you feel like you want to do it, yeah, I think two separate notebooks would be better. But it's not a problem if you want to leave it as it is (we can split things it in the future)

tjburch:
I'm still working to clean up the original notebook, but I've just added a second notebook that's purely dedicated to just the orthogonal polynomials and using `poly`. It's ready for feedback if/when anyone gets time, and in the meantime, I'll be working to update the original notebook too.

tjburch:
@tomicapretto Sorry to bother for more feedback here but I'm realizing that some of these fits aren't optimal because the data is truncated, but I didn't explicitly include that into the model.

image

However, when I do try to add the truncation like what was added in #752 I run into issues. For example:

```
planet_model = bmb.Model('truncated(Height, lb=-0.5)  ~ I(Time**2):Planet + Time + 0', df)
planet_fit = planet_model.fit(idata_kwargs={"log_likelihood": True}, random_seed=SEED)
```

gives a sampling error:

```
SamplingError: Initial evaluation of model at starting point failed!
Starting values:
{'truncated(Height, lb=-0.5)_sigma_log__': array(nan), 'I(Time ** 2):Planet': array([0.07158144, 0.44894538, 0.93178634]), 'Time': array(0.99263572)}

Logp initial evaluation results:
{'truncated(Height, lb=-0.5)_sigma': nan, 'I(Time ** 2):Planet': -inf, 'Time': -inf, 'truncated(Height, lb=-0.5)': nan}
You can call `model.debug()` for more details.
```

Any suggestions on what I'm doing wrong here?

tomicapretto:
@tjburch the orthogonal polynomial notebook is just AMAZING! Thanks!!

tjburch:
Thanks! Hoping to have the original cleaned up shortly. You may have seen that I deleted a previous comment, I'm running into errors when truncating the response variable, which I think is contributing to the underestimates. I was able to reproduce the toy example in #752 this morning though, so I just need to side-by-side to try to understand what's different.

tomicapretto:
@tjburch I could swear I saw a comment and when I came back today, I couldn't find it, haha! If a truncated normal does not work, you could try another distribution with a positive domain (the problem is that you'll likely need a non-identity link function and then the meaning of parameter estimates change and that's another problem)

tjburch:
<p>This is actually something I learned while putting this together. The first call of <code>np.diag</code> extracts the diagonal elements. The second call constructs a diagonalized matrix:</p><p> </p><pre class="ql-syntax" spellcheck="false">In [1]: import numpy as np

In [2]: mat = np.array([[1,2,3],[4,5,6],[7,8,9]])

In [3]: mat
Out[3]:
array([[1, 2, 3],
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; [4, 5, 6],
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; [7, 8, 9]])

In [4]: np.diag(mat)
Out[4]: array([1, 5, 9])

In [5]: np.diag(np.diag(mat))
Out[5]:
array([[1, 0, 0],
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; [0, 5, 0],
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; [0, 0, 9]]) 
</pre><p> </p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/782/discussion/#comment-1976797442'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

tomicapretto:
> This is actually something I learned while putting this together. The first call of `np.diag` extracts the diagonal elements. The second call constructs a diagonalized matrix:
> 
> In [1]: import numpy as np
> In [2]: mat = np.array([[1,2,3],[4,5,6],[7,8,9]])
> 
> 
> In [3]: mat
> Out[3]:
> array([[1, 2, 3],
>        [4, 5, 6],
>        [7, 8, 9]])
> 
> 
> In [4]: np.diag(mat)
> Out[4]: array([1, 5, 9])
> 
> 
> In [5]: np.diag(np.diag(mat))
> Out[5]:
> array([[1, 0, 0],
>        [0, 5, 0],
>        [0, 0, 9]])
> 
> [View entire conversation](https://app.reviewnb.com/bambinos/bambi/pull/782/discussion/#comment-1976797442) on ReviewNB

26ufdipQqU2lhNA4g

tjburch:
Alright, I think this is probably ready to no longer be a WIP and good for final review. 

I took out the last example with multiple astronauts in the original notebook, to me it didn't seem to help users understand more information and injecting priors didn't _really_ help the estimates I hoped it would, so it seemed like not a huge value add. If anyone feels strongly otherwise, I can go put it back in, but I think that the notebooks as-is do a pretty good job conveying the ins-and-outs of polynomial regression.

tomicapretto:
Amazing @tjburch! I'll review it soon

tomicapretto:
@tjburch I didn't mean to close this. I tried to add some changes to this branch but I broke something :S. I'm trying to fix it.

tjburch:
Whoops! If you're unable to recover, let me know and I can open a new PR or something, but all things equal it'd be nice to keep some of this prior discussion.

PR #790, opened 2024-03-29
Type-hint and type-overload a few common Model methods

thomasaarholt:
Hello!

I have used pymc extensively over the last two years, and have finally had a play with bambi. I _really_ like it - it has some really nice defaults, and the formula syntax is great!

I noticed that some type hinting was missing on the most commonly used Model methods, so I decided to add that. I initially just wanted to add some overload methods to `Model.fit`, so that it would clearly be returning an `InferenceData` object when called with the appropriate `inference_method`s. Then I got a bit carried away, since it wasn't very difficult to add the remaining commonly used methods. I've [previously added type hinting](https://github.com/pymc-devs/pymc/pulls?q=is%3Apr+sort%3Aupdated-desc+author%3Athomasaarholt+is%3Aclosed) to some of the functions in pymc.

I am using pyright to check the types. This is what is used in VSCode for type checking, docstrings and autocompletion.
This PR does not change any behaviour or logic, only adds type hinting that make IDEs better at understanding the code.

Here is a screenshot of my editor from before the changes are made:
image

And here they are after:
image

[Here is the run.py file in a gist](https://gist.github.com/thomasaarholt/68e30bfc11e5e9f0cd7534f3b3cb20fb).

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -1,6 +1,7 @@
 # pylint: disable=no-name-in-module
 # pylint: disable=too-many-lines
 import logging
+from typing import Literal, Optional, overload
 import warnings
 
 from copy import deepcopy
@@ -11,6 +12,7 @@ import pymc as pm
 import pandas as pd
 
 from arviz.plots import plot_posterior
+from arviz import InferenceData
 
 from bambi.backend import PyMCModel
 from bambi.defaults import get_builtin_family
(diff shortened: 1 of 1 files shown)


tomicapretto:
Thanks for the contribution @thomasaarholt! I'll let you know what I think once I can review. For now, let's run the CI :)

thomasaarholt:
Great! No rush at all, just had some downtime in my Easter break and thought it would be a nice little contribution.

bambi/models.py
```diff
@@ -225,21 +227,57 @@ def __init__(
...
-        n_init=50000,
-        chains=None,
-        cores=None,
-        random_seed=None,
+        draws: int = 1000,
+        tune: int = 1000,
+        discard_tuned_samples: bool = True,
+        omit_offsets: bool = True,
+        include_mean: bool = False,
+        inference_method: Literal["mcmc", "blackjax_nuts", "numpyro_nuts", "laplace"] = "mcmc",
```
tomicapretto, line 238: One problem I see here is that after #775, users have access to a huge variety of samplers through bayeux. I'm not sure how this is handled with literals. But, is it possible that what we need is to create a Literal on the fly (at import time) where we check whether users have access to bayeux, and depending on that, the Literal contains different types?

Also, I don't think there's a convenient way to access the list of samplers without first creating a model. I'm going to create an issue now.

bambi/models.py
```diff
@@ -763,15 +807,37 @@ def prior_predictive(self, draws=500, var_names=None, omit_offsets=True, random_
...
-        kind="mean",
-        data=None,
-        inplace=True,
-        include_group_specific=True,
-        sample_new_groups=False,
-    ):
+        idata: InferenceData,
+        kind: Literal["mean", "pps"] = "mean",
+        data: Optional[pd.DataFrame] = None,
+        inplace: Literal[True] = True,
```
tomicapretto, line 816: Shouldn't this be `bool` instead of `Literal[True]`? I'm not a typing expert so I may be missing some details, but since this can be either `True` or `False`, I think it should be `bool`

bambi/models.py
```diff
@@ -763,15 +807,37 @@ def prior_predictive(self, draws=500, var_names=None, omit_offsets=True, random_
...
+        sample_new_groups: bool = False,
+    ) -> None: ...
+
+    @overload
+    def predict(
+        self,
+        idata: InferenceData,
+        kind: Literal["mean", "pps"] = "mean",
+        data: Optional[pd.DataFrame] = None,
+        inplace: Literal[False] = False,
```
tomicapretto, line 827: Same as above

thomasaarholt:
Ouch, sorry that I never got around to this!

tomicapretto:
> Ouch, sorry that I never got around to this!

No problem! But since it's been so long ago and the codebase quite a bit, I thought it would be good to just close it.

thomasaarholt:
Oh yes, don't worry about it! I haven't been able to work on pymc/bambi for a while, so I wouldn't be able to justify time on it now anyway!

+1 on closing old PRs!

PR #801, opened 2024-04-09
Fix `get_model_covariates()` utility function

### LINKED ISSUE #797: Non covariate variables being detected as covariates
Thanks to #796 I found that the `get_model_covariates()` function is returning non-covariate variables as covariates.

Example:

```python
import bambi as bmb
import numpy as np
import pandas as pd

from bambi.interpret.utils import get_model_covariates

df = pd.read_csv("https://raw.githubusercontent.com/m-clark/generalized-additive-models/master/data/pisasci2006.csv")

knots_income = np.linspace(np.min(df['Income']), np.max(df['Income']), 5+2)[1:-1]
formula = "Overall ~ 1 + bs(Income, degree=3, knots=knots_income)"
model = bmb.Model(formula, df, dropna=True)
get_model_covariates(model)
```

```
array(['Income', 'knots_income'], dtype='<U12')
```

The non covariate variable is `knots_income`. The variable is not looked up into the data frame. It comes from the outer environment.

tomicapretto:
This PR closes #797. 

It returns those names available in `Model.data`, which is the data frame where covariates are taken from.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 ### Maintenance and fixes
 
 * Fix bug in predictions with models using HSGP (#780)
+* Fix `get_model_covariates()` utility function (#801)
 
 ### Documentation
 
--- a/bambi/interpret/utils.py
+++ b/bambi/interpret/utils.py
@@ -258,6 +258,9 @@ def get_model_covariates(model: Model) -> np.ndarray:
 
     flatten_covariates = [item for sublist in covariates for item in sublist]
 
+    # Don't include non-covariate names (#797)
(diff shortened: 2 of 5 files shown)


tomicapretto:
I think the reason why CI failed is https://github.com/pymc-devs/pymc/releases/tag/v5.13.0 and https://github.com/pymc-devs/pymc/pull/7047. I'm setting `pymc<5.13.0` in our `pyproject.toml`. If CI passes I will merge. I think we need to resolve the issues created by the changes in PyMC in a different PR.

GStechschulte:
Thanks a lot @tomicapretto 👍🏼

PR #809, opened 2024-04-27
Add polynomial regression example (reopen)

tjburch:
Reopening #782 after git messiness. Sorry for not using a feature branch! 😬

I think this should all be right, I just merged changes in my fork and immediately opened. There's a couple of comments in the changelog that I'm a little concerned might be a mistake, but I'll let you took a look.


tomicapretto:
Thanks so much @tjburch

PR #823, opened 2024-07-05
Update docstring and args

GStechschulte:
This PR changes the docstring of `kind` in `model.predict` to reflect the changes in #804. 

Additionally, in the `interpret` function `predictions`, we allow the user to compute posterior predictive samples. I updated the argument passed to `model.predict` from `kind="pps"` to `kind="response"` to reflect the changes in the merged PR above.

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -826,7 +826,7 @@ class Model:
             ``idata`` with the predictions added. If ``kind="response_params"``, a new variable
             with the name of the parent parameter, e.g. ``"mu"`` and ``"sigma" for a Gaussian
             likelihood, or ``"p"`` for a Bernoulli likelihood, is added to the ``posterior`` group.
-            If ``kind="response"``, it appends a ``posterior_predictive`` group to ``idata``. If 
+            If ``kind="response"``, it appends a ``posterior_predictive`` group to ``idata``. If
             any of these already exist, it will be overwritten.
         include_group_specific : bool
             Determines if predictions incorporate group-specific effects. If ``False``, predictions
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -823,8 +823,8 @@ class Model:
         inplace : bool
             If ``True`` it will modify ``idata`` in-place. Otherwise, it will return a copy of
             ``idata`` with the predictions added. If ``kind="response_params"``, a new variable
-            with the name of the parent parameter, e.g. ``"mu"`` or ``"p"``is added to the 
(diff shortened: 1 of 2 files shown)


bambi/models.py
```diff
@@ -812,20 +812,20 @@ def predict(
...
             The ``InferenceData`` instance returned by ``.fit()``.
         kind : str
-            Indicates the type of prediction required. Can be ``"mean"`` or ``"pps"``. The
-            first returns draws from the posterior distribution of the mean, while the latter
-            returns the draws from the posterior predictive distribution (i.e. the posterior
-            probability distribution for a new observation) in addition to the mean posterior
-            distribution. Defaults to ``"mean"``.
+            Indicates the type of prediction required. Can be ``"response_params"`` or
+            ``"response"``. The first returns draws from the posterior distribution of
+            the mean, while the latter returns the draws from the posterior predictive
```
tomicapretto, line 817: Can it be "from the posterior distribution of the parameters of the likelihood"? As those parameters could be something different from the mean

bambi/models.py
```diff
@@ -812,20 +812,20 @@ def predict(
...
+            distribution (i.e. the posterior probability distribution for a new observation)
+            in addition to the mean posterior distribution. Defaults to ``"response_params"``.
```
tomicapretto, line 819: Similar here, it says "mean posterior distribution"

bambi/models.py
```diff
@@ -812,20 +812,20 @@ def predict(
...
             An optional data frame with values for the predictors that are used to obtain
             out-of-sample predictions. If omitted, the original dataset is used.
         inplace : bool
             If ``True`` it will modify ``idata`` in-place. Otherwise, it will return a copy of
-            ``idata`` with the predictions added. If ``kind="mean"``, a new variable ending in
-            ``"_mean"`` is added to the ``posterior`` group. If ``kind="pps"``, it appends a
-            ``posterior_predictive`` group to ``idata``. If any of these already exist, it will be
-            overwritten.
+            ``idata`` with the predictions added. If ``kind="response_params"``, a new variable
+            with the name of the parent parameter, e.g. ``"mu"`` or ``"p"``is added to the
```
tomicapretto, line 826: Here it can be multiple parameters (e.g. both mu and sigma)

GStechschulte: Good point. I updated it to include 2 minimal examples (Gaussian and Bernoulli) so the user can reason what parameters to expect in the posterior group.

tomicapretto (approved):

PR #836, opened 2024-08-23
Adding Horseshoe Prior

julianlheureux:
Hello! In this pull request, I added the horseshoe prior to the dictionary where Bambi stores specific prior distributions.

This is Carvalho's horseshoe prior, designed for the coefficients of a multiple regression model with a large number of explanatory variables, where only a few are statistically significant.


bambi/backend/utils.py
```diff
@@ -5,7 +5,16 @@
...
 import pymc as pm
 
-MAPPING = {"Cumulative": pm.Categorical, "StoppingRatio": pm.Categorical}
+
+def horseshoe(name, tau_nu=3, lam_nu=1, dims=None):
+    tau = pm.HalfStudentT(f"{name}_tau", nu=tau_nu)
+    lam = pm.HalfStudentT(f"{name}_lam", nu=lam_nu, dims=dims)
+    beta_raw = pm.Normal(f"{name}_raw", 0, 1, dims=dims)
+    beta = pm.Deterministic(name, beta_raw * tau**2 * lam**2, dims=dims)
+    return beta
```
tomicapretto, line 14: Could you add a docstring here explaining what the arguments are? 

* `name` is the name of the parameters as registered in the PyMC model
* `tau_nu` the degrees of freedom of `tau`
* `lam_nu` the degrees of freedom of `lam`
* `dims` dimensions passed to PyMC. 

Also, 

* Mention in the docstring this is an internal function which is not supposed to be used by users. 
* Include that by default the prior on `lam` is equivalent to a `HalfCauchy` (because of `lam_nu=1`)

tomicapretto: @julianlheureux you can have a look at some examples of docstrings here 

https://github.com/bambinos/bambi/blob/038610c306007484c259e24ddbd6461669868066/bambi/transformations.py

and this is just a particular example

https://github.com/bambinos/bambi/blob/038610c306007484c259e24ddbd6461669868066/bambi/transformations.py#L152-L166

tomicapretto:
@julianlheureux thanks for the contribution! This is something we wanted for a long time.

Before merging, it would be nice to add a test. I imagine the following would be OK as it covers the basic use cases:

* A test where we use the prior with the default arguments.
* A test where we use the prior with non-default arguments.

In both tests I would check that

* The underlying PyMC distributions are of the expected families with the expected arguments
* The sampler runs

I would not check for the absence of divergences, etc. because we can't guarantee they won't happen.

I know the PyMC part of the tests is not trivial as one has to interact with PyMC internals. So, if you want you can get started with the test and I can share a snippet showing how to test that.

tomicapretto:
Thanks @julianlheureux!

There's something happening with the tests, but that's unrelated to the code contribution. It has to be with data loading.

Issue #418, opened 2021-10-14
Add example using `offset()`


tomicapretto:

hadjipantelis:
If someone needs an `offset`-like solution "here and now" my work-around would be something like: `priors={'log_exposure': Prior("Normal", auto_scale=False, mu=1.0, sigma= 0.01)}`, i.e. you define a really tight prior around 1. It gave me the exact same results as using offset functionality.

DrEntropy:
Shouldn't this issue be closed due to #842  ?

tomicapretto:
> Shouldn't this issue be closed due to [#842](https://github.com/bambinos/bambi/pull/842) ?

Yes!

PR #842, opened 2024-09-06
Add an example for using `offset`, addressing issue #418

DrEntropy:
* Add a simple example of using offset in count data using 'Roaches' example from Regression and Other Stories. 
* Address issue #418


DrEntropy:
I added at the end the equivalent pymc model just to illustrate what 'offset' is doing.   BTW as far as I can determine this `offset` is not documented.  If so where should it be documented?

GStechschulte:
Thanks a lot for the PR! Adding the equivalent PyMC model is a good idea. 

> BTW as far as I can determine this offset is not documented.

You are right, it is not documented anywhere. It should probably be documented in the [Getting started](https://bambinos.github.io/formulae/notebooks/getting_started.html#Built-in-transformations) section of `formulae` (our library for formulas for mixed effects models. But as far as Bambi goes, I think having the notebook is great for the moment.

tomicapretto:
Hi @DrEntropy! Thanks a lot for going ahead and writing this example. It's great to have it in our documentation.

I have very minor stylistic comments (code formatting) but it'll be easier for me to change them in a separate PR, so I'm merging this now.

PR #843, opened 2024-09-07
change os.rename to os.replace in pre-render.py

DrEntropy:
Building the `Bambi` docs with quarto was giving a "file exists" error during pre-render while trying to rename "CHANGELOG.md" to "changelog.qmd".    Simply changing `os.rename` to `os.replace` fixes this. 

An alternative would be to add the CHANGELOG.md to the `_quarto.yml` explicitly (quarto can render md too)

### DIFF
--- a/docs/pre_render.py
+++ b/docs/pre_render.py
@@ -3,7 +3,7 @@ import shutil
 
 # Copy CHANGELOG file to 'docs' and change its extension to '.qmd'
 shutil.copyfile("../CHANGELOG.md", "CHANGELOG.md")
-os.rename("CHANGELOG.md", "changelog.qmd")
+os.replace("CHANGELOG.md", "changelog.qmd")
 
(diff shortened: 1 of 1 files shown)


tomicapretto:
> Building the `Bambi` docs with quarto was giving a "file exists" error during pre-render while trying to rename "CHANGELOG.md" to "changelog.qmd". Simply changing `os.rename` to `os.replace` fixes this.
> 
> An alternative would be to add the CHANGELOG.md to the `_quarto.yml` explicitly (quarto can render md too)

Thanks @DrEntropy for the contribution! I'll merge this once CI is green. Regarding Quarto, I'm aware we can add CHANGELOG.md to `_quarto.yml`. However, the last time I tried it, it was not working as expected. If I recall correctly, a non-rendered version of the file was being displayed in our webpage.

Issue #850, opened 2024-10-18
Non-parent parameter not found in posterior when using fixed Data prior


ivanistheone:
Hi all. I ran into an issue similar to https://github.com/bambinos/bambi/issues/750 where a variable required for posterior predictive of the response variable is not included in the inference data object.

I'm trying to fit a Gaussian model with known, fixed variance sigma=15, and custom prior norm(100,40) on the mean. This is for educational purposes, to show the simplest possible model. I found a way to add `sigma` as constant, by setting a `bmb.Prior("Data", value=15)`, and the complete code example is like this:

```Python
# toy dataset
import pandas as pd
iqs = [ 82.6, 105.5,  96.7,  84.0, 127.2,  98.8,  94.3]
df = pd.DataFrame({"iq":iqs})


# Gaussian model with known variance sigma=15 and norm(100,40) prior on mean
import bambi as bmb
priors = {
    "Intercept": bmb.Prior("Normal", mu=100, sigma=40),
    "sigma": bmb.Prior("Data", value=15),
}
mod = bmb.Model("iq ~ 1",
                priors=priors,
                family="gaussian",
                link="identity",
                data=df)
mod
#      Formula: iq ~ 1
#       Family: gaussian
#         Link: mu = identity
# Observations: 7
#       Priors: 
#   target = mu
#       Common-level effects
#           Intercept ~ Normal(mu: 100.0, sigma: 40.0)
#       Auxiliary parameters
#           sigma ~ Data(value: 15.0)

idata = mod.fit()
# WORKS OK
```

Here sigma is [not included](https://github.com/bambinos/bambi/blob/main/bambi/backend/pymc.py#L196-L198) in `vars_to_sample`, but the sigma info is preserved in `idata` under `constant_data`:

```Python
list(idata["constant_data"].keys())
# ['sigma']
```

If I then try to sample response variable I get this error:

```Python
mod.predict(idata, kind="response")

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[7], line 1
----> 1 mod.predict(idata, kind="response")

File [.../bambi/models.py:877], in Model.predict(self, idata, kind, data, inplace, include_group_specific, sample_new_groups)
    874 required_kwargs = {"model": self, "posterior": idata.posterior}
    875 optional_kwargs = {"data": data}
--> 877 posterior_predictive = self.family.posterior_predictive(
    878     **required_kwargs, **optional_kwargs
    879 )
[...]

PR #851, opened 2024-10-25
Check if  is an attribute before trying to access it

tomicapretto:
Resolves https://github.com/bambinos/bambi/issues/850. See the comment at https://github.com/bambinos/bambi/issues/850#issuecomment-2437705557.


tomicapretto:
The error is not related to our codebase, but to something not available anymore in `jax.interpreters` which is still being used by the current release of `oryx` (and modified here https://github.com/jax-ml/oryx/commit/5272a3d586db462528c4fd8d7eacb25582594bcc)

PR #856, opened 2024-11-10
Handle multivariate responses with HSGP

tomicapretto:
This PR does two things.

* Main: Make it possible to use HSGP with multivariate responses, such as a multinomial likelihood.
* Adjacent: Make multivariate families use two dims. So far, it created an extra dimension which was not equal to the response dimension.

There's something I would like to clarify about HSGP with multivariate responses. With this PR it's possible to do something like

```python
formula = "c(y1, y2, y3) ~ 0 + hsgp(x, m=30, c=2)"
hsgp_model = bmb.Model(formula, df, family="multinomial")

# Setting aliases for a nicer graph
hsgp_model.set_alias({"c(y1, y2, y3)": "result"})
hsgp_model.set_alias({"hsgp(x, m=30, c=2)": "hsgp"})
hsgp_model.build()
hsgp_model.graph()
```

and the graph will look like:

image

See all the dimensions of the response share the same priors for `hsgp_sigma` and `hsgp_ell`. Theoretically, it's possible to use a different coefficient for each response dimension. However, Bambi does not support that, and after some thought I decided it is fine that way. 

The implementation is _very_ complicated already, and it would be much more complicated if we decided to handle this. There's the `by` (and `share_cov`) argument in `hsgp()`, which could make us think we can use it for this purpose. However, that argument expects categories to be values within a given variable. In the multivariate family case, the dimensions are different columns (i.e. `y1`, `y2`, and `y3` in the example above). So, at least, we would need a special way to tell Bambi to handle things differently in this special case.

On top of that, a multivariate model with an HSGP is already a fairly complex model. To have a more granular control of HSGP, one should use a PPL like PyMC. 

I'm open to change my mind in the future, but for now, I think this is good enough.

### DIFF
--- a/bambi/backend/model_components.py
+++ b/bambi/backend/model_components.py
@@ -53,7 +53,7 @@ class DistributionalComponent:
             self.build_intercept(bmb_model)
             self.build_offsets()
             self.build_common_terms(pymc_backend, bmb_model)
-            self.build_hsgp_terms(pymc_backend)
+            self.build_hsgp_terms(bmb_model, pymc_backend)
             self.build_group_specific_terms(pymc_backend, bmb_model)
 
     def build_intercept(self, bmb_model):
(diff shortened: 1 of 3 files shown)


AlexAndorra (approved):
LGTM @tomicapretto , thanks a lot for adding this!
I'd vote to allow `share_cov=False` also for these cases, but I understand it's hard to implement. I'm happy to contribute this though, if you think it's not too big

Issue #813, opened 2024-05-27
Sampler parameters not used when calling JAX samplers


GStechschulte:
To change `bayeux` based sampler parameter values, we require passing `kwargs` like

```python3
kwargs = {
        "adapt.run": {"num_steps": 500},
        "num_chains": 4,
        "num_draws": 250,
        "num_adapt_draws": 250
}

blackjax_nuts_idata = model.fit(inference_method="blackjax_nuts", **kwargs)
```

However, if a user attempts to **also** pass `draws` and `tune`

```python3
kwargs = {
        "adapt.run": {"num_steps": 500},
        "num_chains": 4,
        "num_draws": 500,
        "num_adapt_draws": 500
}

blackjax_nuts_idata = model.fit(draws=250, tune=250, inference_method="blackjax_nuts", **kwargs)
```
the values for `draws` and `tune` are never used. Should we print to stdout that these values are not used? Or using some other means to make the user aware?

Of course, after sampling one can infer the number of chains, draws, etc. Just a thought.

PR #855, opened 2024-11-07
Pass kwargs to nutpie + create env.yml file

AlexAndorra:
Currently, some kwargs are not passed to Bayeux when fitting the model. This PR makes sure it does.
The only change is on line 270 of `bambi/backend/pymc.py` -- the rest is only formatting.

Also added an env file for creation with mamba. Ready for review!

PR #863, opened 2024-12-12
Fix for issue # 813 modified the way kwargs are handled

tanishy7777:
Solves issue #813 
 - [x] Make sure all test pass.
 - [x] Make sure your code passes `black`.
 - [x] Make sure your code passes `pylint`.


tomicapretto:
Hi @tanishy7777 thanks for working on this, and sorry for the delay in the response. I think we should do one of the following

* Omit a warning saying that some kwargs are not used by the JAX based sampler (as @GStechschulte mentions in the issue)
* Do nothing, which is the current behavior.

I'm not sure if the proposed changes are what we want. If I understand correctly, it will set up `kwargs["num_steps"]` to `draws` when it's not `None`, but would leave it unset when it's `None` (using whatever default value applies for that sampler in the library that implements it).

What if we have the following example:

```python
kwargs = {
        "num_draws": 500,
}

blackjax_nuts_idata = model.fit(draws=250, inference_method="blackjax_nuts", **kwargs)
```

The user is passing through `kwargs["num_draws"]` they want `500` draws but we then replace that with `250`, which is the one passed to `draws=250`. 

This example makes me think we should just omit a warning in that case, and respect the name required by the underlying sampler.

---

A separate comment, it's always good to add a test when one implements a change in behavior like this one. Just let me/us know if you would need help with that.

GStechschulte:
Thanks @tomicapretto for catching this. As an aside, we should make sure changes here are compatible with #855. I do not think there is any major conflict, but good to double check.

GStechschulte:
Hey @tanishy7777 any updates regarding my comment above? Thanks!

tanishy7777:
> Hey @tanishy7777 any updates regarding my comment above? Thanks!

Sorry for the late respons, I think I might have missed this. Will look into it!

tanishy7777:
> Hey @tanishy7777 any updates regarding my comment above? Thanks!

If I understand correctly, you are saying that we should handle `kwargs` on a case by case basis, based on which library is being used(that we get from the `inference_method` parameter). 

### Fix 1:
So, if some argument is passed in `kwargs` which is not used in the underlying library we emit a warning to the user? 

### Fix 2a:
Also that we should map the extra arguments like `draws` -> `num_draws` based on the underlying library? 

### Fix 2b:
And give a warning to the user when both `kwargs` and an extra argument is passed (as for draws in this case), saying that `num_draws` in kwargs is being overridden by the arguments(`draws`) passed by them?

As show in https://github.com/bambinos/bambi/pull/863#issuecomment-2544330496
where

```python
kwargs = {
        "num_draws": 500,
}

blackjax_nuts_idata = model.fit(draws=250, inference_method="blackjax_nuts", **kwargs)
```

should produce warning because we use `draws`  which overrides the `num_draws` from kwargs for `blackjax_nuts`

GStechschulte:
#919 will close this PR.

PR #876, opened 2025-01-22
[ENH] Add tqdm Progress Bar for model.predict()

### LINKED ISSUE #818: Progress bar for model.predict
Hello,

Thank you for this excellent package!

`model.predict` takes up to 30 minutes for my model. Is there a way to add a progress bar?

jgyasu:
Closes #818 

This PR adds a `tqdm` progress bar to visualise the progress during the running of `model.predict()` method. The progress bar is put around the loop in `_compute_likelihood_params()` method where `distributional_components` are getting processed.

- Passes `black`
- Passes `pylint`

Thanks :)

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -804,6 +804,7 @@ class Model:
         include_group_specific=True,
         sample_new_groups=False,
         num_draws=None,
+        random_seed=42,
     ):
         """Predict method for Bambi models
 
@@ -843,6 +844,8 @@ class Model:
         num_draws : int or None, optional
             Determines the number of draws to use for the posterior predictive distribution.
             If `None`, all available draws from the posterior are used.
+        random_seed : int
+            Seed for the random number generator.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


jgyasu:
Should I list `tqdm` in `pyproject.toml`?

tomicapretto:
Hi again @jgyasu and thanks for all the contributions!

I'm not sure it's easy to resolve #818. While what you're doing effectively adds a progress bar, it adds it at the distributional component level (i.e. parameters of a likelihood for which we have at least one predictor). That is too high level. The user will see one or two "steps" and that's it.

The real difficulty is in what happens in the `.predict()` method (see here https://github.com/bambinos/bambi/blob/6b66691ed08d88d98acbb5a229f5aeb258e7ab44/bambi/model_components.py#L128). There are several steps that can take time (`predict_common`, `predict_group_specific`, etc.)

Another pain point, and I think it's the one that usually takes the longest, is what happens when users do `model.predict(... kind="response")`. After computing the draws for the response parameters (e.g. mu and sigma in gaussian models), it uses `pymc.draw()` which usually takes some time, especially when we pass large arrays (that happens here https://github.com/bambinos/bambi/blob/6b66691ed08d88d98acbb5a229f5aeb258e7ab44/bambi/families/family.py#L161)

It's not that I want to ruin the party, but I think this one is not possible to be resolved until we change how we get predictions, which would be through the proper usage of `pymc.sample_posterior_predictive()`. But for that to happen, we first need to modify other internal things.

PR #892, opened 2025-04-03
Add random_seed argument to predict

aloctavodia:


tomicapretto:
I guess we need to add another dependency?

aloctavodia:
Bayeux may have to pin flowMC version or update their code. @ColCarroll?

aloctavodia:
Can we merge this as the test failure is unrelated to the PR?

tomicapretto:
Makes sense

PR #896, opened 2025-05-01
Add R2

aloctavodia:
This adds the Bayesian R2, as defined in https://doi.org/10.1080/00031305.2018.1549100 and as computed by ArviZ.

In the "new arviz", we already have other metrics like loo_rmse, and we will soon have loo_R2.

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -11,6 +11,7 @@ import pymc as pm
 import pandas as pd
 
 from arviz.plots import plot_posterior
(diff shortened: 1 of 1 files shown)


tomicapretto:
@aloctavodia thanks for this. Do we need to upgrade the minimum version of ArviZ we require?

aloctavodia:
No need for this PR. We will need to update the version to use the other metrics.

PR #902, opened 2025-05-15
Add link to the book "Causal Inference: What if"

star1327p:
Add link to the book "Causal Inference: What if"
https://miguelhernan.org/whatifbook


tomicapretto:
Thanks again! :)

PR #903, opened 2025-05-18
Updated links to PyMC and Bambi pages

star1327p:
Updated links to PyMC and Bambi pages


tomicapretto (approved):
Thanks very much again!

PR #913, opened 2025-08-02
DOC Updated links in Robust Linear Regression

star1327p:
Updated two links to pymc.io in the Robust Linear Regression example.


tomicapretto:
Thanks!!

PR #923, opened 2025-09-18
Use mock sampling in tests and incorporate other improvements to the test suite.

tomicapretto:
In this PR I'm using `mock_pymc_sample` in many cases to speed-up our test suite.
The version we need is still not released; there's a PR opened in the PyMC repo: https://github.com/pymc-devs/pymc/pull/7906.

I'm also considering:

 * Migrate simulated datasets and perhaps other utils to `conftest.py`
 * Compile and evaluate logp and dlogp of all models. This will add some overhead, but I think it's required for making our tests more robust.


tomicapretto:
I think there's still room for improvement in `tests/test_models.py`... but this was enough for me :)

I'm going to move into other things, like documentation building and CI.

williambdean:
What was the speed up? 😄

tomicapretto:
Hard to really tell, because before we would have our tests failing for a long time. 
But, last time I see is 22 minutes for end-to-end CI, and now it was 15 minutes. We would have had a much greater speedup if I didn't include dlogp compilation and evaluation at the initial point, but I included it because I think it's safer to have those. 
Anyway, locally it takes around 4 minutes and that is way better than what I had before (which I don't remember, but always hated how much I had to wait).

Now I'm looking to speed-up Bambi's CI by modernizing our workflows :)

@williambdean

In #925, we're around ~13 minutes... which is way better than what we had before (of course I'd love to have it <2 mins, but it doesn't seem to be possible for now)

PR #925, opened 2025-09-27
Modernize development infraestructure

tomicapretto:


ahartikainen:
I personally like ruff, but ofc I'm much less restrictive what goes in to the codebase.

And it is not about the speed but less tools (ruff format + ruff check).

tomicapretto:
> I personally like ruff, but ofc I'm much less restrictive what goes in to the codebase.
> 
> And it is not about the speed but less tools (ruff format + ruff check).

@ahartikainen do you have an example at hand I can have a look at? I'm curious about how people set it up (I'm trying to find if I can eventually get convinced that ruff is all I need)

PR #929, opened 2025-09-29
Fix build status badge

tomicapretto:
It was pointing to the now old `test.yml`


tomicapretto:
Good to see this does not run CI or anything, given it doesn't modify the relevant files.

Although some checks (i.e. trailing whitespaces, end of file) should be included somehow... argh

Issue #484, opened 2022-04-11
Explain how hierarchical models work in Bambi


tomicapretto:
Include a comparison with how we define hierarchical models with PyMC too.

speco29:
Hi @tomicapretto , I would like to work on this issue. Can you just tell me whether should I add this explanation, I'm thinking it would be here bambi/models.py, or should I add another py file for it?

aloctavodia:
This should be an example similar to those in the gallery of examples https://bambinos.github.io/bambi/notebooks/
The source file should be a notebook and should be here https://github.com/bambinos/bambi/tree/main/docs/notebooks

speco29:
I saw the docs and found there are already working hierarchial models in the examples.
I'm bit confused about what should I add?
Can you brief me a bit so that i can work on it?

pranavduraisamy:
Hi @aloctavodia @tomicapretto , Just wanted to check if anyone is currently working on it or is it resolved already. I'd be happy to help out. Let me know!

Thanks!

aloctavodia:
@pranavduraisamy thanks for offering help. I don't think this has been solved, although I remember @tomicapretto showing me some side by side comparison of pymc and bambi. But was for a talk or course, I don't think that was integrated into the docs

pranavduraisamy:
@aloctavodia thanks for the reply, currently I'm thinking of 2 examples, one similar to the reed frogs example in statistical rethinking and other is updating the 'pigs dataset' with pymc... 

Also, having the comparison of the prior diagnostics (from arviz-plots/stats, [Sensitivity checks](https://arviz-devs.github.io/EABM/Chapters/Sensitivity_checks.html)) would be better ig...

aloctavodia:
Explaining how hierarchical models work and prior/likelihood sensitivity analysis are two different topics, and thus should be discussed in separated notebooks/examples

PR #939, opened 2025-10-27
Updated Hierarchical Linear Regression (Pigs dataset) example (#484)

pranavduraisamy:
- Updated with a PyMC replica of original Bambi model
 - Rewritten model definition with widely-used terms for slope and intercept for better clarity
 - Added an EDA and posterior regression lines for each group (/pig), so that the reader could understand the problem scenario and its results by using a partially pooled model
 - Address issue #484


tomicapretto:
@pranavduraisamy, thanks a lot for the contribution. I'll review in detail in a moment. I'm not so sure we can close the linked issue though, as I was thinking about a more fundamental explanation, not a "this is how it looks in PyMC vs this is how it looks in Bambi".

tomicapretto:
@pranavduraisamy just left a review, thanks again for the contribution. My last two points:

* Why are we using `draws=2000`? If there's no specific reason, I would leave the default of `1000`, unset.
* Could you blackify all the code? We use black in our codebase and I think the notebooks should be in line with that for consistency and legibility.

pranavduraisamy:
@tomicapretto Thanks for the review, this is my first time contributing to an open source library. 

> @pranavduraisamy, thanks a lot for the contribution. I'll review in detail in a moment. I'm not so sure we can close the linked issue though, as I was thinking about a more fundamental explanation, not a "this is how it looks in PyMC vs this is how it looks in Bambi".

Are you expecting a more basic explanation for hierarchical modelling... could you give me a little more idea on that.,

tomicapretto:
@pranavduraisamy, apologies for the delay. 
What I mean in the issue is something along the lines of this (unfinished) document https://github.com/bambinos/bambi/blob/main/docs/notebooks/how_bambi_works.ipynb.

I am going to review your notebook again, and if needed, take it to the finish line so we can merge it.

PR #943, opened 2025-11-15
Remove tan_2 link function for vonmises

aloctavodia:
PyMC VonMises distribution already converts values of mu outside of the range [-pi, pi) to that range. So, not sure why we have that link function, maybe it was not the case in the past? I don't think so. Anyway, while testing various R2 implementations in ArviZ, it gets inconsistent results when using the `tan_2` link function, but consistent results when using the `identity` function.

### DIFF
--- a/bambi/model_components.py
+++ b/bambi/model_components.py
@@ -176,9 +176,8 @@ class DistributionalComponent:
 
         # NOTE: Handle VonMises family that internally uses 'identity' but needs angles
         if self.spec.family.name == "vonmises":
-            invlink = lambda x: np.angle(
-                np.exp(1j * x)
-            )  # pylint: disable=unnecessary-lambda-assignment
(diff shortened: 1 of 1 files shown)


tomicapretto:
@aloctavodia, if I rerun the vonmises example that we have [here](https://bambinos.github.io/bambi/notebooks/circular_regression.html)

I get this plot

image

Do you know what's going on?

tomicapretto:
With that said, the last plot in the example does not look so bad (again using Bambi from this branch)

image

aloctavodia:
We are getting the predictions on the unwrapped space, instead of getting them from the circle [-pi, pi). I guess this is related to how we generate predictions "manually". I can check

tomicapretto:
One more interesting thing...

image


I'm fixing and pushing it

PR #948, opened 2025-12-08
interpret refactor

GStechschulte:
This draft PR introduces an initial refactor of the interpret submodule. The goals are to:

- Simplify the plotting control flow
- Simplify how comparisons and slopes are computed
- Standardize how summary data frames are built
- Bridge functionality between interpret and marginaleffects

This is done by adopting a more functional design:

- Structural pattern matching is used to de-structure and validate the conditional and contrast parameters for the predictions, comparisons, and slopes functions.
- These parameters now return a Pandas Series, which is used to declaratively build data grids (with consistent type casting) and summary data frames (via Pandas split/apply/combine).
- Plotting uses the Seaborn objects API to declaratively define the plotting specification, with a PlotConfig type to configure main, group, and panel figures.

To do:

- [x] Add slopes functions
- [x] Add figure-level kwargs configuration (e.g., `sharex=True`)
- [x] Add subplot kwargs (e.g., specify the main x-axis, group/color, and panel/facet).
- [x] Add new tests for added functionality (e.g., comparison functions)
- [x] Expose interpret helper functions? (_these are removed as we now return the posterior samples_)
- [ ] Update logger


tomicapretto:
This looks very promising, thanks! Should I expect existing tests and/or examples to work? Also, I notice you used `uv.lock`, do you think `uv` can help us with anything? Do you need any help with pixi?

GStechschulte:
Thanks @tomicapretto. Yes, once this is no longer a draft PR all existing tests and examples should run. No, I forgot to add `uv.lock` to my local gitignore. Moreover, I used uv instead of pixi because I have not worked with pixi before hah.

aloctavodia:
Hi, this looks great! Have you considered using arviz-plots here? That will most likely require some changes to current ArviZ capabilities, but it can potentially benefit both projects. This suggestion is partially independent of this PR, as we could do it later.

bambi/interpret/effects.py
```diff
@@ -1,890 +1,857 @@
...
 
-@dataclass
-class ResponseInfo:
-    """Stores metadata about the response variable for indexing data in az.InferenceData,
-    computing uncertainty intervals, and creating the summary dataframe in
-    'PredictiveDifferences'
+def validate_values(
+    values: Values, var_name: str, target_dtype: np.dtype | None = None
+) -> Series:
+    """Validate and convert user provided values to a Pandas Series.
```
tomicapretto: Are we converting to a pandas series because that is what seaborn prefers/expects? Have you encountered any serious drawback because of having to work with Series?

bambi/interpret/effects.py
```diff
@@ -1,890 +1,857 @@
...
+        If values are not a list, ndarray, or Series, or if they cannot be converted
+        to the target dtype.
+    ValueError
+        If the values container is empty.
     """
+    match values:
+        # Values can be a list
+        case list() as lst:
+            # Validate all elements are int or float
+            if not all(isinstance(x, (int, float)) for x in lst):
```
tomicapretto: Why does this work with only numeric values? Is there anything separate that handles the validation for non-numeric values (e.g., levels of a categorical variable)?

bambi/interpret/effects.py
```diff
@@ -1,890 +1,857 @@
...
-# pylint: disable=consider-iterating-dictionary
-# pylint: disable=too-many-instance-attributes
-@dataclass
-class PredictiveDifferences:
-    """Computes predictive differences and their uncertainty intervals for
-    'comparisons' and 'slopes' effects and returns a summary dataframe of the results
+    match series.dtype:
+        case pd.CategoricalDtype():
+            # Takes the first mode if there are multiple modes
+            return pd.Series(series.mode().iloc[0], name=var)
```
tomicapretto: I remember being fooled by `pd.Series.mode` in the past, when we were computing modes by group and the group had either zero rows or all nulls... Just raising some validation may be needed somewhere (perhaps outside of this, I don't know).

```python
import pandas as pd
x = pd.Series([None, None], dtype=pd.CategoricalDtype())
x.mode().iloc[0] # Errors

x = pd.Series([], dtype=pd.CategoricalDtype())
x.mode().iloc[0] # Errors
```

The error occurs because of iloc, not because of mode. That is what's surprising.

bambi/interpret/effects.py
```diff
@@ -1,890 +1,857 @@
...
+        Probability for the credible interval (between 0 and 1).
+    transforms : Callable
+        Function to transform the data before computing statistics.
+
+    Returns
+    -------
+    DataFrame
+        A DataFrame containing the estimate (mean) and lower/upper bounds of the credible interval.
+    """
+    x = transforms(x)
```
tomicapretto: Do we need to have `transforms` as a parameter of the function? Asking since this is the only place it's used... one could ask "the user" to pass transformed values.

bambi/interpret/utils.py
```diff
@@ -1,222 +1,171 @@
...
+
+    Returns
+    -------
+    tuple[str, str or None]
+        A tuple containing the response name and the target parameter name.
+        If target is 'mean', returns the response name and the parent parameter.
+        Otherwise, returns the component alias (or response name) and the target (or None).
+    """
+    match target:
+        case "mean":
```
tomicapretto: I have the recollection that we're going for family-specific parameter names, but I guess we allow for this `"mean"` so we can have it as a default upstream? What about `target` being `str | None` where `None` means whatever is the parent of the family? (I don't love it tbh)

tomicapretto (commented):
I'm not sure I fully understand everything, but I really like where this is going.

Is the anything else I can check to help you @GStechschulte?

bambi/interpret/effects.py
```diff
@@ -1,890 +1,857 @@
...
-        Conditional covariates with their names and values (if any).
-    response : ResponseInfo
-        Response variable with its name and target.
-    use_hdi : bool
-        Whether to use the highest density interval (HDI) (True) or quantiles (False).
-    kind : str
-        Type of effect to compute. Either 'comparisons' or 'slopes'.
+    contrast : ContrastParam
+        The contrast specification, either as a string (variable name) or
+        a dictionary mapping variable name to values.
```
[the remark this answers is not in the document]
GStechschulte: Yeah, this should be removed. I removed it from the function signature in `comparisons` and `plot_comparisons`

bambi/interpret/effects.py
tomicapretto: Perfect

bambi/interpret/effects.py
```diff
@@ -1,890 +1,857 @@
...
-        if is_categorical_dtype(model.data[contrast_name]) or is_string_dtype(
-            model.data[contrast_name]
-        ):
-            num_levels = len(model.data[contrast_name].unique())
-            if num_levels > 2:
-                raise ValueError(
-                    f"'conditional' must be specified when 'contrast' has more than 2 values. "
-                    f"{contrast_name} has {num_levels} unique values."
-                )
+    plot(summary_df, plot_config)
```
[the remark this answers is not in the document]
GStechschulte: We should for sure return the plot because the user can then customize the plot further (with Seaborn and or Matplotlib) and or save it. 

I think maybe we should call `plot.show()` implicitly inside of the plot function too because some REPLs for example don't recognize the plot object and automatically plot it, for example, Marimo. By calling `show` internally, we ensure the plot is always shown?

GStechschulte:
> I'm not sure I fully understand everything, but I really like where this is going.
> 
> Is the anything else I can check to help you @GStechschulte?

Thanks! What specifically do you not understand? I am more than happy to provide an explanation :)

bambi/interpret/effects.py
GStechschulte: Categorical values were validated within the `parse_conditional` function. After seeing this comment, I moved out the validation of categorical and numerical (float and integers) into separate functions.

Now, `get_contrast` and `get_conditional` will dispatch function calls to the appropriate validation functions depending on the series data type.

bambi/interpret/effects.py
tomicapretto: What I meant was that Bambi is the only user of `get_summary_stats`, and one could do

```python
stats_data = get_summary_stats(response_transform(y_hat), prob)
```

in line 632 of `effects.py`, within `predictions` definition, and

```python
    summary_draws = {
        k: get_summary_stats(response_transform(v), prob)
        for k, v in compared_draws.items()
    }
```

in line 850 of `effects.py`, within `comparisons` definition; instead of what is done now.

Independent of the function being user-facing or not, the transformation is only used to do `x = transforms(x)`, so I don't see why it is needed within the function body.

bambi/interpret/effects.py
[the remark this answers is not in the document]
tomicapretto: I agree!

bambi/interpret/logs.py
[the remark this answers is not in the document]
tomicapretto: Sounds good!

bambi/interpret/plots.py
```diff
@@ -0,0 +1,176 @@
...
+    panel : str or None
+        Optional faceting variable for creating subplots.
+    """
+
+    main: str
+    group: str | None = None
+    panel: str | None = None
+
+
+def create_plot_config(
```
tomicapretto: I love this one

bambi/interpret/utils.py
```diff
@@ -230,8 +104,17 @@ def get_model_terms(model: Model) -> dict:
 
 
 def get_model_covariates(model: Model) -> np.ndarray:
-    """
-    Return covariates specified in the model.
+    """Return covariates specified in the model.
```
tomicapretto: Reading through this function I'm wondering whether we're making it unnecessarily complicated by appending lists to `covariates` and then flattening it. Wouldn't it be more direct to just append the names themselves? I don't remember why we did things this way.

bambi/interpret/effects.py
GStechschulte: Ah yes. Agreed!

GStechschulte:
One of the nice things about returning a Seaborn Plot type is the ability for the user to further customize and or add layers. For example, we can easily add a layer for the observed data (example from the orthogonal polynomial regression notebook).

```python
# This plot is built using the interpret summary dataframe
plot = bmb.interpret.plot_predictions(
        mpg_hp_sq_mod,
        mpg_hp_sq_fit,
        "horsepower",
        pps=True,
        prob=p,
    )

# Add another layer using a separate dataframe
plot.add(so.Dots(), data=df_mpg, x="horsepower", y="mpg")
```

PR #950, opened 2025-12-11
Support sparse design matrix for group specific effects

### LINKED ISSUE #545: Use a sparse matrix for group specific effects
Background: I recently observed large RAM usage on a model, and opened an issue on pymc discourse, and in that thread Tomás Capretto responded:

"On the other hand, Bambi relies on formulae to generate design matrices, which turns out to generate a regular matrix for group-specific effects. In this case, it’s a very large matrix with almost all zeros, so a sparse matrix would have been better. So I think this explains the large memory comsumption. It is something we still need to improve on our end. However, this matrix is not directly used in the PyMC model, we use slicing to select only non-zero values."

Today this issue hit me again, and this time the RAM requirements were absurd:

```
my_model = bmb.Model("deprived.of.education ~ (sex|country) +    (sex|cluster.id.unique) + per.cent.muslim.in.country*sex*gdp.log + per.cent.hindu.in.country*sex*gdp.log + per.cent.muslim.in.cluster*sex*wealth.at.cluster.level + per.cent.hindu.in.cluster*sex*wealth.at.cluster.level + wealth * sex * religion + urbrur", df, family="bernoulli", dropna=True)

Automatically removing 739811/2772473 rows from the dataset.
Unexpected error while trying to evaluate a Variable. <class 'numpy.core._exceptions._ArrayMemoryError'>

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\hanse\miniconda3\lib\site-packages\bambi\models.py", line 147, in __init__
    self._design = design_matrices(formula, data, na_action, 1, extra_namespace)
  File "C:\Users\hanse\miniconda3\lib\site-packages\formulae\matrices.py", line 523, in design_matrices
    design = DesignMatrices(description, data, env)
  File "C:\Users\hanse\miniconda3\lib\site-packages\formulae\matrices.py", line 54, in __init__
    self.model.eval(data, env)
  File "C:\Users\hanse\miniconda3\lib\site-packages\formulae\terms\terms.py", line 1261, in eval
    term.set_data(encoding)
  File "C:\Users\hanse\miniconda3\lib\site-packages\formulae\terms\terms.py", line 665, in set_data
    self.factor.set_data(True)  # Factor is a categorical term that always spans the intercept
  File "C:\Users\hanse\miniconda3\lib\site-packages\formulae\terms\terms.py", line 468, in set_data
    component.set_data(spans_intercept_)
  File "C:\Users\hanse\miniconda3\lib\site-packages\formulae\terms\variable.py", line 109, in set_data
    self.eval_categoric(self._intermediate_data, spans_intercept)
  File "C:\Users\hanse\miniconda3\lib\site-packages\formulae\terms\variable.py", line 166, in eval_categoric
    value = self.contrast_matrix.matrix[x.codes]
numpy.core._exceptions._ArrayMemoryError: Unable to allocate 857. GiB for an array with shape (2032662, 113158) and data type int32
```

There are two random slopes in the model, the first grouping factor has 88 levels, and the second grouping factor has 145861 levels, so the model in itself is not very big. I am happy to provide the data if it is useful.

tomicapretto:
This pull request does a few things:

* Adds a boolean `SPARSE_DOT` configuration variable to `bambi.config`, defaulting to `False`. When `True`, Bambi will use `pytensor.sparse.structured_dot` for the dot product between `Z` and `u`, the design matrix and vector of coefficients for group-specific effects. By default, the behavior is unchanged.
* Slightly refactors how terms are built in the backend. I think they are clearer and more robust now, as the shapes of random variables are clearer throughout the build process.
* Adds tests for the new configuration variable. They check both that interaction with the configuration variable works as expected, as well as whether setting `SPARSE_DOT` to `True` produces the same model as when it is `False` (in terms of logp and dlogp).
* Other minor changes are commented on here in the pull request.

This is possible now because formulae uses sparse matrices in CSR format for the group-specific effects.

Separate note: while working with the `*Term` classes in the backend (those under `bambi/backend/terms.py`), I started to wonder whether it makes sense to have them as classes, since we instantiate them only to call `.build()` on them. It may make sense to have them as functions instead. But I’m not sure about that, and if we decide to do it, it should be in a separate PR.

Closes #545

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -1351,6 +1351,65 @@ def test_predict_new_groups(data, formula, family, df_new, request, mock_pymc_sa
     model.predict(idata, data=df_new, sample_new_groups=True)
 
 
+@pytest.mark.parametrize(
+    "data,formula,family,df_new",
+    [
+        (
+            "data_sleepstudy",
+            "Reaction ~ 1 + Days + (1 + Days | Subject)",
+            "gaussian",
+            pd.DataFrame({"Days": [1, 2, 3], "Subject": ["x", "y", "z"]}),
+        ),
+        (
+            "data_inhaler",
+            "rating ~ 1 + period + treat + (1 + treat|subject)",
+            "categorical",
+            pd.DataFrame(
+                {
+                    "subject": [1, 999],
+                    "rating": [1, 1],
+                    "treat": [0.5, 0.5],
+                    "period": [0.5, 0.5],
+                    "carry": [0, 0],
+                }
+            ),
+        ),
+        (
+            "data_crossed",
+            "Y ~ 0 + threecats + (0 + threecats | subj)",
+            "gaussian",
+            pd.DataFrame({"threecats": ["a", "a"], "subj": ["0", "11"]}),
+        ),
+    ],
+)
+def test_predict_new_groups_deterministic(data, formula, family, df_new, request, mock_pymc_sample):
+    data = request.getfixturevalue(data)
+    model = bmb.Model(formula, data, family=family)
+    model.build()
+    assert_ip_dlogp(model)
+    idata = model.fit(chains=2)
+
+    pred1 = model.predict(idata, data=df_new, sample_new_groups=True, random_seed=42, inplace=False)
+    pred2 = model.predict(idata, data=df_new, sample_new_groups=True, random_seed=42, inplace=False)
+    pred3 = model.predict(
+        idata, data=df_new, sample_new_groups=True, random_seed=123, inplace=False
+    )
+
+    param_name = list(model.distributional_components.keys())[0]
+    if model.distributional_components[param_name].alias:
+        param_name = model.distributional_components[param_name].alias
+
+    # Extract the parameter values
+    param1 = pred1.posterior[param_name].values
+    param2 = pred2.posterior[param_name].values
+    param3 = pred3.posterior[param_name].values
+
+    assert np.allclose(param1, param2), "Predictions with same random_seed should be identical"
+    assert not np.allclose(
+        param1, param3
+    ), "Predictions with different random_seed should be different"
+
+
 def test_weighted(mock_pymc_sample):
     weights = 1 + np.random.poisson(lam=3, size=100)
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -23,11 +23,12 @@ maintainers = [
 
 dependencies = [
     "arviz_plots @ git+https://github.com/arviz-devs/arviz-plots",
-    "formulae>=0.5.3",
+    "formulae @ git+https://github.com/bambinos/formulae",
     "graphviz",
     "matplotlib>=3.9.0",
     "pandas>=1.0.0",
     "pymc>=5.26.1",
+    "sparse>=0.17.0,<0.18",
```
tomicapretto: xarray does not work with sparse matrices from scipy, thus we need this extra dependency.

tests/test_models.py
tomicapretto: These are just some simplifications and updates for clarity.

bambi/terms/group_specific.py
```diff
@@ -11,20 +11,18 @@ def __init__(self, term, prior, prefix=None):
         self._hyperprior_alias = {}
         self.term = term
         self.prior = prior
-        self.data = np.squeeze(term.data)
+        self.data = term.data
```
tomicapretto: If we squeeze a CSR matrix, we get a numpy array containig a CSR matrix :')

docs/notebooks/mister_p.ipynb
tomicapretto: I updated this one when I was checking my code worked. I pushed the updates because it contained outdated choices.

pyproject.toml
```diff
@@ -23,11 +23,12 @@ maintainers = [
 
 dependencies = [
     "arviz_plots @ git+https://github.com/arviz-devs/arviz-plots",
-    "formulae>=0.5.3",
+    "formulae @ git+https://github.com/bambinos/formulae",
```
tomicapretto: Needs to be updated before making a release (for which we need a formulae release)

tomicapretto:
Tests are failing because of the HTTP 403 error

Issue #953, opened 2026-01-14
Non-deterministic predictions with `sample_new_groups=True` despite setting `random_seed`


Delogon:
Hi 
after spending some hours to create a reproducible analysis, I stumbled on something I think is a bug. 

When calling `model.predict()` with `sample_new_groups=True` and a `random_seed` parameter, predictions are non-deterministic. Multiple calls with the same seed produce different results, making it impossible to reproduce predictions for new groups.

## Steps to Reproduce

```python
import numpy as np
import pandas as pd
import bambi as bmb

# Create dataset with grouped data
np.random.seed(123)
data = pd.DataFrame({
    'y': np.random.randn(100),
    'x': np.random.randn(100),
    'group': np.repeat(['A', 'B', 'C', 'D'], 25)
})

# Fit model with random effects
model = bmb.Model('y ~ x + (1|group)', data)
idata = model.fit(tune=500, draws=500, random_seed=123)

# Create new data with unseen group
new_data = pd.DataFrame({
    'x': [0.5, 1.0, 1.5],
    'group': ['E', 'E', 'E']  # New group not in training data
})

# Make predictions with same random_seed multiple times
pred1 = model.predict(idata, data=new_data, kind='response_params',
                      sample_new_groups=True, random_seed=123)
pred2 = model.predict(idata, data=new_data, kind='response_params',
                      sample_new_groups=True, random_seed=123)

# Check if results are identical
mu1 = pred1.posterior['mu'].values
mu2 = pred2.posterior['mu'].values
print(f"Results identical: {np.allclose(mu1, mu2)}")
```

I think the bug is in the `_construct_u_with_new_groups()` method in `bambi/model_components.py` at line 367:

```python
rng = np.random.default_rng()
```

This method creates a new random number generator without any seed, even though the `random_seed` parameter was provided to `model.predict()`. The `random_seed` was never propagated through the call chain to reach the method that actually performs the random sampling for new groups.

I will create a pull request that fixes this, if this is not intended.

tomicapretto:
@Delogon: thanks for both the issue report and the PR, let's continue the chat over there :)

PR #954, opened 2026-01-14
Fix non-deterministic predictions when `sample_new_groups=True` and `random_seed` is set.

Delogon:
This should fix the issue #953. I ran all tests locally and they passed. I had to download the datasets manually because the AWS WAF is blocking the automatic downloads. 

It is a very simple fix, just passing the random seed to the rng in the `_construct_u_with_new_groups()`. I hope the test is fine like this. Otherwise, please let me know.

### DIFF
--- a/bambi/data/datasets.py
+++ b/bambi/data/datasets.py
@@ -3,11 +3,11 @@
 import hashlib
 import itertools
 import os
+import pathlib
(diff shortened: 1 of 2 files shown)


tomicapretto:
thanks @Delogon !

PR #958, opened 2026-01-21
Auto-detect sparse matmul for JAX backends

korentomas:
I was looking into how PyMC was taking advantage of MLX and found [this thread](https://discourse.pymc.io/t/how-to-use-pymc-with-mlx/17394/9) so I ended up trying to fix what @tomicapretto (Hola!) had mentioned:
> This is one of the cases where I want to change how Bambi works under the hood.
It would be great if Bambi did a (sparse) matmul under the hood for the group-specific effects (also known under many other names). But it does not do it right now, it does some indexing thing to avoid multiplying a large dense matrix. My guess is that if you do implement it with a matmul, perhaps even dense, it should be faster in mlx. 

The sparse matmul for group-specific effects was implemented in PR #950 by @tomicapretto, this PR makes it automatically enabled for JAX backends.

**Changes:**
- Add `sparse_dot` parameter to `Model()` for explicit control
- Auto-detect JAX backends in `fit()` and enable sparse matmul
- Add `None` to valid `config.SPARSE_DOT` values for auto-detect mode

**Priority:** model param > `config.SPARSE_DOT=True` > auto-detect

### DIFF
--- a/bambi/backend/model_components.py
+++ b/bambi/backend/model_components.py
@@ -6,10 +6,8 @@ from pytensor import sparse as ps
 
 from bambi.backend.terms import CommonTerm, GroupSpecificTerm, HSGPTerm, InterceptTerm, ResponseTerm
 from bambi.backend.utils import get_distribution_from_prior
-from bambi.config import config as bmb_config
 from bambi.families.multivariate import MultivariateFamily
 from bambi.families.univariate import Categorical, Cumulative, StoppingRatio
-
 
 ORDINAL_FAMILIES = (Cumulative, StoppingRatio)
 
@@ -154,7 +152,7 @@ class DistributionalComponent:
             predictors.append(term.predictor)
             group_indexes.append(term.group_index)
 
-        if bmb_config["SPARSE_DOT"]:
+        if bmb_model.sparse_dot:
             coefs_reshaped = []
             for coef in coefs:
                 if as_multivariate and coef.ndim == 3:
--- a/bambi/backend/utils.py
+++ b/bambi/backend/utils.py
@@ -205,19 +205,22 @@ def make_weighted_distribution(dist: pm.Distribution):
 
 
 def should_use_sparse_dot(inference_method=None):
-    """Determine whether to use sparse matrix multiplication for group effects.
+    """Determine whether to use sparse matrix multiplication for group-specific effects.
 
-    Sparse matmul is faster on JAX/GPU backends but may be slower on CPU backends.
+    Sparse matrix multiplication is faster on JAX/GPU backends (numpyro, blackjax) but may be
+    slower on CPU backends (pymc, nutpie). This function is used for auto-detection when the
+    user has not explicitly set the `sparse_dot` parameter.
 
     Parameters
     ----------
-    inference_method : str, optional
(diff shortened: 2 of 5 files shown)


tomicapretto:
Hola @korentomas, bienvenido!

Thanks for the contribution, I just approved the CI run. Let me add a few thoughts:

* PyTensor does not support sparse dot product for the MLX backend yter (there's no sparse backend [here](https://github.com/pymc-devs/pytensor/tree/main/pytensor/link/mlx/dispatch)).
* PyTensor does support a sparse dot product for the JAX backend, see [here](https://github.com/pymc-devs/pytensor/blob/main/pytensor/link/jax/dispatch/sparse.py), but not extensively, as there are Ops to be implemented, such as the gradients. Also, it's only using the very generic BCOO data structure. We would benefit more from CSR and its csr_matvec operation.
* In my experience, one starts seeing the advantage of using dot(sparse-matrix, vector or matrix) when sparse-matrix is big and sparse. Do you have any benchmark or example where you can conclude that the sparse dot product is always better for those backends?
* I almost forgot this, but, using the JAX backend does not mean that you're using a GPU... So, even if it was faster on those backends because of GPUs, we can't tell from the backend only if sampling happens in GPU.
* I don't think it's a good idea to make `sparse_dot` an attribute of the model. If something, it could be an argument of the `.build()` or `.fit()` methods. But, anyway, I still think the configuration variable is the best way to handle this for now.

I'm sorry if this comes a bit negative for the PR, since after all I'm saying "well, we can't move on with the change as it is". But reading through your contribution gave me two ideas for how it could continue:

1. We can make the configuration more versatile, allowing it to be used in combination with a context manager. That way, you could do something like...
    ```python
        with bmb.Config(sparse_dot=True):
            model = bmb.Model(...)
            idata = model.fit()
    ```
which is similar to how one can use the config in polars. Personally, I find it useful when I need to display more rows in a certain table.
2. Add an example showing how to use SPARSE_DOT. The simplest could be a simulation where you have a normal outcome, a factor covariate with many levels, and you build a hierarchical model. Something along the lines of `y ~ 1+ (1 | group)`. If you want to add a slope, it can be `y ~ 1 + x + (1 + x | group)`, although the simulation can be a bit more complex. That way, you could show how sampling becomes faster as you have more observations and groups.

If you have any questions, need help, etc. just let me know!

korentomas:
Thanks for the feedback @tomicapretto! Sounds good, I'll take the context manager approach. For the benchmark I am going to do simulated data but I could also try with the `batting` dataset. It can be something like 
`model = bmb.Model("HR ~ yearID + (1 | playerID)", df)`
or
`model = bmb.Model("HR ~ yearID + (1 | playerID) + (1 | teamID)", df)` with crossed random effects

tomicapretto:
> Thanks for the feedback @tomicapretto! Sounds good, I'll take the context manager approach. For the benchmark I am going to do simulated data but I could also try with the `batting` dataset. It can be something like `model = bmb.Model("HR ~ yearID + (1 | playerID)", df)` or `model = bmb.Model("HR ~ yearID + (1 | playerID) + (1 | teamID)", df)` with crossed random effects

Actually, yes, it's a good idea to use that since it's a real dataset. Just don't be too optimistic because the data size is not too big (IIRC).

PR #965, opened 2026-02-06
[Proposal] Fixed, Random Effects and Mundlak Machines Example

juanitorduz:
Hi! I want to suggest adding the port to Bambi of [Fixed and Random Effects Models: A Simulated Study](https://juanitorduz.github.io/fixed_random/) based on the video [Statistical Rethinking 2026 Lecture B04 - Group-level confounding and intro to social networks](https://www.youtube.com/watch?v=XNNcN8sU8us). I think this is a great exposition of the subject (and I also discovered the Mundlak models, which work great). Happy to get feedback!

BTW: I see `*.ipynb` is part of the `.gitignore`, is this a feature or a bug? Also, the pre-commit complains about this notebook being too large ¯\\_(ツ)_/¯

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -145,7 +145,7 @@ The documentation is built using [Quarto](https://quarto.org/) with [quartodoc](
    quarto render docs/
    ```
 
-   > **Tip:** Rendering the full site executes all example notebooks, which can take a significant amount of time.
+   > **Tip:** Rendering the full site renders all example notebooks from scratch, which can take a significant amount of time.
    > To render a single notebook, use `quarto render docs/notebooks/<notebook_name>.ipynb`.
 
    **Preview general docs content** (notebooks, `.qmd` pages — no API reference changes):


tomicapretto:
@juanitorduz, thanks for the contribution, this is a great idea. I'll review the notebook later. In the meantime, I can reply to the other two points:

* gitignore: it's on purpose, it should git ignore notebooks only in the root directory... unless I always got it wrong?
* file size: I'll update pre-commit config so we have a different value for notebooks (I hope that is possible)

juanitorduz:
I added the notebook in the examples folder and it was still ignored. I had to delete the gitignore line to push the nb. I think we need to refine it :)

tomicapretto:
@juanitorduz thanks again for the example. I've just modified a few things (gitignore, pre-commit config) and made sure the example appears in our examples section. If you want to choose a thumbnail, feel free to do so.

The only issue I notice is that plot sizes are too large, see horizontal bar in the following screenshot:

image

If you can't adjust it, I can do it tomorrow :)

juanitorduz:
Thank you @tomicapretto and @aloctavodia ! I added some changes:

- Tried to make the plots much smaller. However, the model comparison plot is wide, so making it smaller makes it look weird.

- Use the new ArviZ :)

As we also have some maintenance commits. I added a guide to build the docs locally (did we have that info somewhere?). I can push this in a different PR if needed. 

Feedback is welcome :)

tomicapretto:
Thanks for the contribution @juanitorduz !

juanitorduz:
> Thanks for the contribution @juanitorduz !

Thanks for the feedback! I have many more applied examples where we can showcase how Bambi shines :D

PR #971, opened 2026-02-17
Interpret refactor

GStechschulte:
This PR refactors the interpret submodule. The objective was to:

- Simplify the plotting control flow
- Simplify how comparisons and slopes are computed
- Standardize how summary data frames are built
- Bridge functionality between interpret and marginaleffects

This is done by adopting a more functional design:

- Structural pattern matching is used to de-structure user-input and validate the conditional, contrast, and wrt parameters for the predictions, comparisons, and slopes functions.
- The core data structure used is a Pandas Series, which is used to declaratively build data grids (with consistent type casting) and summary data frames (via Pandas split/apply/combine).
- Plotting uses the Seaborn objects API to declaratively define the plotting specification, with a PlotConfig type to configure main, group, and panel figures as well as figure-level customization.

Notes:
- predictions/comparisons/slopes now return a `Result` type with a summary dataframe and posterior samples fields.
- comparisons now accepts a function callable so users can compute "any" quantity of interest in addition to the default `diff`, `ratio` and `lift` comparisons.
- Previously exposed helper functions like `create_data_grid` and `select_draws` are no longer exposed since we now return the posterior samples.
- Defaults logger has been removed for now to decrease the complexity of the codebase.
- All notebooks using any interpret function has been updated.

This refactor is a breaking change. Although the API remains mostly the same, the types returned by the functions (`Result` for effects functions and `Plot` for plotting functions) are different and users will need to adapt.


bambi/interpret/types.py
```diff
@@ -0,0 +1,372 @@
...
+                )
+            case dict():
+                return ConditionalVariables(
+                    tuple(
+                        _resolve_values(name, data, values, _conditional_defaults)
+                        for name, values in conditional.items()
+                    )
+                )
+            case None:
+                return ConditionalVariables(())
```
tomicapretto:
```suggestion
                return ConditionalVariables(tuple())
```

A bit more explicit

bambi/interpret/plots.py
```diff
@@ -0,0 +1,237 @@
...
+        return PlottingConfig(
+            subplot=SubplotConfig.from_params(var_names, subplot_kwargs),
+            figure=FigureConfig.from_kwargs(fig_kwargs),
+        )
+
+
+def _add_main_layer(plot: Plot, data: DataFrame, config: PlottingConfig) -> Plot:
+    # Plot labels
+    ymin = next(filter(lambda col: "lower" in col, data.columns))
+    ymax = next(filter(lambda col: "upper" in col, data.columns))
```
tomicapretto:
```suggestion
    ymin = next(col for col in data.columns if "lower" in col)
    ymax = next(col for col in data.columns if "upper" in col)
```

no real preference, just in case you prefer this other option

bambi/interpret/effects.py
```diff
@@ -1,890 +1,923 @@
...
+
+from .ops import ComparisonFunc, SlopeFunc, get_comparison_func, get_slope_func
+from .plots import PlottingConfig, plot
+from .types import (
+    ComparisonVariable,
+    ConditionalVariables,
+    DefaultVariables,
+    Result,
+    SlopeVariable,
+)
```
tomicapretto: Is there a reason to mix absolute and relative imports? Unless there's a blocker, I would go for absolute imports.

GStechschulte:
<p>It can but I was avoiding it because some of the code also uses these functions ;)</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/971/discussion/#comment-3915924329'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

GStechschulte:
<p>Ehhh. Yeah, it may be quicker if you do?</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/971/discussion/#comment-3915924231'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

GStechschulte:
<p>I also really like that :)</p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/971/discussion/#comment-3915924050'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

bambi/interpret/effects.py
GStechschulte: Not sure why I did that (I think my LSP was giving me fits at the beginning). Fixed to use absolute imports.

bambi/interpret/plots.py
```diff
@@ -0,0 +1,237 @@
...
+                so.Range(),
+                so.Dodge(),
+                ymin=ymin,
+                ymax=ymax,
+            )
+        # Line plot if numeric dtype
+        case dtype if is_float_dtype(dtype):
+            plot = plot.add(so.Line())
+            plot = plot.add(
+                so.Band(alpha=0.3),
```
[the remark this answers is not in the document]
GStechschulte: I can just add it as an attribute to the `FigureConfig`? This would also make plotting multiple `prob` values easier since you could adjust the alpha level.

GStechschulte:
<p>Easy fix with adding <code>alpha</code> as a field to the <code>FigureConfig</code>   </p><p> </p>

---

 <a href='https://app.reviewnb.com/bambinos/bambi/pull/971/discussion/#comment-3915923852'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

GStechschulte:
Wow, thanks for the quick and thorough review @tomicapretto. Went through and resolved most of your comments :)

GStechschulte:
> View / edit / reply to [this conversation](https://app.reviewnb.com/bambinos/bambi/pull/971/discussion/#comment-3915924231) on ReviewNB
> 
> _tomicapretto commented on 2026-02-17T17:04:50Z_ ----------------------------------------------------------------
> 
> I can rerun this notebook and push it if it helps
> 
> 
> _GStechschulte commented on 2026-02-17T20:32:15Z_ ----------------------------------------------------------------
> 
> Ehhh. Yeah, it may be quicker if you do?

@tomicapretto this is really the only thing I requested.

tomicapretto:
@GStechschulte, when re-running the cherry-blossom example I get the following plot

image

this happens because `year` is an integer variable. If I cast it to float in the original dataframe, it looks as expected with the band. I'm debating if the plot should be left as it is, if I should manually cast year to float, or if there must be a different way to handle the case where 'x' is integer but you actually want to plot it as a continuous one.

tomicapretto:
@GStechschulte do you think it'll be difficult/complicated to allow for multiple `prob` values? Asking since it can be useful in a few plces, and having to create almost the same plot twice seems unnecessary

GStechschulte:
> this happens because `year` is an integer variable. If I cast it to float in the original dataframe, it looks as expected with the band. I'm debating if the plot should be left as it is, if I should manually cast year to float, or if there must be a different way to handle the case where 'x' is integer but you actually want to plot it as a continuous one.

We had logic before to determine whether to call the strip or line plot. It was something like "if the number of integer values is more than 10, then use a line plot, else strip plot".

I left it like this because I want the interpret code this time around to not have unexpected or surprising behavior, e.g. if you decide to encode a variable with contrasts that has 200 levels, then interpret will attempt to plot those 200 levels as categories. interpret won't attempt to "fix/understand" what you did as a modeler.

But yeah, this being an integer dtype, I can see the argue for a strip plot if the number of values is less than some arbitrary number...

tomicapretto:
> > this happens because `year` is an integer variable. If I cast it to float in the original dataframe, it looks as expected with the band. I'm debating if the plot should be left as it is, if I should manually cast year to float, or if there must be a different way to handle the case where 'x' is integer but you actually want to plot it as a continuous one.
> 
> We had logic before to determine whether to call the strip or line plot. It was something like "if the number of integer values is more than 10, then use a line plot, else strip plot".
> 
> I left it like this because I want the interpret code this time around to not have unexpected or surprising behavior, e.g. if you decide to encode a variable with contrasts that has 200 levels, then interpret will attempt to plot those 200 levels as categories. interpret won't attempt to "fix/understand" what you did as a modeler.
> 
> But yeah, this being an integer dtype, I can see the argue for a strip plot if the number of values is less than some arbitrary number...

I agree, let's leave it as it is and educate ourselves and users (via examples) that if they want bands they should use floats.

I think the supporting logic is: when the predictor is integer we don't perform any interpolation, thus you get a strip plot. If you do want the interpolation, manually cast your values to float.

In the future, we can revisit this choice. For this PR, I don't want us to get stuck because of this (not trivial to resolve) nuance.

tomicapretto:
@GStechschulte don't worry about the docs problem, I'm looking into it

tomicapretto:
@GStechschulte, Griffe, which is used by quartodoc, got a new release with breaking changes. That broke our docs CI. I added an upper pin for now. https://mkdocstrings.github.io/griffe/changelog/#200-2026-02-09

tomicapretto:
@GStechschulte in the pigs example I get the following plot:

```python
time = np.array(range(1, 13))

p = bmb.interpret.plot_predictions(
    model=bmb_model,
    idata=bmb_idata,
    prob=0.95,
    conditional={"Time": time, "Pig": random_pigs},
    fig_kwargs={
        "theme": {
            "figure.figsize": (10, 5)
        },
        "xlabel": "Time (week)",
        "ylabel": "Weight (kg)",
        "title": "Pig {}".format,
        "wrap": 5,
    },
    subplot_kwargs={"main": "Time", "panel": "Pig"}
)
p.add(so.Dot(), data=df[df["Pig"].isin(random_pigs)], x="Time", y="Weight", group="Pig").show()
```

image

The problem is that only one or two dots appear per pig. 

If I remove the panel variable, this goes away:

```python
time = np.array(range(1, 13))

p = bmb.interpret.plot_predictions(
    model=bmb_model,
    idata=bmb_idata,
    prob=0.95,
    conditional={"Time": time, "Pig": random_pigs},
    fig_kwargs={
        "theme": {
            "figure.figsize": (10, 5)
        },
        "xlabel": "Time (week)",
        "ylabel": "Weight (kg)",
    },
    subplot_kwargs={"main": "Time"},
)
p.add(so.Dot(), data=df[df["Pig"].isin(random_pigs)], x="Time", y="Weight").show()
```

image

---

I'm pushing the notebook following what I think is the correct code for the plot

GStechschulte:
@tomicapretto it looks like this is an indexing issue when you add a "new" data source to an existing `Plot`. If you reset the index you get the following

```python
time = np.array(range(1, 13))

p = bmb.interpret.plot_predictions(
    model=bmb_model,
    idata=bmb_idata,
    prob=0.95,
    conditional={"Time": time, "Pig": random_pigs},
    fig_kwargs={
        "theme": {
            "figure.figsize": (10, 5)
        },
        "xlabel": "Time (week)",
        "ylabel": "Weight (kg)",
        "title": "Pig {}".format,
        "wrap": 5,
    },
    subplot_kwargs={"main": "Time", "panel": "Pig"},
)
p.add(
    so.Dot(), 
    data=df[df["Pig"].isin(random_pigs)].reset_index(drop=True), 
    x="Time", 
    y="Weight").show()
```
image

tomicapretto:
@GStechschulte, thanks for the fix!

But I still think there's something wrong. Look at the previous plot we had and the panel for Pig 8269

image

I think the model summary is correct, but the data is wrong. What do you think?

An alternative could be to create a matplotlib figure, draw the seaborn plot in there, and then add the observed data using matplotlib.

GStechschulte:
The problem is when adding a new layer via `.add()` to a faceted Seaborn objects plot, the new data doesn't automatically inherit the faceting assignment. We need to explicitly tell it the facet.

```python
time = np.array(range(1, 13))

p = bmb.interpret.plot_predictions(
    model=bmb_model,
    idata=bmb_idata,
    prob=0.95,
    conditional={"Time": time, "Pig": random_pigs},
    fig_kwargs={
        "theme": {
            "figure.figsize": (10, 5)
        },
        "xlabel": "Time (week)",
        "ylabel": "Weight (kg)",
        "title": "Pig {}".format,
        "wrap": 5,
    },
    subplot_kwargs={"main": "Time", "panel": "Pig"},
)
p.add(
    so.Dot(),
    data=df[df["Pig"].isin(random_pigs)].reset_index(drop=True), 
    x="Time", 
    y="Weight",
    col="Pig" # Explicit facet assignment
    ).show()
```
image

tomicapretto:
There it goes, thanks a lot!! @GStechschulte

Issue #941, opened 2025-10-30
Add `ZeroSumNormal` support


AlexAndorra:
Currently, Bambi uses reference-encoding by default when categorical predictors and / or a Categorical/Multinomial likelihood is present in the model. It'd be great to add the choice to use a zero-sum constraint, instead of a reference-encoding one, since many modeling cases involve categories without clear placebo/reference.

It shouldn't be too hard (famous last words), since PyMC already has [ZeroSumNormal](https://www.pymc.io/projects/docs/en/stable/api/distributions/generated/pymc.ZeroSumNormal.html).

cc @jessegrabowski @canyon289

tomicapretto:
There's a hack to do it, which is to build the indicator columns yourself and then pass them to the formula and use a ZeroSumNormal prior for that term. But it's ugly and very confusing.

The problem I see with a straightforward implementation is that we would need to somehow intercept the creation of the design matrix to remove the factor with the ZeroSumNormal prior. I think it would be better to modify formulae so it supports something like `ensure_full_rank` in here https://matthewwardrop.github.io/formulaic/latest/guides/model_specs/

Chirag3841:
@AlexAndorra I guess here we can add a sum encoding option: use Formulaic sum-to-zero contrasts, then assign pm.ZeroSumNormal to the categorical coefficients instead of reference (treatment) dummies.

AlexAndorra:
What would the API look like for the user @Chirag3841 ?

Chirag3841:
@AlexAndorra model = bmb.Model(
    "y ~ x + cat",
    data,
    categorical_encoding="sum"
)
Like that 
where we are assigning categorical_encoding in __init__ part of model.py and we can add zero-sum encoding for categorical predictors in else part.

AlexAndorra:
That's be great @Chirag3841 ! Something like `categorical_encoding="zero-sum"` and `categorical_encoding="reference"` would be even clearer, if possible

Chirag3841:
@AlexAndorra I would like to solve this issue too if anyone had not taken it yet.

AlexAndorra:
That'd be great @Chirag3841 ! I think it's up for grabs (cc @tomicapretto )

PR #973, opened 2026-02-18
Add zero-sum encoding option for categorical predictors

Chirag3841:
**Thank your for opening a PR!**

Before you proceed, please check the following notes.

* Make sure you describe the changes introduced in the PR.
* If this PR addresses an existing issue, please mention it here.
* Check our [Guidelines for Contributing](https://github.com/bambinos/bambi/blob/main/CONTRIBUTING.md).
In particular, pay attention to:
    * [Contributing steps](https://github.com/bambinos/bambi/blob/main/CONTRIBUTING.md#steps).
    * [Pull request checklist](https://github.com/bambinos/bambi/blob/main/CONTRIBUTING.md#pull-request-checklist).

Issue #941 
Description

This PR adds support for an alternative categorical encoding strategy using a zero-sum constraint via pm.ZeroSumNormal.
Users can now choose between "reference" (default) and "zero-sum" encoding through the new categorical_encoding argument in bmb.Model.

Changes done 
Added categorical_encoding argument to bmb.Model
Implemented "zero-sum" encoding for categorical predictors in PyMC backend using pm.ZeroSumNormal
Added unit test to verify model builds correctly with categorical_encoding="zero-sum"

### DIFF
--- a/bambi/backend/terms.py
+++ b/bambi/backend/terms.py
@@ -68,6 +68,27 @@ class CommonTerm:
                 if value.ndim == 1:
                     kwargs[key] = np.hstack([value[:, np.newaxis]] * response_dims_n)
 
+        if self.term.categorical and spec.categorical_encoding == "zero-sum":
+            kwargs = dict(self.term.prior.args)
+
+            # ZeroSumNormal does NOT accept mu
+            kwargs.pop("mu", None)
+
+            # ZeroSumNormal requires sigma to be scalar (not vector)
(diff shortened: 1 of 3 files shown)


tomicapretto:
@Chirag3841, thanks for the initiative to start this contribution. I'm afraid supporting zero-sum encoding in Bambi is a way more complex endeavor than what this PR attempts to do. For instance, we'd need to modify the model-building library, formulae, to return non-full rank matrices in cases we want to use ZeroSum normal. And then we'd need to modify multiple parts of Bambi to account for that change. The addition of a model-level argument is not a suitable solution, as the encoding of each categorical covariate is controlled at the covariate level.

PR #977, opened 2026-03-09
Remove pps and update target params in interpret effects signatures

GStechschulte:
@tomicapretto and I had a conversation in Discord about removing the `pps` parameter and only "using" the `target` parameter. 

The default of `target=mean` remains unchanged. With `pps` removed, the user is now required to pass `target=response` where `response` is the name of the response variable in their Bambi formula, e.g., `mpg`.

I am okay with this API change since the interpret refactor was already a breaking change.


bambi/interpret/effects.py
```diff
@@ -576,9 +576,10 @@ def comparisons(
     average_by : str, list or None
         Variables to average comparisons over.
     target : str
-        The target parameter to compare. Default is "mean".
-    pps : bool
-        Whether to use posterior predictive samples. Default is False.
+        Which quantity to extract. `mean` (default) for the posterior of the parent
+        parameter (e.g. `mu`). Pass the response variable name (e.g. `mpg`) for posterior
+        predictive samples. Pass a distributional component name (e.g. `sigma`) for the
+        posterior of that component.
```
tomicapretto:
```suggestion
        Which quantity to extract. `"mean"` (default) for the posterior of the parent
        parameter (e.g. `"mu"`). Pass the response variable name (e.g. `"mpg"`) for posterior
        predictive samples. Pass a distributional component name (e.g. `"sigma"`) for the
        posterior of that component.
```

tomicapretto (approved):
Just a few nits that I suggested. Otherwise it looks good!

PR #984, opened 2026-05-13
Use pymc 6

aloctavodia:


tomicapretto:
Let's get this merged and Bambi released. 
But, what about nutpie? It still uses main. 
Also, we should make it trigger tests when we change pyproject.toml

aloctavodia:
Updated nutpie, but now pixie is complaining about that...

aloctavodia:
Ohhh nutpie is not available yet on conda-forge.

danieltomasz:
@aloctavodia the conda recipe for nutpie still allows Python 3.11 via the selector, so the nutpie-feedstock CI tests are failing https://github.com/conda-forge/nutpie-feedstock/pull/92

maresb:
I'm on it...

tomicapretto (approved):

PR #988, opened 2026-06-16
Warn when nuts_sampler_kwargs is used with the default PyMC sampler

nickeubank:
## TL;DR

While working with bambi 0.17, I discovered that `nuts_sampler_kwargs` are silently ignored when using the default sampler. Looks like dev requires pymc 6, where this is (kind of) fixed — there's a FutureWarning, but it isn't bambi specific. So this adds a warning.  

The reference to PyMC < 6 can be dropped if you'd like since in *theory* by the time this gets merged users won't be allowed PyMC < 6, but up to y'all.

## Summary

Routing NUTS settings (most commonly `target_accept`) through `nuts_sampler_kwargs` is a quiet footgun with the **default PyMC sampler**:

- **PyMC < 6**: `nuts_sampler_kwargs` is *silently ignored* for the `"pymc"` sampler, so e.g. `model.fit(nuts_sampler_kwargs={"target_accept": 0.99})` runs at the default `target_accept=0.8`. Users chasing divergences can crank this value indefinitely with no effect and no feedback.
- **PyMC >= 6**: it is honored but emits a `FutureWarning` telling you to "pass NUTS keyword arguments via the `nuts={...}` argument to `pm.sample`" — guidance that's leaky in a Bambi context, since Bambi users never call `pm.sample` directly.

Because Bambi just forwards `nuts_sampler_kwargs` into `pm.sample`, neither situation is surfaced in Bambi's own terms.

This PR makes `Model.fit` emit a clear, version-independent warning when `nuts_sampler_kwargs` is passed while using the default PyMC sampler, steering users to the path that works on every PyMC version:

```python
model.fit(target_accept=0.9)   # instead of nuts_sampler_kwargs={"target_accept": 0.9}
```

The warning only fires when `sampler_backend == "pymc"` and a non-empty `nuts_sampler_kwargs` is supplied; external samplers (`nutpie`, `numpyro`, `blackjax`), for which `nuts_sampler_kwargs` is the correct channel, are unaffected.

## Changes

- `bambi/backend/pymc.py`: guard in `_run_mcmc` that warns for the default PyMC sampler.
- `tests/test_inference_methods.py`: tests that the warning fires for non-empty `nuts_sampler_kwargs`, that passing `target_accept` directly does **not** warn, and that an empty `nuts_sampler_kwargs={}` is a no-op.
- `CHANGELOG.md`: entry under *Unreleased*.

## Testing

`pytest tests/test_inference_methods.py` → 8 passed, 5 skipped (JAX/nutpie not installed). `black --check` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ## Maintenance and fixes
 
 * Non-Normal priors with random hyperpriors now build under the centered parameterization.
+* `Model.fit` now warns when `nuts_sampler_kwargs` is passed while using the default PyMC sampler, steering users to pass NUTS settings such as `target_accept` directly. Routing them through `nuts_sampler_kwargs` is silently ignored on PyMC < 6 and deprecated on PyMC >= 6.
 
 <a id="0.18.0"></a>
 # [Bambi 0.18.0](https://github.com/bambinos/bambi/releases/tag/0.18.0) - 2026-05-19
--- a/bambi/backend/pymc.py
+++ b/bambi/backend/pymc.py
@@ -206,6 +206,22 @@ class PyMCModel:
         sampler_backend,
         **kwargs,
     ):
+        # Routing NUTS settings (most commonly `target_accept`) through
+        # `nuts_sampler_kwargs` is a footgun with the default PyMC sampler: PyMC < 6
+        # silently ignores it, so the setting never takes effect, while PyMC >= 6
+        # deprecates it in favor of `pm.sample(nuts=...)`. Bambi users don't call
+        # `pm.sample` directly, so steer them to the supported, version-independent
+        # path: pass these settings straight to `Model.fit()`.
+        if sampler_backend == "pymc" and kwargs.get("nuts_sampler_kwargs"):
+            warnings.warn(
+                "Passing NUTS settings through `nuts_sampler_kwargs` is not recommended with the "
+                "default PyMC sampler (PyMC < 6 silently ignores it; PyMC >= 6 deprecates it). "
+                "Pass settings such as `target_accept` directly to `Model.fit()` instead, e.g. "
+                "`model.fit(target_accept=0.9)`.",
+                UserWarning,
+                stacklevel=2,
+            )
+
         # Don't include the parameters of the likelihood, which are deterministics.
         # They can take lot of space in the trace and increase RAM requirements.
         vars_to_sample = get_default_varnames(
--- a/tests/test_inference_methods.py
+++ b/tests/test_inference_methods.py
@@ -1,3 +1,5 @@
+import warnings
+
 import bambi as bmb
 import numpy as np
 import pandas as pd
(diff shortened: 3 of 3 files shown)


GStechschulte:
Hey @nickeubank thanks for opening the PR. Silently ignoring `nuts_sampler_kwargs` when using the default `pymc` sampler is definitely problematic.

Raising the `UserWarning` is better than nothing, but I think we could do better. Maybe Bambi could replace the warning with a remapping of kwargs. For example

```python
if sampler_backend == "pymc":
	nuts_sampler_kwargs = kwargs.pop("nuts_sampler_kwargs", None)
	if nuts_sampler_kwargs:
		nuts = kwargs.get("nuts", {})
		nuts.update(nuts_sampler_kwargs)
		kwargs["nuts"] = nuts
```

ensuring that sample kwargs are always passed correctly.

@tomicapretto in the future, maybe we should think about a better function signature so we don't have to do this workaround.

nickeubank:
Thanks @GStechschulte! Certainly a more elegant solution, assuming PyMC keeps the syntax stable. My one concern would be that if PyMC changes, the silent redirection of key-value-pairs from `nuts_sampler_kwargs` to `nuts` might lead to some rather confusing errors for users. Raising the error that users shouldn't use `nuts_sampler_kwargs` anymore has the advantage that if something goes wrong, the user would likely have an easier time finding the appropriate documentation to correct the error.

Is my interpretation of the project direction correct that the next release will require PyMC >= 6 so the most worrying behavior will be ruled out (and we're down to worrying about the lack of an informative and appropriate error)?

tomicapretto:
> Hey @nickeubank thanks for opening the PR. Silently ignoring `nuts_sampler_kwargs` when using the default `pymc` sampler is definitely problematic.
> 
> Raising the `UserWarning` is better than nothing, but I think we could do better. Maybe Bambi could replace the warning with a remapping of kwargs. For example
> 
> ```python
> if sampler_backend == "pymc":
> 	nuts_sampler_kwargs = kwargs.pop("nuts_sampler_kwargs", None)
> 	if nuts_sampler_kwargs:
> 		nuts = kwargs.get("nuts", {})
> 		nuts.update(nuts_sampler_kwargs)
> 		kwargs["nuts"] = nuts
> ```
> 
> ensuring that sample kwargs are always passed correctly.
> 
> @tomicapretto in the future, maybe we should think about a better function signature so we don't have to do this workaround.

@GStechschulte I agree with your point. I think we should directly go for the better function signature.

GStechschulte:
I don't think we need to change the signature significantly. 

We keep common MCMC parameters (draws, tune, chains, cores, random_seed, etc) as named args. Then, for NUTS specific parameters we introduce the `nuts`  parameter similar to pymc `pm.sample(..., nuts={"target_accept": 0.9})` ensuring that PyMC handles the internal routing of NUTS parameters to their appropriate samplers (blackjax, numpyro, etc.). Finally, we still keep kwargs in case the user wants to pass other things such as `idata_kwargs`, `compile_kwargs `, etc.

@nickeubank I have a working implementation. Would you like me to push it to your branch or close this PR and open a new one? Again, thanks for raising this issue!

tomicapretto:
> We keep common MCMC parameters (draws, tune, chains, cores, random_seed, etc) as named args. Then, for NUTS specific parameters we introduce the nuts parameter similar to pymc pm.sample(..., nuts={"target_accept": 0.9}) ensuring that PyMC handles the internal routing of NUTS parameters to their appropriate samplers (blackjax, numpyro, etc.). Finally, we still keep kwargs in case the user wants to pass other things such as idata_kwargs, compile_kwargs , etc.

I really like this!

nickeubank:
Seems like closing this and openning a clean branch makes the most sense! Thanks @GStechschulte !

GStechschulte:
> Seems like closing this and openning a clean branch makes the most sense! Thanks @GStechschulte !

Thanks for opening the issue. We wouldn't have been able to do this without your help :)

PR #989, opened 2026-06-21
Remove private type aliases in interpret API

GStechschulte:
I noticed that we were using private type alias in the type hints of `plot_comparisons` and `plot_slopes`. I removed them and added "standard" type hints and improved the docstring.


bambi/interpret/effects.py
```diff
@@ -555,7 +553,8 @@ def comparisons(
     conditional: Optional[str | list[str] | dict[str, np.ndarray | list | int | float]] = None,
     average_by: str | list[str] | None = None,
     target: str = "mean",
-    comparison: ComparisonFunc | str = "diff",
+    pps: bool = False,
```
tomicapretto: I am not sure I understand why you are adding this here, maybe this explains the CI error?

GStechschulte: hah. I messed something up on a rebase. I will fix this.

GStechschulte:
@tomicapretto I needed to add a specific arviz dependency to the Pixi dependencies because of conda pinning it to 0.23.4 but PyMC 6.0.0 required >= 1.1.0. 

I also exported the `Result` type since the effects functions returns this type to the user.

tomicapretto:
@GStechschulte all good!

tomicapretto (approved):

PR #990, opened 2026-06-24
Fix PyMC NUTS sampler kwargs silent failure

GStechschulte:
Addresses the original PR #988

We keep common MCMC parameters (draws, tune, chains, cores, random_seed, etc) as named args. Then, for NUTS specific parameters we introduce the `nuts`  parameter similar to pymc `pm.sample(..., nuts={"target_accept": 0.9})` ensuring that PyMC handles the internal routing of NUTS parameters to their appropriate samplers (blackjax, numpyro, etc.). Finally, we still keep kwargs in case the user wants to pass other things such as `idata_kwargs`, `compile_kwargs `, etc.

A control flow block is also added to: (1) raise a `FutureWarning` to inform the user that the legacy `nuts_sampler_kwargs` is deprecated, and (2) to re-route the `nuts_sampler_kwargs` to the `nuts` parameter ensuring that don't break user's code.

### DIFF
--- a/tests/test_inference_methods.py
+++ b/tests/test_inference_methods.py
@@ -1,3 +1,5 @@
+import warnings
+
 import bambi as bmb
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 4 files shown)


tests/test_inference_methods.py
```diff
@@ -100,3 +100,61 @@ def test_legacy_nuts_numpyro_warning(data_random_n100):
 
     with pytest.warns(FutureWarning, match="'numpyro_nuts' has been replaced by 'numpyro'"):
         model.fit(inference_method="numpyro_nuts", draws=10, tune=10)
+
+
+def test_nuts_parameter_for_default_sampler(data_random_n100):
```
tomicapretto: Could you use `mock_pymc_sample` so the actual sampler does not run? Or is it important to run the sampler? 
You can check other tests that use `mock_pymc_sample`

tomicapretto (commented):
Just one suggestion for the tests, all good!

PR #992, opened 2026-06-30
Add compute_log_prior

### LINKED ISSUE #972: Add compute_log_prior
Similarly to `compute_log_likelihood`, we could have `compute_log_prior`, which is useful when using prior/likelihood sensitivity check see `psense` functions in arviz-plots and arviz-stats

aloctavodia:
Closes #972

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -9,6 +9,7 @@ from importlib.metadata import version
 import formulae as fm
 import pandas as pd
 import pymc as pm
+import xarray as xr
 from arviz_plots import plot_dist
 from arviz_stats import residual_r2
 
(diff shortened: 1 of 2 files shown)


bambi/models.py
```diff
@@ -1110,6 +1111,65 @@ def compute_log_likelihood(self, idata, data=None, inplace=True):
         else:
             return idata
 
+    def compute_log_prior(self, idata, inplace=True):
```
tomicapretto, line 1116: Looks good. Could you add a test?

tests/test_log_prior.py
```diff
@@ -0,0 +1,47 @@
+import numpy as np
+import pytest
+
+import bambi as bmb
+
+
+@pytest.mark.usefixtures("mock_pymc_sample")
+def test_inplace(data_beetle):
+    model = bmb.Model("prop(y, n) ~ x", data_beetle, family="binomial")
+    idata = model.fit(tune=20, draws=20, chains=1, random_seed=1234)
```
tomicapretto: Any reason to not use more chains, draws, and tune? `mock_pymc_sample` makes sampling cheap as it gets them from prior predictive.

aloctavodia: No reason. Updated

tomicapretto (approved):

PR #994, opened 2026-07-07
add _re_center_intercept

aloctavodia:
Not sure this is the best approach, or even if we want to add this given the upcoming changes in Bambi. But conditional on the current state of Bambi, a function like this will be helpful for ArviZ's moment_matching and for some plotting in Kulprit. The reason is that in both cases there is a mismatch between the intercept in ``idata.posterior`` and the PyMC value variable.

### DIFF
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -7,6 +7,7 @@ from copy import deepcopy
 from importlib.metadata import version
 
 import formulae as fm
+import numpy as np
 import pandas as pd
 import pymc as pm
 import xarray as xr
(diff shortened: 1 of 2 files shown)


tomicapretto:
I think it's good to add it if it helps now. Could you add a test? Considering the following cases:

1. A regression model with a 1d outcome (could be gaussian, poisson, etc.) and numerical predictors.
2. Like above, but including categoric predictors plus interaction between categorics and numericals
3. A model with a categorical family, or multinomial. Numerical predictors.
4. Like above, but also with categoric predictors.

The reason why I mention those is the linear predictor could be constructed in different ways and result in different shapes, and I would like to check this addition covers all the cases. You don't need to invent any new model in tests, you can reuse models that already exist, just test the `_re_center_intercept`.

tomicapretto (approved):

PR #997, opened 2026-07-09
Add ExGaussian family (closes #995)

### LINKED ISSUE #995: Add ExGaussian family

yogendrarau:
This PR adds the ExGaussian family, resolving #995.

The ExGaussian distribution (Gaussian convolved with an exponential
component) is commonly used to model reaction time data. It uses PyMC's
`pm.ExGaussian(mu, sigma, nu)` parametrization directly, so no
reparametrization in `transform_kwargs` is needed.

Design notes:
- Links: `mu` gets identity/log/inverse (matching Gaussian); `sigma` and
  `nu` get `log` since both must be positive.
- Default priors: `HalfNormal` for both `sigma` and `nu`. Note that unlike
  StudentT, whose `nu` is degrees of freedom and uses a `Gamma` prior,
  ExGaussian's `nu` is the scale of the exponential component, so
  `HalfNormal` is consistent with how `sigma` is handled.

Tests: added a fit/predict test plus entries in the family/link
construction tests. Verified locally that the model builds with correct
automatic priors, NUTS sampling recovers the true parameters on synthetic
data, posterior predictive works, and invalid links are rejected.

I followed the existing built-in family pattern (e.g. VonMises, Wald) and
didn't add a separate example notebook, since those families don't have
dedicated ones either but I'd be happy to add one if you'd prefer.

### DIFF
--- a/bambi/defaults/families.py
+++ b/bambi/defaults/families.py
@@ -7,6 +7,7 @@ from bambi.families.univariate import (
     Binomial,
     Categorical,
     Cumulative,
+    ExGaussian,
     Exponential,
     Gamma,
     Gaussian,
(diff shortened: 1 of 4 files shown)


tomicapretto:
Thank you, @yogendrarau!

PR #1001, opened 2026-07-31
Raise an error for prior names that match no term

### LINKED ISSUE #815: Raise an error when there are invalid prior specifications
Is there a way for `bambi` to throw an error if there are entries in the `priors` dict passed to `bmb.Model` that are not assigned to any term?

AlexanderFengler:
Closes #815.

A `priors` entry whose key matches no term was silently dropped, so the model was fit with Bambi's automatic prior while the user believed theirs was used.

Adds `bmb.config["UNUSED_PRIORS"]`, per your suggestion on the issue. It defaults to `"warn"` so existing code keeps working; `"error"` and `"ignore"` are also available, and the default can move to `"error"` later if that seems right.

### The two entry points disagreed

Not in the issue, and the reason the first version of this PR had an awkward `from_model` flag. `Model()` and `set_priors()` consumed the same dictionary shape by different rules:

| Input | `bmb.Model(..., priors=…)` | `model.set_priors(priors=…)` |
|---|---|---|
| `{"TYPO": p}` | silently ignored | `KeyError: 'TYPO'` |
| `{"common": p}` | accepted | `KeyError: 'common'` |
| `{"x": p}`, >1 distributional component | applied to the parent's `x` | **silently dropped** |

The last row is a bug in its own right. Rather than teach the validator about both rule sets, `_set_priors` now dispatches exactly as `Model.__init__` does:

- the parent takes the whole dictionary unless it is named explicitly
- every other component is addressed by name
- `common` / `group_specific` work as keys as well as arguments, and the arguments win when both are given

So a priors dictionary means the same thing wherever it is passed, validation needs a single rule, and the boolean is gone.

`update_priors` now skips names it doesn't have instead of raising `KeyError`, since `_check_prior_names` owns the reporting and `"warn"`/`"ignore"` must not crash afterwards. `test_set_prior_unexisting_term` is updated accordingly — it asserted the old bare `KeyError`.

### Two silent cases found while writing this

Both are the same class as the reported bug, both now caught:

- naming the parent routes its terms into that nested dictionary, so a bare term name left at the top level was never read: `{"mu": {...}, "Intercept": p}` gave `Intercept` the automatic prior
- `set_priors({"mu": {...}})` on a single-component model raised a bare `KeyError`

### Also here: a `numba` cap

Unrelated to #815, but CI could not run without it. `numba` 0.66.0 exceeds `pytensor`'s ceiling of `<=0.65.1`, so the solve failed before any test ran — this has been breaking `main` since 2026-07-17, not just this branch. Capped at `<0.66` in the two places `numba` is declared. Happy to split it out if you'd rather it were its own PR.

### DIFF
--- a/bambi/config.py
+++ b/bambi/config.py
@@ -5,7 +5,11 @@ class Config:
     When a user tries to set a configuration variable to a non-supported value, it raises an error.
     """
 
-    __FIELDS = {"INTERPRET_VERBOSE": (True, False), "SPARSE_DOT": (False, True)}
+    __FIELDS = {
+        "INTERPRET_VERBOSE": (True, False),
+        "SPARSE_DOT": (False, True),
+        "UNUSED_PRIORS": ("warn", "error", "ignore"),
+    }
 
     def __init__(self, config_dict: dict = None):
         config_dict = {} if config_dict is None else config_dict
--- a/bambi/models.py
+++ b/bambi/models.py
@@ -17,6 +17,7 @@ from pymc.backends.arviz import apply_function_over_dataset, coords_and_dims_for
 from pymc.model.transform.conditioning import remove_value_transforms
 
 from bambi.backend import PyMCModel
+from bambi.config import config
 from bambi.defaults import get_builtin_family
 from bambi.families import Family, univariate
 from bambi.formula import Formula, check_ordinal_formula
@@ -63,9 +64,15 @@ class Model:
         `"poisson"`, `"t"`, and `"wald"`. Defaults to `"gaussian"`.
     priors : dict, optional
         Optional specification of priors for one or more terms. A dictionary where the keys are
-        the names of terms in the model, "common," or "group_specific" and the values are
-        instances of class `Prior`. If priors are unset, use automatic priors inspired by
-        the R rstanarm library.
+        the names of terms in the model, "common," "group_specific," or the name of a model
+        component, and the values are instances of class `Prior`. A distributional component
+        name (e.g. "sigma" when it is modeled with a formula) maps to a nested dictionary of
+        the same form; a constant component name maps directly to a `Prior`, a number, or an
+        array. If priors are unset, use automatic priors inspired by the R rstanarm library.
+        Names that don't match any term or component are reported with a warning; set
+        `bmb.config["UNUSED_PRIORS"]` to `"error"` or `"ignore"` to change that.
+        Bare term priors can be combined with priors nested under the parent component. If both
+        specify the same term, the nested parent prior takes precedence.
     link : str or dict of str to str, optional
         The name of the link function to use. Valid names are `"cloglog"`, `"identity"`,
         `"inverse_squared"`, `"inverse"`, `"log"`, `"logit"`, `"probit"`, and
@@ -192,10 +199,11 @@ class Model:
                 "Please specify an outcome variable using the formula interface."
             )
 
-        if self.family.likelihood.parent in priors:
-            parent_priors = priors[self.family.likelihood.parent]
-        else:
-            parent_priors = priors
+        # Merge bare term priors with nested parent priors; nested entries take precedence.
+        parent_name = self.family.likelihood.parent
+        parent_priors = {name: prior for name, prior in priors.items() if name != parent_name}
+        if parent_name in priors:
+            parent_priors.update(priors[parent_name])
 
         # Add response component
         self.response_component = ResponseComponent(design.response, self)
@@ -242,6 +250,9 @@ class Model:
         for name in auxiliary_parameters:
             component_prior = priors.get(name, None)
             self.components[name] = ConstantComponent(name, component_prior, self)
+
+        # Validate prior names, now that every component and its terms are known.
+        self._check_prior_names(priors)
 
         # Validate per-component noncentered dict, now that all components are known.
         if isinstance(self.noncentered, dict):
@@ -421,13 +432,20 @@ class Model:
         Parameters
         ----------
         priors : dict or None, optional
-            Dictionary of priors to update. Keys are names of terms to update; values are the new
-            priors (either a `Prior` instance, or an int or float that scales the default priors).
+            Dictionary of priors to update. Accepts the same specification as the `priors`
+            argument of `Model`, and entries here take precedence over the `common` and
+            `group_specific` arguments. Names that don't match any term or component are
+            reported with a warning; set `bmb.config["UNUSED_PRIORS"]` to `"error"` or
+            `"ignore"` to change that.
         common : Prior, int, float or None, optional
             A prior specification to apply to all common terms included in the model.
         group_specific : Prior, int, float or None, optional
             A prior specification to apply to all group specific terms included in the model.
         """
+        if priors is not None:
+            # Validate before touching any state, so a rejected call leaves the model as it was.
+            self._check_prior_names(priors)
+
         kwargs = dict(zip(["priors", "common", "group_specific"], [priors, common, group_specific]))
         self._added_priors.update(kwargs)
         self._build_priors()  # After updating, we need to rebuild priors.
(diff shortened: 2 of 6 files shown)


bambi/config.py
```diff
@@ -5,7 +5,13 @@ class Config:
     When a user tries to set a configuration variable to a non-supported value, it raises an error.
     """
 
-    __FIELDS = {"INTERPRET_VERBOSE": (True, False), "SPARSE_DOT": (False, True)}
+    __FIELDS = {
+        "INTERPRET_VERBOSE": (True, False),
+        "SPARSE_DOT": (False, True),
+        # What to do when a `priors` dictionary has keys matching no term or component.
+        # Erroring is the default because an unused prior silently changes the fitted model.
```
tomicapretto: Let's remove this comment, the name and entries are quite self explanatory.
Also, I think the first option should be "warning" to not make working code fail unexpectedly. Later we can change this from warning to error if it make sense.

bambi/models.py
```diff
@@ -65,7 +66,10 @@ class Model:
         Optional specification of priors for one or more terms. A dictionary where the keys are
         the names of terms in the model, "common," or "group_specific" and the values are
         instances of class `Prior`. If priors are unset, use automatic priors inspired by
-        the R rstanarm library.
+        the R rstanarm library. Names that don't match any term or component raise a
```
tomicapretto: Adapt to reflect a warning is emited.

bambi/models.py
```diff
@@ -153,6 +157,12 @@ def __init__(
         # Handle priors
         priors = {} if priors is None else deepcopy(priors)
 
+        # Record what the user asked for. Components below `.pop()` names off these dictionaries
```
tomicapretto: The users "asks for" many things in the initialization. I'm not sure this comment is adequate. I would say something along the lines of "record names of terms with manual priors for later verification"

tomicapretto: The users "asks for" many things in the initialization. I'm not sure this comment is adequate. I would say something along the lines of "record names of terms with manual priors for later verification"

bambi/models.py
```diff
@@ -460,6 +487,67 @@ def _build_priors(self):
...
             self.scaler.scale()
 
+    def _check_prior_names(self, given_keys, from_model=True):
+        """Report prior names that don't match any term or component.
+
+        Terms pop their own entry out of the priors dictionary and the leftovers are never read
+        again, so an unmatched name is silent: the model is fit with Bambi's automatic prior
+        while the user believes theirs was used. `given_keys` maps each supplied name to the set
+        of nested names under it, or `None` when the value wasn't a dictionary, and must be
+        captured before the components consume the priors.
```
tomicapretto:
```suggestion

```

tomicapretto: I think this docstring talks about things that happen outside of this method. I'm not against explaining things, but I am trying to not make the codebase longer and redundant.

tomicapretto (commented):
I still have to review the body of `_check_prior_names`, I couldn't follow it in this quick read. But leaving some other comments for now.

tomicapretto:
@AlexanderFengler thanks for the PR, but let's not include messages from CC saying "all tests pass" and "lint is good" since we already have CI for that.

I still need to double check how the method that checks names work. I have seen LLMs make things work by introducing an additional logical argument, such as `from_model`, in cases where the root of the problem was different. Not against the boolean parameter _per se_, but I'd like to see if there's something else that needs to be modified to make this feel less hacky.

AlexanderFengler:
@tomicapretto tried to address everything inc. extra cleanup pass.
Can benefit from one round of targeted feedback if anything remains sub-optimal, will address that tomorrow to see it through.

Note that tests were failing because of numba version upper bound induced by pytensor.
You may prefer to make that a separate small PR, fix currently included here.

bambi/models.py
```diff
@@ -63,9 +64,11 @@ class Model:
         `"poisson"`, `"t"`, and `"wald"`. Defaults to `"gaussian"`.
     priors : dict, optional
         Optional specification of priors for one or more terms. A dictionary where the keys are
-        the names of terms in the model, "common," or "group_specific" and the values are
-        instances of class `Prior`. If priors are unset, use automatic priors inspired by
-        the R rstanarm library.
+        the names of terms in the model, "common," "group_specific," or the name of a model
+        component (e.g. "sigma") mapped to a nested dictionary of the same form, and the values
```
tomicapretto: If the component is constant (does not depend on any covariate), the value will not be a nested dict. It will be a Prior, a number, or even an array (when the parameter is  at least a vector).

tests/test_priors.py
```diff
@@ -217,9 +217,11 @@ def test_set_priors(data_random_n100):
 
 
 def test_set_prior_unexisting_term(data_random_n100):
+    # This used to surface as a bare `KeyError: 'z'` from indexing the terms dictionary. It is
+    # now reported by `UNUSED_PRIORS`, which warns by default and names the offending key.
     prior = bmb.Prior("Uniform", lower=0, upper=50)
     model = bmb.Model("continuous1 ~ continuous2", data_random_n100)
-    with pytest.raises(KeyError):
+    with pytest.warns(UserWarning, match=r"Unused name\(s\) in `priors`: \['z'\]"):
```
tomicapretto: Do we want to keep this test now the behavior changes and is configurable? This is covered by `test_unused_prior_in_model_warns`

tests/test_priors.py
```diff
@@ -340,3 +342,149 @@ def CustomPrior(name, *args, dims=None, **kwargs):
...
+        bmb.Model("continuous1 ~ continuous2", data_random_n100, priors={"TYPO": prior})
+
+
+def test_unused_prior_error_mode(data_random_n100, monkeypatch):
+    monkeypatch.setattr(bmb.config, "UNUSED_PRIORS", "error")
+    prior = bmb.Prior("Normal", mu=0, sigma=1)
+    with pytest.raises(ValueError, match=r"Unused name\(s\) in `priors`: \['TYPO'\]") as info:
+        bmb.Model("continuous1 ~ continuous2", data_random_n100, priors={"TYPO": prior})
+
+    # The message has to be actionable: say what is valid, and how to opt out.
```
tomicapretto:
```suggestion
```

tests/test_priors.py
```diff
@@ -340,3 +342,149 @@ def CustomPrior(name, *args, dims=None, **kwargs):
...
+    assert "Intercept" in message
+    assert "UNUSED_PRIORS" in message
+
+
+def test_unused_prior_ignore_mode(data_random_n100, monkeypatch):
+    monkeypatch.setattr(bmb.config, "UNUSED_PRIORS", "ignore")
+    prior = bmb.Prior("Normal", mu=0, sigma=1)
+    model = bmb.Model("continuous1 ~ continuous2", data_random_n100, priors={"TYPO": prior})
+    model.build()
+    assert "TYPO" not in model.components["mu"].terms
```
tomicapretto: Is there a way to test no warning/error is emitted? That is what we need (actually, the second one is already implicitly checked).

tomicapretto: The assert is not doing anything relevant

tomicapretto (changes requested):

bambi/models.py
AlexanderFengler: Reworded — the docstring now distinguishes distributional components (nested dict of the same form) from constant components (a `Prior`, number, or array directly). The validator already treated both correctly, so this is docs-only.

tests/test_priors.py
AlexanderFengler: Dropped. One note for the record: it was the only test going through `set_priors()` — the call site that used to raise the bare `KeyError` — while `test_unused_prior_in_model_warns` covers the constructor. Both entry points share `_check_prior_names` today, so it is indeed redundant now; if the two paths ever diverge, this is the test to resurrect.

tests/test_priors.py
AlexanderFengler: Removed.

tests/test_priors.py
AlexanderFengler: Rewrote it to assert the actual contract: construct with a typo'd name under `ignore` and assert no "Unused name" warning was recorded — via the `recwarn` fixture filtered to our message, so unrelated warnings can't flake it. No-error stays implicit in the construction succeeding, as you noted.

AlexanderFengler: @tomicapretto `recwarn` seems to allow what you asked. 
I actually wasn't aware of it myself before.

tests/test_priors.py
AlexanderFengler: @tomicapretto note that I went against claude recommendation here, hence the note above.

tomicapretto:
@AlexanderFengler I have just force-pushed so we can keep two commits separating what happened here

PR #1002, opened 2026-08-15
Bambi rewrite

tomicapretto:
I know this is a huge PR, so I took some time to write notes to explain the most important changes below. 

# Summary

## Table of contents

- [Summary](#summary)
- [New features](#new-features)
  - [Log-logistic family](#log-logistic-family)
  - [Competing risks models](#competing-risks-models)
  - [Predictions in censored and truncated models](#predictions-in-censored-and-truncated-models)
  - [Predictions for new groups or groups of unknown identity](#predictions-for-new-groups-or-groups-of-unknown-identity)
  - [Count responses with explicit totals](#count-responses-with-explicit-totals)
  - [Estimation and visualization with `interpret`](#estimation-and-visualization-with-interpret)
  - [Automatic sampler selection](#automatic-sampler-selection)
- [Changes to existing functionality](#changes-to-existing-functionality)
  - [Response variable names](#response-variable-names)
  - [Prediction organization](#prediction-organization)
  - [Progress bar for prediction and log-likelihood](#progress-bar-for-prediction-and-log-likelihood)
  - [Model parameters and components](#model-parameters-and-components)
  - [Family definitions](#family-definitions)
  - [Custom links](#custom-links)
- [Dependency, packaging, and CI changes](#dependency-packaging-and-ci-changes)
- [Significant internal changes](#significant-internal-changes)
  - [Frontend](#frontend)
  - [Backend](#backend)
- [Documentation](#documentation)
- [Compatibility and release plan](#compatibility-and-release-plan)

This PR rewrites important parts of Bambi to make it easier and safer to extend. We aim to reduce technical debt and replace shortcuts we have taken to implement certain features, leaving behind a simpler and better-organized codebase.

The central change is the separation between the frontend and the backend. Bambi now maintains a representation of the model that is completely agnostic to its representation in the backend. Although this would make it possible to add other backends, the main motivation was to clearly define the responsibilities of the library's different modules. For example, parameter transformations that use PyTensor now live in the backend, whereas they were previously methods on
family definitions. This separation also provides clear guidance for implementing new features in the future. The structured description maintained by Bambi remains the foundation for high-level tools such as `interpret`, even though model execution is delegated to the backend.

We also aim to make better use of PyMC and PyTensor. Historically, Bambi used PyMC to create and fit the model, but then performed calculations such as evaluating the observational model parameters or making predictions for new observations on its own, among others.

That design duplicated code implementing the same logic: we had components in both the frontend and the backend, with methods that implemented equivalent operations on NumPy/xarray and PyMC/PyTensor arrays, respectively. For example, the contributions of common and group-specific terms were specified when building the model and then reconstructed separately when making predictions. Maintaining both paths led us to write complex, difficult-to-follow methods to reproduce with `xarray` arrays what was already expressed in the PyMC model.

All necessary operations are now performed on the PyMC/PyTensor graph itself, avoiding duplicate implementations and reducing the risk of errors when reconstructing the model. Graph interventions also let us express the different operations we need. Although some remain complex, once implemented, we can continue taking advantage of all the tools PyMC/PyTensor offer for prediction, log-likelihood computation, and so on.

Along with this rewrite, the PR introduces new features, fixes several bugs, and revises the terminology used to describe the different parts of the model more precisely. It also introduces changes to the public API that may require users to adjust their code. We aim to make this transition gradual through deprecation notices and compatibility mechanisms that minimize breakage. The following sections describe these changes and their motivation in detail.

# New features

## Log-logistic family

The log-logistic family is added and is available directly with `family="loglogistic"`. This family is generally used in survival models and was already constructed in the example notebook; it is now built in. We implemented the `LogLogistic` distribution in Bambi because it is not available in the version of PyMC used by this PR, with the intention of contributing it to PyMC later.

## Competing risks models

The `cr(time, status, cause)` term has been implemented to model times to the first event when there are several possible causes. The response indicates the observed time, whether an event occurred or the observation was right-censored, and the cause when known.

For these models, `Model.predict(kind="time_and_cause")` has been added, allowing the time of the first event and its cause to be generated jointly. With these simulations, we can estimate the probability that the first event occurs from each cause before a given time—that is, the cumulative incidence functions. The following section describes other prediction options.

## Predictions in censored and truncated models

Although Bambi already supported censored and truncated responses, predictions did not correctly handle all the quantities that can be obtained from these models. We can make predictions from the underlying distribution, without censoring or truncation restrictions, or from the distribution conditioned on the information provided by those mechanisms.

For example, for a subject right-censored at time `t`, we might want to predict from `p(y | y > t)`. To do so, we intervene on the PyMC model and replace censoring with a distribution truncated from below at `t`. Here, truncation is a way to restrict the distribution to express that conditioning, not a change in the observation mechanism.

In a model with truncated responses—for example, if only values between `a` and `b` are recorded—we can generate samples from `p(y | a < y < b)` to predict within that interval, or from the underlying distribution `p(y)` to predict without that restriction.

These two options are selected through the `kind` argument of `Model.predict`:

- `kind="response"`: underlying distribution.
- `kind="response_conditional"`: distribution conditioned on censoring or truncation.

Both options can be used with the fitting data or with new observations. The log-likelihood can also be computed on new data by providing the corresponding censoring information or truncation bounds.

Conditional prediction is also available for competing risks: for right-censored observations, it can generate first-event times conditioned on the event not having occurred by the recorded time.

## Predictions for new groups or groups of unknown identity

In hierarchical models, predicting for a group that was not included in the fit requires generating the coefficients associated with that group, such as its intercept or slopes. Bambi already supported this through the `sample_new_groups` argument, which enabled or disabled that behavior. However, it always did so by sampling coefficients from the observed groups. We now distinguish between two situations.

For a new group whose identity is known but which was not observed during fitting, we generate its coefficients from the population-level model and the learned hyperparameters. If multiple observations belong to the same new group, they share the coefficients generated for that group.

For a group whose identity is unknown, we retain the previous approach: we assume it belongs to one of the observed groups and sample among them. The selection is shared across terms belonging to the same factor to preserve their association.

These operations are implemented through interventions on the PyMC/PyTensor computational graph. Bambi automatically selects the strategy based on the value of the grouping factor in the input data: a missing value (`None`, `np.nan`, or `pd.NA`) indicates an unknown identity, while a non-null value that was not observed during fitting identifies a new group. Therefore, the `sample_new_groups` argument is no longer used. We can combine all three cases in the prediction
data, with both dense matrices and `SPARSE_DOT`.

## Count responses with explicit totals

Bambi already supported the multinomial and Dirichlet-multinomial families, with count columns specified through `c(y1, y2, ...)`. The total `n` was obtained by summing those counts, so predicting with new totals also required providing the response columns.

The `counts(y1, y2, ..., n=n)` transformation allows the total to be specified explicitly, either as a constant or as a variable for each observation. During fitting, Bambi verifies that the counts in each row sum to that total. If `n` is omitted, it is obtained by summing the counts.

When `n` is given by a column, we can provide new totals along with the covariates and generate predictions without including the response counts. This lets us, for example, predict how a given number of cases will be distributed among the different categories.

If the formula omits `n`, the total from the first fitting observation is used when predicting on new data, and a warning is emitted. To update the totals with the prediction data, `n` must be specified as a variable in the formula and that column must be provided.

## Estimation and visualization with `interpret`

The estimation and visualization functions include the following changes:

- The `plot_*` functions now return a Matplotlib `Figure`. Plots are still built with  `seaborn.objects`, but returning a figure makes it possible to modify them through an interface that is more familiar to most users.
- The `prob` argument accepts `None` to obtain point estimates without computing credible intervals. In the `plot_*` functions, this also makes it possible to draw plots without intervals.
- The `on=` argument of the `plot_*` functions makes it possible to draw on an existing `Axes`, `Figure`, or `SubFigure`. This allows us to combine multiple `interpret` plots in the same figure or integrate them with other visualizations.
- For responses with multiple categories, the response name can be used in `subplot_kwargs` to organize groups or panels. The ordering of levels specified by the user is also respected.
- The `legend` option in `fig_kwargs` accepts a boolean to control whether the legend is displayed; for example, `fig_kwargs={"legend": False}` hides it.

## Automatic sampler selection

The default value of the `inference_method` argument in `Model.fit` changes from `"pymc"` to `None`. With this value, PyMC chooses the sampler based on which samplers are available and compatible with the model and fitting options. This includes the possibility of using `nutpie` automatically when it is installed.

# Changes to existing functionality

## Response variable names

Transformed responses use simpler names in the PyMC model and in the results. For example, `censored(y, status)` and `truncated(y, lower, upper)` use `y` instead of the entire formula expression. The same applies to `constrained(y, ...)`, `weighted(y, ...)`, and `prop(y, n)`. For `cr(time, status, cause)`, the base name is `time`, and for `counts(y1, y2, ...)`, the column
names are combined, as in `y1_y2`.

This makes variables easier to access and prevents their names from depending on auxiliary transformation arguments. Aliases still make it possible to choose another name. Code that accesses results using the full expression will need to use the new name or an alias.

## Prediction organization

When we pass `data` to `Model.predict`, results are stored in the `predictions` group of the `DataTree`, both for response parameters and predictive samples. Auxiliary data is stored in `predictions_constant_data` when applicable. This keeps evaluations on new observations separate from the posterior and from predictions on the fitting data, which may have a different number of observations. `interpret` also uses these groups when evaluating its covariate grids.

Without `data`, parameters continue to be stored in `posterior` and response samples in `posterior_predictive`.

## Progress bar for prediction and log-likelihood

`Model.predict` and `Model.compute_log_likelihood` accept a `progressbar` argument, whose default value is `False`. It can be enabled with `progressbar=True` to monitor the progress of these operations. PyMC logging messages are also silenced during these evaluations to avoid unnecessary output.

## Model parameters and components

The names Bambi uses to describe the model have been revised. `parameters` replaces `components` and distinguishes between `marginal_parameters` and `conditional_parameters`. Here, “marginal” refers to the covariates: the parameter does not depend on them and is shared by all observations. “Conditional” indicates that the parameter depends on the covariates and is obtained by transforming a systematic component, so it may vary between observations. That transformation may be the identity. These names replace `constant_components` and `distributional_components`, respectively. The response is described through `response_term` instead of the `response_component` wrapper.

The intention is to distinguish the parameters of the response distribution from the coefficients that are part of the terms in the systematic component. The name “components” was too generic to express that distinction.

## Family definitions

Families become lightweight descriptions in the frontend: they specify the likelihood, links, default priors, and metadata for the response and its parameters. They are no longer separated into univariate and multivariate base classes; attributes such as `Family.DATA_TYPE`, `Family.RESPONSE_NDIM`, and `Family.PARAMETERS` describe the response type, dimensions, and allowed links. In `Family.PARAMETERS`, each `ParamSpec` indicates the supported links, the dimensionality of the parameter, and the dimensions of its coefficients, described through `DimType`.

If a family needs to transform data, parameters, or the predictor to build the model, that implementation is defined in the backend, outside the family class. The transformation addresses a need of the statistical model, but how it is implemented depends on the operations and parameterization provided by the backend.

For example, Bambi describes the Beta family in terms of `mu` and `kappa`, while the backend uses `alpha` and `beta`. That conversion is no longer in the `Beta` class; it is now registered in the backend. Previously, it was duplicated in `Beta.transform_backend_kwargs` and `Beta.transform_kwargs` to build the model and evaluate predictions, respectively.

With this separation, several family methods are removed, including:

- `Family.posterior_predictive` and `Family.log_likelihood`: these operations are delegated to the backend.
- `Family._make_dist_kwargs_and_coords`: distribution arguments and coordinates are no longer reconstructed from posterior samples.
- Hooks for specific families such as `Beta.transform_backend_kwargs`, `Beta.transform_kwargs`, `Multinomial.transform_backend_eta`, `Multinomial.transform_linear_predictor`, and `Multinomial.transform_coords`: separate paths for PyTensor and `xarray` are no longer maintained.

To define a custom family, we can create a subclass of `Family` containing its description and, when necessary, register adaptations through the backend decorators `TransformsRegistry.transform_predictor`, `TransformsRegistry.transform_parameters`, and `TransformsRegistry.transform_data`. Built-in families use the same mechanism.

## Custom links

When defining a custom `Link`, a single `inverse_link` function compatible with the PyMC/PyTensor backend is used instead of the `linkinv` and `linkinv_backend` functions. We no longer need one inverse for NumPy computations and another for the graph. The forward `link` function is optional.

# Dependency, packaging, and CI changes

This PR updates several major dependencies: `pymc`, `pytensor`, `formulae`, `arviz`, `arviz_plots`, `arviz_stats`, and `matplotlib`.

Several libraries with which Bambi interacts directly are also declared explicitly, even though some are already included as transitive dependencies of PyMC or other packages: `numpy`, `pandas`, `scipy`, `xarray`, `seaborn`, and `graphviz`.

`sparse` is no longer a dependency because Bambi no longer performs manual operations with xarray sparse matrices.

The `jax` extra updates `blackjax` and `numpyro` and adds `fastprogress`. The `nutpie` extra no longer pins `numba` directly. `preliz` is also added to the dependencies used for tests and documentation.

`pytest-forked` is added to the test dependencies because it is used in manual testing to prevent the numba backend from exhausting the available RAM.

Black is updated with Jupyter support, and pre-commit now checks notebooks as well. I updated the line-length to be at most 88 in notebooks because that is the maximum supported in the documentation webpage.

Finally, setuptools now discovers subpackages through `tool.setuptools.packages.find`, which is necessary now that the PyMC backend is organized as a package with multiple submodules.

# Significant internal changes

## Frontend

Families, parameters, and terms now focus on describing the model, including its data and specifications. They still handle validation and prepare the information needed to build the model, but the backend is responsible for constructing and evaluating the PyMC model. The frontend can still treat particular families differently when applying statistical rules, such
as prior scaling. What it no longer contains is logic specific to running models with PyMC.

### Prediction and log-likelihood moved out of families and parameters

`Family` no longer has the `posterior_predictive` and `log_likelihood` methods. The backend is now responsible for those tasks. The family describes important aspects of the model, but does not build it.

The same separation applies to parameters. `ConditionalParameter` retains the description of terms and the preparation of their priors, but not the prediction methods of the former `DistributionalComponent`: `predict`, `predict_common`, `predict_group_specific`, and `_construct_u_with_new_groups`. This removes the reconstruction, with NumPy and `xarray`, of term contributions and the coefficients needed to predict for new groups.

### More descriptive terms

Implementation responsibilities are also removed from terms, although this change is smaller than the changes to families and parameters. Properties such as `CommonTerm.coords`, `GroupSpecificTerm.coords`, and `HSGPTerm.coords` are removed. Coordinates are not inherent to the terms; they are needed by the PyMC backend. Terms describe their levels and structure, and the backend decides how to translate that information into coordinates for the PyMC model.

Accessors such as `BaseTerm.label`, `GroupSpecificTerm.expr`, and `GroupSpecificTerm.factor` are added, along with the names and types of the expression and factor. This allows the backend to consume the term description without repeatedly traversing Formulae's internal structure. `ResponseTerm` no longer stores a reference to the family or queries its hooks to obtain data, levels, and coordinates.

### Default prior specification

Family prior defaults are expressed through self-contained dictionaries containing the distribution name and its arguments. PyMC-specific transformations, such as `"ordered"` for ordinal thresholds, are specified by name and resolved in the backend. As a result, preparing the defaults does not require constructing those PyMC objects in the frontend.

### Binomial prediction grids in `interpret`

When preparing a grid for a binomial model, `interpret` uses one trial when the total is given by a variable that does not appear in the grid. This makes it possible to summarize the probability `p` without requiring that additional column. If the total is provided in the grid or is a literal value in the formula, it is preserved.

### Function-based construction

The backend class hierarchy that duplicated the representation of terms and components is removed. In many cases, we created one of those objects from the frontend object, called its construction method, and used the result. For example, the backend instantiated its own `InterceptTerm` class to call `InterceptTerm.build`.

The approach is now more functional: a function receives the frontend description, together with the metadata and PyMC model it needs, and translates it into graph data, variables, and operations. Functions such as `build_intercept_term`, `build_common_term`, `build_conditional_parameter`, and `build_marginal_parameter` replace those builder classes. The term or parameter description does not need to be duplicated in an object with its own `build` method.

This does not mean that the backend no longer has classes or state. Structures such as `CommonTermInfo` and `ConditionalParameterInfo` are implemented to organize metadata, along with others that store the graph references needed during prediction. The difference is that this information is kept separate from the functions that build or modify the model.

The construction of potentials also validates the referenced variables and verifies that the constraints are callable. Error messages indicate which variables are missing and which are available in the model.

### Graph evaluation and interventions

Conditional parameters are represented as deterministic variables in the PyMC model. The backend can evaluate them with `pymc.compute_deterministics` when needed, without storing their values for every observation during fitting or reimplementing their computations.

The data needed to evaluate the model is registered in PyMC through `pymc.Data`. To obtain predictions for new observations, the backend clones the model, replaces the data, and updates the coordinates with `pymc.set_data`. It then uses `sample_posterior_predictive` to delegate the generation of predictions to PyMC. When an operation requires changing the graph, such as excluding group-specific effects or generating a response conditioned on censoring, interventions such as `pymc.do` are used on a clone. This keeps the originally constructed PyMC model intact.

Other tasks delegated to the backend include:

- Drawing samples from the prior and prior predictive distribution.
- Computing the log-prior.

### Predictor centering and the intercept

Predictor centering and conversion of the intercept back to the original scale are expressed in the PyMC graph. Previously, intercept samples were corrected after fitting, and that correction had to be reversed for some operations through `Model._re_center_intercept`. The intercept on the original scale is now a deterministic variable in the model.

### Package organization

`bambi.backend.pymc` changes from a module to a package. This organization consolidates the PyMC/PyTensor implementation in the backend and makes it possible to modify each part without redistributing the logic between frontend and backend classes again.

# Documentation

The notebooks are updated to reflect the API changes, result names, and visualization changes. The survival examples cover competing risks and cumulative incidence functions, while the `interpret` examples show how to compose figures with the new interface.

We also aim to unify the presentation of plots in the documentation. At least temporarily, `docs/notebooks/bambi.mplstyle` is added as a shared theme, with a common palette and a consistent base figure size. The goal is for all plots to follow the same theme and maintain a consistent width without exceeding the maximum supported by the Bambi website. This prevents the scaling of each figure from producing differences in the apparent size of text, lines, and other visual elements.

# Compatibility and release plan

The intention is to merge this branch into `dev` after adding warnings and adapters for the public interfaces that need them. We want to announce deprecations before removing APIs or changing behavior that may affect user code.

In addition, the warnings will also be incorporated into the current `main` branch and published in a release prior to the release that includes the rewrite. The idea is to have at least one release warning about the changes before they take effect, followed by another release that includes the changes while retaining the necessary adapters. There will very likely be some breaking changes, even though we will try to minimize them.


GStechschulte:
😮

PR #1006, opened 2026-09-08
Add nonlinear formulas

tjburch:
# Add nonlinear formulas

This PR adds nonlinear expressions for a likelihood's parent parameter. Users provide the expression and an ordinary Bambi formula for each named nonlinear parameter:

```python
formula = bmb.Formula(
    "y ~ a + b * exp(-k * x)",
    "a ~ 1 + treatment",
    "b ~ 1",
    "k ~ 1",
    nonlinear=True,
)
model = bmb.Model(formula, data, priors=priors)
```

The main right-hand side is an arithmetic expression with no implicit intercept. Parameter formulas retain ordinary Bambi semantics, including common, group-specific, and offset terms.

## Implementation

- Support for the following operations: `+`, `-`, `*`, `/`, `**`, `exp`, `log`, and `sqrt`.
- Nonlinear predictors reuse `ConditionalParameter` and the existing prior and term machinery.
- The expression is evaluated on the family link scale. The backend applies the inverse link once before building the likelihood.
- Ordinary auxiliary formulas, such as `sigma ~ z`, and existing alias conventions remain available.
- Prediction reuses the fitted PyMC graph. Expression covariates are mutable data, so `Model.predict()` supports new observations without a separate NumPy evaluator.
- Missing-row filtering covers the response, parameter formulas, auxiliary formulas, and raw expression inputs together. All required prediction inputs must be complete.

The nonlinear models guide in the PR covers priors, group-specific parameters, auxiliary formulas, links, aliases, and prediction. Would appreciate feedback on this.

## Current limitations

Still a decent amount of compatibility/functionality to add, but I think this is a good starting point.

- Only one nonlinear expression for a scalar likelihood parent is supported. This excludes vector-valued parents (e.g. categorical and ordinal).
- The response must be univariate and represented by one observed column. So, binomial count/trial pairs and censored, truncated, weighted, or multiple responses are not yet supported.
- Nonlinear auxiliary expressions and dependencies among nonlinear parameters are not supported. Auxiliary parameters can still use ordinary Bambi formulas.
- The `interpret` API does not support nonlinear models yet. Use `Model.predict()` instead, passing explicit data when predicting for new predictor values.
- Nonlinear predictor coefficients are not automatically scaled from the response. Explicit priors are recommended because nonlinear models can be weakly identified.
- The operations listed above are the full list of supported operations. Things like `sin`, `cos`, `tan`, `tanh`, `abs`, `log1p`, `expm1`, and a sigmoid are likely future adds.

## Changes that also affect ordinary models

A few small changes were needed to make aliases and offsets behave consistently with the new nonlinear parameters. They apply to all models, not just ones with `nonlinear=True`.

- Non-centered offset variables are now recorded in `model.__bambi_attrs__["offset_names"]` when they are created. `omit_offsets` filters on that set instead of the `_offset` name suffix, so a user alias ending in `_offset` is no longer dropped.
- `omit_group_specific` and prior-only predictive filtering now match on aliased labels rather than original names.
- The model summary prints marginal parameter aliases (e.g. `noise ~ HalfStudentT(...)` after aliasing `sigma`).


tomicapretto:
Thanks a lot for the contribution @tjburch!

I can't provide a comprehensive review right now, but I can drop some questions in the meanwhile:

- Is there anything fundamental that prevents us from using these nonlinear expressions in non-parent parameters? It's OK if we don't support it now, but if the current design is blocking it, I'd like to know sooner than later.
- Do you think it would be possible to have both this approach and the approach I suggested in the roadmap? I suggested the function based approach because it would be simpler to handle. But if we can have both, then that's amazing.
- What limits the usage of the interpret sub-package? Again, it's OK if it is not supported in this PR, but if there's anything fundamental blocking it, I'd like to work on the necessary modifications.
- As for the example, I think the one you're adding does not have executable code and thus `.qmd` is good. But if we wanted to have it executed, we would still need a notebook. In the future, if migrate from quartodoc to great-docs (first one kind of supported by posit then abandoned, second is their main offering now), I think it's possible to have qmd files that are executed only if needed.
- Is it possible to not pass formulas such as `a ~ 1` when the parameter `a` does not depend on any covariate?

tjburch:
> Is there anything fundamental that prevents us from using these nonlinear expressions in non-parent parameters?

Nothing fundamental, mostly effort and a lot of additional bookkeeping. E.g. if building a normal and you want sigma to depend on mu, then you have to auto-detect which to build first, you have to avoid circular definitions, etc.

> Do you think it would be possible to have both this approach and the approach I suggested in the roadmap? I suggested the function based approach because it would be simpler to handle. But if we can have both, then that's amazing.

Don't see any stoppers to dual-support. This approach basically converts the expression into a small internal expression tree, the registered function could easily do the same.

> What limits the usage of the interpret sub-package?

~~Also nothing fundamental here. Mostly just effort and trying to get a first-version up quickly.  The main add would be discovering variables appearing only in the nonlinear expression. Since they're already recorded by the model this is probably a lower barrier to entry relative to the first point above, so if we want to sweep it in, I could add it.~~

Enabled on latest commit.

> As for the example, I think the one you're adding does not have executable code and thus .qmd is good

Yeah - the current one really reads more as a tour of the functionality more than an example. I think a strong practical use case example would be helpful here, just haven't had time to set one up. 

> Is it possible to not pass formulas such as a ~ 1 when the parameter a does not depend on any covariate?

~~Good question. In the current setup, that basically serves to inform the parent that `a` is a parameter and not a data column. We _could_ swap that to basically "if it's not a data name, assume it's an intercept-only parameter." Might be a little risky and hide things like spelling errors though. Maybe assume it's an intercept-only and give a logging statement informing that we've inferred it as a parameter? or possibly expose a parameter list, like~~

```python
  bmb.Formula("y ~ a * x", nonlinear=True, nlpars=("a",))
```

~~I'll leave that to your preference ultimately.~~

Switched to this per-guidance

bambi/formula.py
```diff
@@ -18,13 +18,37 @@ class Formula:
         A model description written using the formula syntax from the `formulae` library.
     *additionals : tuple of str
         Additional formulas that describe model parameters rather than a response variable.
+    nonlinear : bool, optional
```
tomicapretto, line 21: I'd rather have `nlpars=(name1, name2, ...)` instead of a global `nonlinear`. With the former we know formulas using at least one of those parameters have a non-linear expression. With the latter, we know some formula has a nonlinear expression (or assume it's all of them). But on top of that the latter also requires the explicit `~ 1` formula. With `nlpars` we only require it when the parameter is not constant.

bambi/models.py
```diff
@@ -1327,7 +1510,64 @@ def marginal_parameters(self):
...
+
+        Examples
+        --------
+        For ``Formula("y ~ a * x", "a ~ 1", nonlinear=True)``, retrieve the predictor
+        with ``model.nonlinear_predictors["a"]``.
+        """
+        return self._nonlinear_predictors.copy()
+
+    @property
+    def additive_parameters(self):
```
tomicapretto, line 1551: I am not sure about what is the right abstraction here, but I don't like `additive_parameters` much. 
I think we could add `nonlinear_parameters` but that would be against the fact that `parameters` represent parameters of the observational model: marginal when do not depend on any covariate, conditional when they do depend on covariates. Fun fact: I was considering additive parameters for conditional parameters. 

Anyway, I think the parameters of non-linear expressions play the role of a coefficient in a regular model term (e.g., `beta` in `beta * x`). Thus `nonlinear_coefficients` or `nonlinear_coefs` is prefered. 

On top of that, I'm realizing that nonlinear_coefficients should not be a model-level attribute, but a conditionalparameter attribute. That way it becomes almost trivial to support nonlinear expressions in any parameters (yes, we would only be able to write `b ~ ...` for the parent par and `sigma_b ~ ...` for a sigma par).

Does this make sense? 

Yes, I also write as I think. Nope, I am not an LLM (yet)

tomicapretto:
@tjburch I think https://github.com/bambinos/bambi/pull/1006#discussion_r3964368889 is the most important comment. If we agree on it, then more changes should be coming so I'd rather review those.

And for the tests, better to not create so many test files. You could add tests to test_formulas, test_model_construction, and test_models. I can see the tests for the `nonlinear.py` do not fit in there. No problem to have a new test file for those.

bambi/models.py
tjburch: Believe it does make sense. I'll work on updating and possibly follow up with questions as I have them.

bambi/formula.py
tjburch: latest commit ([ce62553](https://github.com/bambinos/bambi/pull/1006/commits/ce62553be7a7181a3421962c1b8df344444be46b)) covers this

PR #1066, opened 2026-09-13
Fix non-centered Normal group locations

### LINKED ISSUE #1003: Non-centered group-specific Normal silently drops mu and leaves mu hyperpriors disconnected
### Summary

When a group-specific `Normal` prior has a nonzero or hierarchical `mu` and
its effective parameterization is non-centered, Bambi accepts and displays
the prior but builds the group effect as only `offset * sigma`.

For a scalar nonzero `mu`, the value is silently ignored. For a `mu`
hyperprior, Bambi first creates the hyperprior RV and then leaves it
disconnected from both the group coefficient and likelihood.

I reproduced this with Bambi 0.20.0 and current `main` at
`c2642c5bd5f824c4ba441a8338d86c07adad6341`.

### Minimal reproducer

```python
import bambi as bmb
import pandas as pd
from pytensor.graph.traversal import ancestors

data = pd.DataFrame(
    {
        "y": [0.2, 0.4, 0.3, 0.7, 0.9, 1.0],
        "x": [-1.0, 1.0, -1.0, 1.0, -1.0, 1.0],
        "group": ["a", "a", "b", "b", "c", "c"],
    }
)

prior = bmb.Prior(
    "Normal",
    mu=bmb.Prior("Normal", mu=2.0, sigma=0.25),
    sigma=bmb.Prior("HalfNormal", sigma=1.0),
)

model = bmb.Model(
    "y ~ 0 + (0 + x | group)",
    data,
    priors={"x|group": prior},
    noncentered=True,
)
model.build()

pm_model = model.backend.model
coefficient_ancestors = set(ancestors([pm_model["x|group"]]))
likelihood_ancestors = set(ancestors(pm_model.observed_RVs))
mu_rv = pm_model["x|group_mu"]

print("bambi", bmb.__version__)
print("mu used by coefficient:", mu_rv in coefficient_ancestors)
print("mu used by likelihood:", mu_rv in likelihood_ancestors)
print(
    "disconnected:",
    [rv.name for rv in pm_model.free_RVs if rv not in likelihood_ancestors],
)
```

Output:

```text
bambi 0.20.0
mu used by coefficient: False
mu used by likelihood: False
disconnected: ['x|group_mu']
```

The current implementation recursively creates `dist_kwargs["mu"]`, but the
non-centered branch returns:

```python
pm.Deterministic(label, offset * sigma, dims=dims)
```

so `mu` is not used:
https://github.com/bambinos/bambi/blob/c2642c5bd5f824c4ba441a8338d86c07adad6341/bambi/backend/terms.py#L182-L214

A scalar `mu=2` with a hierarchical `sigma` has the same semantic problem but
does not leave a named orphan: the resulting group coefficient remains centered
at zero. With `noncentered=False`, the same `mu` is used. Thus changing the
parameterization can currently change the specified model rather than only its
geometry.

### Expected behavior

Issues #785 and #807 explain that Bambi group-specific effects are intended to
be zero-centered deflections and that the corresponding common effect owns the
population location. Given that convention, the least surprising behavior
seems to be:

- when non-centering is effective for a group-specific Normal, require `mu` to
  be scalar/array zero;
- reject a nonzero or `Prior`-valued `mu` before recursively creating it; and
- give an actionable error, for example:

  > Non-centered group-specific effects are zero-centered deviations, so `mu`
  > must be 0. Add the corresponding common effect for the population location,
  > or set `noncentered=False` for this prior.

This preserves Bambi's established semantics while avoiding both silent model
changes and disconnected variables.

If arbitrary group means are intended to be supported instead, the
mathematically equivalent non-centered transform would be
`mu + offset * sigma`. That would be a larger semantic change and can be
redundant with a matching common effect, so validation seems more consistent
with the explanations in #785 and #807.

### Suggested tests

1. `mu=0` plus hierarchical `sigma` still builds `offset * sigma`.
2. Scalar nonzero `mu` plus hierarchical `sigma` raises an actionable error.
3. `mu=Prior(...)` plus hierarchical `sigma` raises before any `*_mu` RV is
   created.
4. `noncentered=False` continues to allow and connect scalar or hierarchical
   `mu`.
5. A per-prior `noncentered=False` override permits `mu` when the model default
   is `True`.
6. A per-prior `noncentered=True` override rejects `mu` when the model default
   is `False`.
7. Vector all-zero `mu` is accepted; a vector containing a nonzero value is
   rejected.

### Related issues

- #785 explains the zero-centered-deflection convention but does not cover
  accepting and then dropping `mu`: https://github.com/bambinos/bambi/issues/785
- #807 describes the same common-effect/group-deflection semantics:
  https://github.com/bambinos/bambi/issues/807
- PR #983 added parameterization controls while retaining the `offset * sigma`
  construction: https://github.com/bambinos/bambi/pull/983

This was initially surfaced while diagnosing HSSM discussion #948:
https://github.com/lnccbrown/HSSM/discussions/948

The downstream HSSM default-prior bug is tracked separately in
https://github.com/lnccbrown/HSSM/issues/1224.

AlexanderFengler:
Closes #1003.

Companion main-line backport: #1069. The two PRs provide the correction on their respective backend implementations; maintainers can choose merge order and release timing independently.

## Goal

This PR makes non-centered Normal group-specific priors respect the location specified by the user. Choosing between centered and non-centered parameterizations should change how Bambi represents the model, not which statistical model it fits.

For a group coefficient $b_j$, the intended prior is

$$
b_j \mid \mu, \sigma \sim \mathcal{N}(\mu, \sigma^2).
$$

Its non-centered representation is

$$
z_j \sim \mathcal{N}(0, 1), \qquad b_j = \mu + \sigma z_j.
$$

Currently, Bambi constructs $b_j = \sigma z_j$, dropping $\mu$. A fixed nonzero location is therefore ignored, and a location hyperprior no longer influences the group coefficient or its likelihood contribution. This PR restores the missing location term. For $\mu=0$, the expression reduces to the usual zero-mean case.

## What changes in regression formulas?

Formula syntax, design matrices, links, and default priors do not change. The correction changes the linear predictor implied by a *supplied* Normal group prior when its non-centered construction previously ignored the location. Write $\eta_i$ for the linear predictor and $j[i]$ for observation $i$'s group.

- **Group intercept without a common intercept:** `y ~ 0 + (1|group)` has $\eta_i=b_{j[i]}$. With a supplied location $\mu$, the old non-centered construction gave $\eta_i=\sigma z_{j[i]}$; the corrected construction gives $\eta_i=\mu+\sigma z_{j[i]}$. Thus `0 +` removes the common intercept, but does **not** require the group coefficients' population location to be zero. A location hyperprior can now be learned through the group coefficients.
- **Common intercept plus group deviations:** `y ~ 1 + (1|group)` with the usual zero-location group prior remains $\eta_i=\alpha+\sigma z_{j[i]}$. If the user instead supplies a group location $\mu$, the corrected predictor is $\eta_i=\alpha+\mu+\sigma z_{j[i]}$. If both $\alpha$ and $\mu$ are free, the likelihood identifies their sum, not the two locations separately; this PR does not resolve that redundancy automatically.
- **Group slopes:** `y ~ 0 + (0 + x|group)` has $\eta_i=x_i b_{j[i]}$. The same correction restores $x_i\mu$, giving $\eta_i=x_i(\mu+\sigma z_{j[i]})$ rather than $x_i\sigma z_{j[i]}$.

These expressions are on the **linear-predictor/link scale**, not generally the response scale. No nonzero location or location hyperprior is introduced automatically. The existing hierarchical-scale requirement for Normal non-centering still applies.

## Why prediction and log-density evaluation also change

By default, `fit()` omits the auxiliary offsets $z_j$ from the returned draws, while retaining the group coefficients $b_j$. When prediction or log-likelihood evaluation needs those offsets, the correct inverse is

$$
z_j = \frac{b_j - \mu}{\sigma},
$$

not $b_j / \sigma$. Keeping the old inverse after fixing the forward expression would add the location a second time when recomputing $b_j$.

The same correction is needed at each non-centered Normal node in a nested prior. For example, if $\mu$ itself has a non-centered Normal prior, its offset must also be recoverable. Bambi records the transformations used when building the model so reconstruction follows the actual locations, dimensions, aliases, and per-prior settings. Offsets already present in the draws are left unchanged.

Log-prior evaluation may also need a missing nested offset to evaluate another variable's density. Those offsets are restored as inputs when needed; this does not add them to the variables reported in the log-prior output.

## Compatibility and boundaries

This keeps the existing Normal non-centering eligibility rules, including the requirement for a hierarchical scale. It does not add non-centering support for other distributions.

Models previously fitted with an ignored nonzero or random location need to be refitted. Adding the location to old draws does not, in general, recover the posterior of the corrected model.

Testing also exposed a separate issue on `dev`: with `SPARSE_DOT=True`, predictions for previously unseen group labels can ignore the supplied location and scale draws. This is now tracked in #1068, with a reproducer on unmodified upstream `dev`. The same reproducer works correctly with `SPARSE_DOT=False`; the failure also occurs with centered priors and predates this PR. Sparse new-group prediction is not changed here and should be fixed separately.


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   ([#1003](https://github.com/bambinos/bambi/issues/1003)). Previously, nonzero
   locations were discarded. Affected models need to be refitted; adding `mu` to
   old posterior draws does not generally recover the corrected posterior.
+- Reconstruct omitted Normal offsets with `(coefficient - mu) / sigma` for
+  prediction and log-likelihood, including nested hyperpriors and per-prior
+  non-centering overrides. Restore missing auxiliary inputs when needed by
+  log-prior evaluation without adding them to the reported log-prior variables.
```
tomicapretto: This is added automatically later, could you remove?

tomicapretto (changes requested):

tomicapretto:
@AlexanderFengler could you write the description in a more human-readable way? It's quite painful to read, at least to me. Also, try to keep the details that are only relevant to Bambi. For example, the very first link is not only not relevant but it's also broken.

AlexanderFengler:
> @AlexanderFengler could you write the description in a more human-readable way? It's quite painful to read, at least to me. Also, try to keep the details that are only relevant to Bambi. For example, the very first link is not only not relevant but it's also broken.

My bad this is ended up inside a larger effort on the HSSM side (HSSMSpine is the dev superpackage) so it's referring to it's place inside that effort, but the context is not useful here obviously.

Improving it now.

tomicapretto:
I am not against adding this feature to Bambi, but I would like better understand the need. Did you have to use it because the usual way of writing it didn't work and this did?

That is, did it happen to you that `y ~ 1 + (1 | group)` with `{"Intercept": Prior("Normal", ...), "1|group": Prior("Normal", mu=0, sigma=...)}` didn't work but `y ~ 0 + (1 | group)` with `{"1|group": Prior("Normal", mu=Prior("Normal", ...), sigma=...)}` did?

Issue #1068, opened 2026-09-16
Sparse new-group prediction ignores supplied location and scale draws on dev


AlexanderFengler:
## Summary

On the current `dev` branch, predictions for a previously unseen group label with `SPARSE_DOT=True` do not follow changes in the supplied draws of the group-location and scale hyperparameters. Dense prediction does. The setting should change the computation, not the conditioning of the predictive distribution.

This reproduces with centered priors on unmodified `dev` at `5b7e1b1f42097adeb1b9d846e30c31aa91616332`, without the changes in #1066. I have not established whether a released version is affected.

## Reproducer

The test uses PyMC's mock sampler only to obtain a small, correctly shaped trace. It then checks a deterministic property of prediction, not posterior convergence. Holding the prediction seed fixed, shifting both the stored coefficients and their population location by 3 should shift both known- and new-group predictions by 3.

```python
from unittest.mock import patch

import bambi as bmb
import numpy as np
import pandas as pd
from pymc.testing import mock_sample

data = pd.DataFrame({
    "y": [0., 1., 2., 3., 4., 5.],
    "g": ["a", "a", "b", "b", "c", "c"],
})
new_data = pd.DataFrame({"g": ["a", "new_group"]})
original_sparse = bmb.config.SPARSE_DOT
try:
    for sparse in (False, True):
        bmb.config.SPARSE_DOT = sparse
        model = bmb.Model(
            "y ~ 0 + (1|g)", data, noncentered=False,
            priors={"1|g": bmb.Prior(
                "Normal",
                mu=bmb.Prior("Normal", mu=2, sigma=1),
                sigma=bmb.Prior("HalfNormal", sigma=1),
            )},
        )
        with patch("pymc.sample", mock_sample):
            trace = model.fit(draws=10, chains=2, random_seed=12)

        before = model.predict(trace, data=new_data, random_seed=42, inplace=False)
        shifted = trace.copy(deep=True)
        shifted.posterior["1|g"] += 3
        shifted.posterior["1|g_mu"] += 3
        after = model.predict(shifted, data=new_data, random_seed=42, inplace=False)
        delta = (after.predictions["mu"] - before.predictions["mu"]).values
        print(sparse, np.unique(np.round(delta[..., 0], 8)),
              np.unique(np.round(delta[..., 1], 8)))
finally:
    bmb.config.SPARSE_DOT = original_sparse
```

Output (sparse flag, known-group shift, new-group shift):

```text
False [3.] [3.]
[...]

PR #1069, opened 2026-09-16
Fix non-centered Normal group locations on main

AlexanderFengler:
Companion to #1066 (`dev`), targeting `main` for #1003. It's the same Normal-location correction on the current release without depending on the backend refactor, in case it makes sense to include this into an intermediate release.

## Goal

Selecting non-centering should change the representation, not the statistical model. For

$$
b_j \mid \mu, \sigma \sim \mathcal{N}(\mu, \sigma^2),
$$

the equivalent non-centered representation is

$$
z_j \sim \mathcal{N}(0,1), \qquad b_j = \mu + \sigma z_j.
$$

The current implementation drops the location and builds only $\sigma z_j$. This ignores fixed nonzero locations and disconnects location hyperpriors from the group coefficients. The fix restores $\mu$, while preserving the existing eligibility rules and parameterization settings.

## What changes in regression formulas?

Formula syntax, design matrices, links, and default priors do not change. The correction changes the linear predictor implied by a *supplied* Normal group prior when its non-centered construction previously ignored the location. Write $\eta_i$ for the linear predictor and $j[i]$ for observation $i$'s group.

- **Group intercept without a common intercept:** `y ~ 0 + (1|group)` has $\eta_i=b_{j[i]}$. With a supplied location $\mu$, the old non-centered construction gave $\eta_i=\sigma z_{j[i]}$; the corrected construction gives $\eta_i=\mu+\sigma z_{j[i]}$. Thus `0 +` removes the common intercept, but does **not** require the group coefficients' population location to be zero. A location hyperprior can now be learned through the group coefficients.
- **Common intercept plus group deviations:** `y ~ 1 + (1|group)` with the usual zero-location group prior remains $\eta_i=\alpha+\sigma z_{j[i]}$. If the user instead supplies a group location $\mu$, the corrected predictor is $\eta_i=\alpha+\mu+\sigma z_{j[i]}$. If both $\alpha$ and $\mu$ are free, the likelihood identifies their sum, not the two locations separately; this PR does not resolve that redundancy automatically.
- **Group slopes:** `y ~ 0 + (0 + x|group)` has $\eta_i=x_i b_{j[i]}$. The same correction restores $x_i\mu$, giving $\eta_i=x_i(\mu+\sigma z_{j[i]})$ rather than $x_i\sigma z_{j[i]}$.

These expressions are on the **linear-predictor/link scale**, not generally the response scale. No nonzero location or location hyperprior is introduced automatically. The existing hierarchical-scale requirement for Normal non-centering still applies.

## Backport details

When `fit()` omits offsets, their inverse must likewise use

$$
z_j = \frac{b_j - \mu}{\sigma}.
$$

The backend records the Normal transformations actually built, including nested nodes. Reconstruction follows their actual names and dimensions and leaves retained offsets unchanged. Log-prior evaluation supplies missing nested offsets only when needed as inputs; it does not add density terms to the reported output.

It updates the existing `_re_center_intercept()` reconstruction path; ordinary prediction and log-likelihood on this branch already use natural coefficients directly. It does not port the refactored prediction machinery, expand support to other prior families, or change dependencies.

Models previously fitted with an ignored nonzero or random location need to be refitted. Shifting old draws does not generally recover the corrected posterior.

## Verification and boundaries

The independent sparse new-group conditioning defect on `dev` is tracked in #1068 and is not changed here. Main's donor-based new-group path passes the location-shift check in both sparse and dense modes. Expression and response dimensions are tested separately; existing combined categorical-slope shape problems are outside this patch. No manual changelog entry is added, following the review of #1066.


AlexanderFengler:
@tomicapretto in case it make sense to release this before the refactor gets put onto main.

bambi/backend/pymc.py
```diff
@@ -56,6 +57,7 @@ def __init__(self):
         self.spec = None
         self.components = {}
         self.response_component = None
+        self.noncentered_distributions = {}
```
tomicapretto, line 60: Could this be `_noncentered_group_specific`?

bambi/backend/terms.py
```diff
@@ -106,6 +106,7 @@ class GroupSpecificTerm:
     def __init__(self, term, noncentered):
         self.term = term
         self.noncentered = noncentered
+        self.noncentered_distributions = {}
```
tomicapretto, line 109: Here `noncentered_group_specific`, but not private since the `PyMCModel` will use it.

bambi/backend/terms.py
```diff
@@ -149,7 +150,9 @@ def build(self, spec):
         # * (f_j, K):      when factor_dims and response_dims
         # * (f_j, e_j):    when factor_dims and expr_dims
         # * (f_j, ):       when factor_dims
-        return self.build_distribution(prior=self.term.prior, label=self.name, dims=dims)
+        coefficient = self.build_distribution(prior=self.term.prior, label=self.name, dims=dims)
+        spec.backend.noncentered_distributions.update(self.noncentered_distributions)
+        return coefficient
```
tomicapretto, line 155: `GroupSpecificTerm` could populate its own mapping and then `PyMCModel` use that mapping to perform the update on its side. I'd like to not have the term modify the PyMC model representation. Instead, the PyMC model representation is responsible for modifying itself.

bambi/models.py
```diff
@@ -1225,14 +1226,22 @@ def compute_log_prior(self, idata, inplace=True):
             rv for rv in umodel.free_RVs if rv.name not in det_names and rv.name in posterior
         ]
         target_names = [rv.name for rv in target_rvs]
-        value_vars = [umodel.rvs_to_values[rv] for rv in target_rvs]
+        # Omitted offsets may be needed to evaluate a retained variable's density,
+        # but are not themselves added to the reported log-prior targets.
```
tomicapretto, line 1230:
```suggestion
        # Recover missing offsets needed to compute other variables' log-priors.
        # Do not include the offsets' own log-priors in the output.
```

I think this is easier to read.

tomicapretto (changes requested):
Thanks for the fix! I'm requesting some changes related to the organization of a few things.

tomicapretto:
A final ask regarding tests: I think they're testing things we need to have tested. However the code amount of tests and the code look a bit too much/not-clean. Even though cleanness can be consider personal choice, I'd ask if it's possible to reduce test in some way to reduce CI time. However, if you can show me the tests don't really take much, just a few seconds (<10?) then all good.

